PR #4, opened 2019-01-31
Resolving malformed data service definitions.

sumanapp:
Signed-off-by: sumanapp <sappadwedula@rocketsoftware.com>


c/dataservice.c
```diff
@@ -214,18 +214,25 @@ WebPlugin *makeWebPlugin(char *pluginLocation, JsonObject *pluginDefintion, Inte
...
-        /* Node services will be handled by node without ever going to the MVD server */
+      char *serviceName = jsonObjectGetString(serviceDef, "name");
+      if (!serviceName) {
+        // Return a null plugin since the dataservice definition is not correct.
+        printf("*** PANIC: Returning NULL for plugin. Check pluginDefinition for correct dataservice syntax ***\n");
+        plugin = NULL;
+        return plugin;
       } else {
-        /* import, ignore */
+        if (!type || !strcmp(type, "service")) {
```
1000TurquoisePogs: I think this area of the code could use a minor improvement:
if !type: issue a warning, return NULL
if group: as-is, not sure how this is used personally.
if import || router || nodeService, ignore
else print warning saying type unknown, but don't return NULL... since it may just be something the server can't handle but the nodejs server can instead.

sumanapp: This should be all set. Thanks.

c/dataservice.c
```diff
@@ -257,18 +274,21 @@ WebPlugin *makeWebPlugin(char *pluginLocation, JsonObject *pluginDefintion, Inte
...
-  httpService->authType = SERVICE_AUTH_NONE; /* TODO: this needs to be fleshed out a lot based on your specification */
-  registerHttpService(server, httpService);
-  httpService->userPointer = dataService;
+  int result;
+  HttpService *httpService = (HttpService*) NULL;
+  result = makeHttpDataServiceUrlMask(dataService, urlMask, sizeof(urlMask), server->defaultProductURLPrefix);
+  if (result != -1) {
+    printf("installing service %s at URI %s\n", dataService->identifier, urlMask);
+    HttpService *httpService = makeGeneratedService(dataService->identifier, urlMask);
+    httpService->authType = SERVICE_AUTH_NONE; /* TODO: this needs to be fleshed out a lot based on your specification */
```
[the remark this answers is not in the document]
1000TurquoisePogs: Yes - I guess this got lost as a result of independent pull requests. Will need this ON.

c/dataservice.c
```diff
@@ -214,20 +214,40 @@ WebPlugin *makeWebPlugin(char *pluginLocation, JsonObject *pluginDefintion, Inte
...
-      if (!type || !strcmp(type, "service")) {
+      char *serviceName = jsonObjectGetString(serviceDef, "name");
+      char *sourceName = jsonObjectGetString(serviceDef, "sourceName");
+      bool isImport = strcmp(type, "import") ? false : true;
+      if (!type) {
+        printf("*** PANIC: Returning NULL for plugin. Check pluginDefinition for correct 'type' fields on dataservices. ***\n");
+        plugin = NULL;
+        safeFree((char*)plugin,sizeof(WebPlugin));
+        return NULL;
+      } else if (!isImport && !serviceName) {
```
ifakhrutdinov: It seems like in the first 3 ```if``` statements you're trying to validate the input and they could be isolated in their own validate function, then you'd have a simple code like this:
```c
...
/* validate */
if (!isServiceDefinionValid(serviceDef)) {
  safeFree((char *)plugin, sizeof(WebPlugin));
  plugin = NULL;
  return NULL;
}

/* determine the type */
...
````

ifakhrutdinov:
In general I don't see obvious bugs, but I have some comments (I may be nitpicking):

* The formatting should be consistent, if one new function uses spaces between the arguments, all should do the same
* We should be using ```zowelog``` instead of naked printfs (especially for debug and info messages)
* The length of lines should be 120 chars at most (80 is even better)
* Please see my comment https://github.com/zowe/zowe-common-c/pull/4#discussion_r255064573, we could make the validation logic more obvious and have just a single free call if the validation function fails

Compile test and smoke tests were successful. @1000TurquoisePogs please merge if the above items are okay.

PR #5, opened 2019-02-01
Const and security

1000TurquoisePogs:
New services were being made as insecure unless otherwise stated, so this does the opposite: opt-out rather than opt-in.

lastIndexOf wasnt working for me because I had a const char. Shouldn't it be a const?


ifakhrutdinov:
@1000TurquoisePogs Did you get a compiler warning message? You can either: change your string to be non-const (bad solution), cast to the non-const char (also bad), make the function accept const strings (I think this is preferable, as if your function isn't intending to modify the value we should let the caller/devs know). 

Could you also change the security type of rasService?

PR #6, opened 2019-02-01
Discard LPA module if a different version has been detected.

ifakhrutdinov:
It is unsafe to call CSVDYLPA DELETE. This feature will only be
available in dev mode. In release mode, we'll never attempt to
delete the server module from LPA, and will just abandon it if
the module timestamp doesn't match the timestamp of the private
storage module.

Signed-off-by: Irek Fakhrutdinov <ifakhrutdinov@rocketsoftware.com>

### DIFF
--- a/c/crossmemory.c
+++ b/c/crossmemory.c
@@ -2945,52 +2945,51 @@ static int allocateGlobalResources(CrossMemoryServer *server) {
     moduleAddressLPA =
         globalArea->lpaModuleInfo.outputInfo.stuff.successInfo.loadPointAddr;
 
-#ifdef CMS_LPA_DEV_MODE
-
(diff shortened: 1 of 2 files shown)


fkovinAtRocket:
I took a look. I don't see any obvious problems like leaks, wild pointers, etc, but I know little about this LPA thing and what exactly is going on.

ifakhrutdinov:
> 
> 
> The current code does not perform the timestamp checking logic when compiled in development mode, which means you are not testing it in development mode and should be. The code should always do the check and accept the current module if it succeeds. Only if it fails should the code behave differently in development mode.

You logic makes more sense. Please see the update. Now the only difference between the LPA dev mode and release mode is module removal, the rest of the logic is performed in both modes.

PR #7, opened 2019-02-04
Remove use of ASCII as EBCDIC; fix unterminated string handling

ifakhrutdinov:
1. httpserver uses an ASCII string as an EBCDIC string
2. Multiple bugfixes in csi code:
    -  No need to allocate DSCB1 buffers on the heap
    -  No need to check a struct field for NULL (entry->name) - it can't be NULL
    -  Use jsonAddUnterminatedString for space padded unterminated strings
    -  Fix a double free - dsn pointer derived from either an SLH or a local buffer


ifakhrutdinov:
@gadyck I agree. I think any data structure that has more than one internal resource, needs its own "constructor" and "destructor" functions. There is definitely room for improvement in this code.

PR #8, opened 2019-02-06
Warn if need TLS and TLS isnt being used

1000TurquoisePogs:
Because the ZSS server doesnt implement TLS of its own, if we determine that we're in a case where it is important (say, listening on external interface), then we want to alert the admin if TLS was needed but not being used. When using AT-TLS, this warning should not be seen.


1000TurquoisePogs:
What you'll see if this flag is set:
`*** WARNING: Connection is insecure! TLS needed but not found on socket. ***` ... on every single request. It's intentionally very chatty to discourage this for all but development.

PR #10, opened 2019-02-15
Make stcbase.h and socketmgmt.h self-contained

ifakhrutdinov:
While working on a plugin, it was discovered that ```stcbase.h``` and ```socketmgmt.h``` could not be included by themselves. This fix adds includes headers so that ```stcbase.h``` and ```socketmgmt.h``` don't depend on the order they're included or on what other header are included (or not) before them.

### DIFF
--- a/h/socketmgmt.h
+++ b/h/socketmgmt.h
@@ -13,6 +13,8 @@
 #ifndef __SOCKETMGMT__
 #define __SOCKETMGMT__ 1
 
+#include "bpxnet.h"
+#include "utils.h"
 
 typedef struct SXChainElement_tag{
   char *data;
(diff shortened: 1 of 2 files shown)


1000TurquoisePogs:
Did this come from review with @jordanfilteau1995 ? I recall when he was making a zss plugin that header order mattered. Perhaps this is a fix for that.

ifakhrutdinov:
At some point ```void wsSendError(WSSession *session, char *error);``` got included in ```h/serviceUtils.h```, which brought ```#include "httpserver.h"``` and the serviceUtils header started affecting the files it was included in. Those files then needed to include everything httpserver.h depended on. I fixed the headers that affected httpserver.h. There are more headers like that but we can fix them as needed.

PR #12, opened 2019-02-15
Add uint64 support to json

ifakhrutdinov:


ifakhrutdinov:
There is no point in adding uint64, since we'll be losing precision with numbers < INT64_MAX. Closing the PR. I'll stick to the exiting int64 function.

PR #14, opened 2019-02-20
Add a core service that returns status information.

ifakhrutdinov:
The pull-request adds a cross-memory core service that can be used to retrieve the server status. This initial version includes basic information only (call RC, description).

Other changes:
* Check the server name for NULL in cmsCallService
* Comment about a flaw in the cmsGetGlobalArea implementation

**It is a prerequisite for https://github.com/zowe/zss/issues/8**


ifakhrutdinov:
@1000TurquoisePogs We could add a version field to the status struct but then it won't always be valid because the server may be just missing so no version information can be extracted. Another option is to actually provide the version value inside the description message when a version mismatch is detected. The current implementation provides the same amount of information as before minus that SAF call.

ifakhrutdinov:
@1000TurquoisePogs Speaking of different versions, the error mismatch reporting isn't very good right now. The function that locates the server simply skips servers with non matching versions. In a situation when the caller has version 2 and the server has version 3, it'll look as if no server has been found. Changing that behaviour is a bit out of scope of this PR. Here's the change b180f997bcb91d6f29ebd9a18ae14bed1a142354

hogstrom:
Sorry I'm late to the party.  I approve as well.

PR #15, opened 2019-02-22
Add convenience function to roll up headers used for REST JSON responses

1000TurquoisePogs:
Implements https://github.com/zowe/zss/issues/11
Accomplished by adding a utility function for calling to set up some default headers. I put this in dataservice.h since its most likely to be used by dataservices. Couldn't think of a better place.

Aside from adding headers instructing receivers not to cache the results, this change also unifies the content type of json. The majority of our code used MIME "text/json", but the official MIME type is "application/json", so hopefully this improves compatibility in edge cases.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3801,10 +3801,12 @@ void respondWithUnixFileNotFound(HttpResponse* response, int jsonMode) {
 void respondWithJsonError(HttpResponse *response, char *error, int statusCode, char *statusMessage) {
   jsonPrinter *out = respondWithJsonPrinter(response);
 
-  setContentType(response, "text/json");
+  setContentType(response, "application/json");
   setResponseStatus(response,statusCode,statusMessage);
   addStringHeader(response, "Server", "jdmfws");
   addStringHeader(response, "Transfer-Encoding", "chunked");
+  addStringHeader(response, "Cache-control", "no-store");
+  addStringHeader(response, "Pragma", "no-cache");
   writeHeader(response);
 
   jsonStart(out);
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3804,6 +3804,13 @@ void respondWithUnixFileNotFound(HttpResponse* response, int jsonMode) {
   }
 }
 
+void setDefaultJSONRESTHeaders(HttpResponse *response) {
+  setContentType(response, "application/json");
+  addStringHeader(response, "Server", "jdmfws");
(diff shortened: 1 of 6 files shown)


ifakhrutdinov:
> Accomplished by adding a utility function for calling to set up some default headers. I put this in dataservice.h since its most likely to be used by dataservices. Couldn't think of a better place.

I don't think all the code that wants to return JSON should depend on dataservice.

What about httpserver? It already has general functions for setting up a response, this will be a concrete implementation for JSON.

c/httpserver.c
```diff
@@ -3800,11 +3808,8 @@ void respondWithUnixFileNotFound(HttpResponse* response, int jsonMode) {
 // Response must ALWAYS be finished on return
 void respondWithJsonError(HttpResponse *response, char *error, int statusCode, char *statusMessage) {
   jsonPrinter *out = respondWithJsonPrinter(response);
-
-  setContentType(response, "text/json");
   setResponseStatus(response,statusCode,statusMessage);
-  addStringHeader(response, "Server", "jdmfws");
-  addStringHeader(response, "Transfer-Encoding", "chunked");
+  setDefaultJSONRESTHeaders(response);
```
ifakhrutdinov: 1. httpserver.c doesn't include dataservice.h thus doesn't have the correct declaration of that new function. This leads to subtle bugs because to the compiler this call will be equivalent to ```int foo(int var)```.
2. httpserver.c, which, I think, shouldn't include dataservice.h will now have to include it. And anyone who wants to use it will have to link with dataservice.c.

1000TurquoisePogs: So, do you recommend putting setDefaultJSONRESTHeaders into httpserver.c & httpserver.h directly? Was trying to avoid bloating this file, but it seems similar functions are already there.

ifakhrutdinov: For now yes, because otherwise we make the code more intertwined. I'd make some httputils.c in the future.

h/dataservice.h
```diff
@@ -83,6 +83,8 @@ void initalizeWebPlugin(WebPlugin *plugin, HttpServer *server);
 HttpService *makeHttpDataService(DataService *dataService, HttpServer *server);
 int makeHttpDataServiceUrlMask(DataService *dataService, char *urlMaskBuffer, int urlMaskBufferSize, char *productPrefix);
 
+
```
ifakhrutdinov: I'd remove these so we don't push empty changes. Same for dataservice.c

PR #16, opened 2019-03-02
Changes for ZSS Cross-Memory server plugins

ifakhrutdinov:
These are the zowe-common-c changes for https://github.com/zowe/zss/issues/13

**Must be merged with https://github.com/zowe/zss/pull/14**

Major changes:
* Refactor crossmemory.c
  * Add LPA wrappers
  * Add ISGENQ wrappers
  * Add a source file for various cross-memory utility functions (cross memory memcpy)
* Add a common storage map (for the plugin service table)
* Add callbacks to the cross-memory server 
  * Call the init callback function when the server has initialized
  * Call the term callback function when the server has started
termination
  * Call the modify command callback when the server receives a modify
command
  * All modify commands are processed in their own sub-tasks


ifakhrutdinov:
@hogstrom, One change we may want to revert is bumping the version of the cross-memory server's global area (it was changed from 2 to 3). If someone runs older ZSS, they won't be able to call a newer server. Making it 3 was not necessary as you can still use version 2 of the global are. Do we expect people run older ZSSes with the new version or do people usually update everything?

This should eventually be fixed in https://github.com/zowe/zowe-common-c/issues/17, but that's not in this PR.

ifakhrutdinov:
@gadyck 
Actually the code makes a copy by de-referencing that pointer.
```

                                             3322 *   below2G->header = *header;                                           000398
 0000135C E3E0 D148 0004          00000148   3323          LG    14,328(0,13)            #SR_PARM_8                        000398
                     R:E 00000000            3324          USING @@PARMD@8,14                                              000398
 00001362 E3F0 E028 0004          00000028   3325          LG    15,@194header                                             000398
 00001368 E3E0 D110 0004          00000110   3326          LG    14,@197below2G                                            000398
 0000136E D217 E040 F000 00000040 00000000   3327          MVC   64(24,14),0(15) 
```
So the header we pass will have 24 bytes below the bar.

PR #23, opened 2019-03-15
Add "mode=base64" option to unixfile API

kbeausoleil:
Added new default behavior to /unixfile/contents/ GET calls. Base64 is the new default. Old behavior can be seen with "mode=raw" parameter.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -1420,6 +1420,7 @@ static int decodeSessionToken(ShortLivedHeap *slh,
 HttpServer *makeHttpServer2(STCBase *base,
                            InetAddr *addr,
                            int port,
+                           int tlsFlags,
                            int *returnCode, int *reasonCode){
 
   SessionTokenKey sessionTokenKey = {0};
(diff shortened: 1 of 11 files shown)


c/httpserver.c
```diff
@@ -3814,32 +3835,50 @@ void respondWithJsonError(HttpResponse *response, char *error, int statusCode, c
...
 }
 
-int streamBinaryForFile(Socket *socket, UnixFile *in){
+#define ENCODE64_SIZE(SZ) (2 + 4 * ((SZ + 2) / 3))
+
+int streamBinaryForFile(Socket *socket, UnixFile *in, bool asB64) {
   int returnCode = 0;
   int reasonCode = 0;
-  char buffer[FILE_STREAM_BUFFER_SIZE+4];
+  int bufferSize = 3*FILE_STREAM_BUFFER_SIZE;
```
1000TurquoisePogs: I'm not aware of if we need to observe a size limit here, but it seems like getting the buffer divisible by 3 by tripling it may be better done by adding the remainder to make it divisible by 3.

1000TurquoisePogs:
https://github.com/zowe/zowe-common-c/pull/22 states:
DO NOT MERGE THIS UNTIL THE FOLLOWING IS RESOLVED

File Transfer App needs to ask for "mode=raw" to get raw bytes again.
File Editor needs to Base64 Decode the byte stream before displaying on the screen.
Other users of /unixfile/ API need to be modified similarly.


... I don't see these PRs yet, so I'm not merging so that nothing becomes broken.

NolanRogers:
This is tagged for 1.1.1 and on the zlux waffle assigned to @reetchow

PR #26, opened 2019-03-27
Update the z/OS version check to support z/OS 2.4

ifakhrutdinov:
There is an OS version check that prevents the LE related code from running on an unsupported z/OS version. This protects the code from using CAA incorrectly in case it gets updates with the OS. The current version is 2.3, this pull-request updates the corresponding ```#define``` to use the 2.4 value.

Fixes zowe/zss#15

### DIFF
--- a/c/le.c
+++ b/c/le.c
@@ -107,7 +107,7 @@ char *getCAA(){
 }
 
 #ifndef LE_MAX_SUPPORTED_ZOS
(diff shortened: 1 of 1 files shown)


ifakhrutdinov:
@1000TurquoisePogs are we good to merge this? No code freeze or anything like that for the upcoming release?

dmitrynikolaev:
The PR has been successfully tested.

PR #28, opened 2019-04-04
No longer asssume the file is text if it is untagged.

jordanfilteau1995:
Currently, if the file is untagged in USS, then the server will try to stream it as text. This can result in a conversion error which causes unexpected behavior, since the file could actually be binary. The server will now stream the file in binary if the file is untagged.


1000TurquoisePogs:
This seems good, but just needs a bit of testing first.
The idea is that if something's untagged, we don't have a good idea of if it is text, or if it is binary, so we shouldn't try to convert if it's just going to end up throwing some garbage if it is indeed a binary file.

PR #30, opened 2019-04-23
Add BPXPOL to open-source library; poll on socket before BPXWRT

jordanfilteau1995:
This should help the performance of downloading files by eliminating the constant failed BPXWRT calls when attemping to write to a full socket.

### DIFF
--- a/c/http.c
+++ b/c/http.c
@@ -50,6 +50,7 @@
 #include "le.h"
 #include "scheduling.h"
 #include "socketmgmt.h"
+#include "fdpoll.h"
 
 #include "http.h"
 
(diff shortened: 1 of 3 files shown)


jordanfilteau1995:
@1000TurquoisePogs

c/http.c
```diff
@@ -200,11 +203,32 @@ int writeFully(Socket *socket, char *buffer, int len){
       if (WRITE_FORCE && returnCode == EAGAIN) {
 #else
       if (WRITE_FORCE && returnCode == 1102) {
```
rocketjared: We should be conditionally define'ing EAGAIN to 1102 on platforms where that's reasonable instead of hardcoding a magic value here.

PR #29, opened 2019-04-19
STCK and ENQ fixes

fkovinAtRocket:
- ENQ:
-- add IS_ISGENQ_LOCK_OBTAINED to isgenq.c
-- Workaround for a bug in the Metal headers with stddef.h
- STCK:
-- Fix variable passing in getSTCK() and getSTCK()

PR #32, opened 2019-04-29
Fix variable passing in getSTCK() and getSTCKU()

fkovinAtRocket:
The output variable was passed to the asm block as an input variable
instead, thus with some compile options (not always!) the function
failed to produce an output.

Second part of what used to be https://github.com/zowe/zowe-common-c/pull/29


ifakhrutdinov:
I will need to verify the change, this is used in some of our code.

PR #33, opened 2019-04-29
ISGENQ wrapper fixes/improvements:

fkovinAtRocket:
- IS_ISGENQ_LOCK_OBTAINED macro
- Workaround for a bug in the Metal headers

First part of what used to be https://github.com/zowe/zowe-common-c/pull/29


ifakhrutdinov:
From PR #29 

> Is $ allowed by the C standard or is it an extension?
>> $ is a GCC extension also supported by XL C.

PR #36, opened 2019-05-09
Added a side file for external plugin developers.

jordanfilteau1995:
I added on to the previous side file before source code was on GitHub. I added several more utility functions which will allow the community to build external plugins without having to statically link every file, as well as gives a list of core functions that should be used when building the data service. A wiki page should follow this, which should give an example. I included the script to generate the side file in case anyone wants to add anything that is needed for their project, but the side file should be added to and not removed from.

I tested the build process on a data service and it builds. It also works as expected.


jordanfilteau1995:
@1000TurquoisePogs

PR #39, opened 2019-05-29
DLL's can now reside in respective plugin folders.

jordanfilteau1995:

### DIFF
--- a/c/dataservice.c
+++ b/c/dataservice.c
@@ -165,9 +165,21 @@ static DataService *makeDataService(WebPlugin *plugin, JsonObject *serviceJsonOb
     printf("*** PANIC ***  service loading in METTLE not implemented\n");
 #else
     char dllLocation[1023] = {0}; /* USS max path length */
-    snprintf(dllLocation, sizeof(dllLocation), "%s/lib/%s", plugin->pluginLocation, libraryName);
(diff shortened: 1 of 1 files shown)


c/dataservice.c
```diff
@@ -164,8 +164,10 @@ static DataService *makeDataService(WebPlugin *plugin, JsonObject *serviceJsonOb
     service->initializer = NULL;
     printf("*** PANIC ***  service loading in METTLE not implemented\n");
 #else
-    printf("going for DLL EP lib=%s epName=%s\n",libraryName,initializerName);
-    service->initializer = (InitializerAPI*) lookupDLLEntryPoint(libraryName,initializerName);
+    char dllLocation[1023] = {0}; /* USS max path length */
+    snprintf(dllLocation, sizeof(dllLocation), "%s/%s", plugin->pluginLocation, libraryName);
```
1000TurquoisePogs: Right idea, but wrong path.
pluginLocation = root folder.
Server logic goes into root/lib
See https://github.com/zowe/zlux/wiki/ZLUX-App-filesystem-structure#lib for what we want to follow.


Additionally, plugin->pluginLocation is sadly not santized: https://github.com/zowe/zss/blob/master/c/zss.c#L443
It may or may not already have a trailing slash.

jordanfilteau1995: Change has been made.

PR #40, opened 2019-06-05
Dataset eof bug

chcaron01:
While reading in lines from the dataset, the function streamDataset() did not check if an entry was of size 0 (indicating end of file). By adding a conditional check, the extra blank line is no longer printed.

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -90,7 +90,6 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter 
         contentLength = contentLength + bytesRead;
       }
       else if (bytesRead == 0){
-        printf("Reached end of dataset\n");
         break;
       }
       else {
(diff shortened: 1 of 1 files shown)


chcaron01:
@1000TurquoisePogs @jordanfilteau1995

jordanfilteau1995:
Hi @1000TurquoisePogs,

I saw the demo of these changes in person. It fixes the problem. I don't see anything wrong with it besides that we need to eventually migrate all the printf to zowelog, which is out of scope. I had him try a few different kinds of datasets. Do you see any issues with this? You know this API much better than I do.

c/datasetjson.c
```diff
@@ -79,22 +79,26 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter
...
-          printf("Error reading DSN=%s, rc=%d\n",filename,bytesRead);
-        }
-        else{
-          contentLength = contentLength + bytesRead;
-        }
+        contentLength = contentLength + bytesRead;
       }
-      fclose(in);
+      else if (bytesRead == 0){
+        printf("Reached end of dataset\n");
```
1000TurquoisePogs: Since we didn't use any way to differentiate between debug & non debug messages, please remove this printf entirely. It is meaningless to the end user and they will see zillions of them.

chcaron01: I've updated this in the latest commit

PR #42, opened 2019-06-05
Added function to not follow symbolic links

elizabethkirschner:
Added function to not follow symbolic link. Called "symbolicFileInfo"


elizabethkirschner:
@jordanfilteau1995

jordanfilteau1995:
Hi @1000TurquoisePogs,

This will solve the problem with whether or not we choose to follow symbolic links or not. Another syscall "lstat" was implemented and is now used in "makeJSONForDirectory".

1000TurquoisePogs:
I see a  lot of duplication in zosfile, so we should consider if it's worth simplifying this at some point in the future.

PR #44, opened 2019-06-07
Bugfix forbidden returns 200 OK

elizabethkirschner:
Calling get on a directory the user does not have permission for returns the correct status and error message.

Signed-off-by: E Kirschner <ekirschner@rocketsoftware.com>

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3761,30 +3761,43 @@ void respondWithUnixFile(HttpResponse* response, char* absolutePath, int jsonMod
 
 // Response must ALWAYS be finished on return
 void respondWithUnixDirectory(HttpResponse *response, char* absolutePath, int jsonMode) {
+  int returnCode;
+  int reasonCode;
+    UnixFile *directory = NULL;
+
+  
 #ifdef DEBUG
   printf("Directory case: %s\n",absolutePath);
   fflush(stdout);
 #endif
-  setResponseStatus(response,200,"OK");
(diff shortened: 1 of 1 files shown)


elizabethkirschner:
@jordanfilteau1995

c/httpserver.c
```diff
@@ -3761,30 +3761,43 @@ void respondWithUnixFile(HttpResponse* response, char* absolutePath, int jsonMod
...
-  if (jsonMode == 0) {
-    setContentType(response,"text/html");
-    writeHeader(response);
-    StringListElt *parsedFileTail = firstStringListElt(response->request->parsedFile);
-    while (parsedFileTail->next){
-      parsedFileTail = parsedFileTail->next;
-    }
-    makeHTMLForDirectory(response, absolutePath, parsedFileTail->string,TRUE);
+  
+ if (directory = (directoryOpen(absolutePath,&returnCode,&reasonCode)) == NULL){
```
ifakhrutdinov: * This produces a warning message
```
[exec] WARNING CCN3068 /u/tsirf/zowe-secmgmt/zowe-common-c/c/httpserver.c:3774  Operation between types "struct UnixFileStream_tag*" and "int" is not allowed.
```
* The if condition seems to have a bug - first we compare the result of ```dorectoryOpen``` with NULL and then assign the result (either 1 or 0 depending on the result) to ```directory```. I think you meant the following:
```
if ((directory = directoryOpen(absolutePath,&returnCode,&reasonCode)) == NULL){
```

PR #48, opened 2019-06-27
PDS members can be created and deleted now

chcaron01:
@jordanfilteau1995 @ifakhrutdinov

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -90,7 +90,6 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter 
         contentLength = contentLength + bytesRead;
       }
       else if (bytesRead == 0){
-        printf("Reached end of dataset\n");
         break;
       }
       else {
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -53,6 +53,7 @@ static int defaultVSAMCSIFieldCount = 4;
 
 static char getRecordLengthType(char *dscb);
 static int getMaxRecordLength(char *dscb);
+static int getDSCB(char* datasetName, char* dscb, int bufferSize);
 
 typedef struct DatasetName_tag {
   char value[44]; /* space-padded */
@@ -326,7 +327,6 @@ static void addDetailsFromDSCB(char *dscb, jsonPrinter *jPrinter, int *isPDS) {
 static int isPartionedDataset(char *dscb) {
   int posOffset = 44;
   int dsorgHigh = dscb[82-posOffset];
-
   if (dsorgHigh & 0x02) {
     return TRUE;
   }
@@ -367,21 +367,19 @@ static int obtainDSCB1(const char *dsname, unsigned int dsnameLength,
       } parmList;
     )
   );
-
+  
   memset(mem31->dsnameSpacePadded, ' ', sizeof(mem31->dsnameSpacePadded));
   memcpy(mem31->dsnameSpacePadded, dsname, dsnameLength);
   memset(mem31->volserSpacePadded, ' ', sizeof(mem31->volserSpacePadded));
   memcpy(mem31->volserSpacePadded, volser, volserLength);
 
   memset(mem31->workArea, 0, sizeof(mem31->workArea));
-
   mem31->parmList.operationCode = SVC27_OPCODE_SEARCH;
   mem31->parmList.option = SVC27_OPTION_EADSCB_OK;
   mem31->parmList.dscbNumber = 0;
   mem31->parmList.dsname = mem31->dsnameSpacePadded;
   mem31->parmList.volser =  mem31->volserSpacePadded;
   mem31->parmList.workArea = mem31->workArea;
-
   int rc = 0;
 
   __asm(
@@ -434,7 +432,7 @@ void addMemberedDatasetMetadata(char *datasetName, int nameLength,
                                 char *memberQuery, int memberLength,
                                 jsonPrinter *jPrinter,
                                 int includeUnprintable) {
-
+                                    
   int isPDS = FALSE;
   char dscb[INDEXED_DSCB] = {0};
   int rc = obtainDSCB1(datasetName, nameLength,
@@ -453,7 +451,6 @@ void addMemberedDatasetMetadata(char *datasetName, int nameLength,
    * memory.
    */
   isPDS = isPartionedDataset(dscb);
-
   /* If it's not a PDS, we exit. */
   if (!isPDS) {
     printf("Cannot print members. Selected dataset is not a PDS.");
@@ -513,6 +510,7 @@ void addMemberedDatasetMetadata(char *datasetName, int nameLength,
 
 #ifdef __ZOWE_OS_ZOS
 static void updateDatasetWithJSON(HttpResponse *response, JsonObject *json, char *datasetPath) {
+
   JsonArray *recordArray = jsonObjectGetArray(json,"records");
   int recordCount = jsonArrayGetCount(recordArray);
   int maxRecordLength = 80;
@@ -533,13 +531,15 @@ static void updateDatasetWithJSON(HttpResponse *response, JsonObject *json, char
   else{
     memcpy(dsn.value,datasetPath+3,datasetPathLen-4);
   }
+  
   int volserSuccess = getVolserForDataset(&dsn, &volser);
   if (!volserSuccess){
-    
+        
     char dscb[INDEXED_DSCB] = {0};
     int rc = obtainDSCB1(dsn.value, sizeof(dsn.value),
                          volser.value, sizeof(volser.value),
                          dscb);
+                         
     if (rc == 0){
       if (DSCB_TRACE){
         printf("DSCB for %.*s found\n", sizeof(dsn.value), dsn.value);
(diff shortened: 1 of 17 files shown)


c/datasetjson.c
```diff
@@ -53,6 +53,7 @@ static int defaultVSAMCSIFieldCount = 4;
 
 static char getRecordLengthType(char *dscb);
 static int getMaxRecordLength(char *dscb);
+static int prepareDatasetMember(HttpResponse* response, char* datasetName);
```
jordanfilteau1995: This function doesn't exist anymore. The prototype should be changed.

c/datasetjson.c
```diff
@@ -367,21 +367,20 @@ static int obtainDSCB1(const char *dsname, unsigned int dsnameLength,
       } parmList;
     )
   );
-
```
jordanfilteau1995: There's a lot of random white space. This should be avoided if possible. I'm not super picky about it, but others are.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
   finishResponse(response);     
 #endif /* __ZOWE_OS_ZOS */
 }
+//new function
+
+static int getDSCB(char* datasetName, char* dscb, int* dscbLength){
```
jordanfilteau1995: The point of having the length was to do a check to make sure the caller didn't call it with a bad array size.

`static int getDSCB(char *datasetName, char *dscb, int dscbLength) {
   if (dscbLength < INDEX_DCSB) {
     zowelog("");
     respondWithJsonError(internal server error);
     return 0;
   }
}`

Something like this.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+    
+  int volserSuccess = getVolserForDataset(&dsn, &volser);
+  
+  if(!volserSuccess){
+    
+    int rc = obtainDSCB1(dsn.value, sizeof(dsn.value),
+                           volser.value, sizeof(volser.value),
+                           dscb);
+                           
+    *dscbLength = sizeof(dscb);
```
jordanfilteau1995: Value never used. Also, no point in passing it. Also, the value of this will be the size of the pointer. You can't use the sizeof function on a passed in pointer. It's passed into this function, but also passed back from obtainDSCB1, so even though it has a size and data was copied, it's still a pointer.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+void newDatasetMember(HttpResponse* response, char* datasetPath, char* memberName) {
+  char dscb[INDEXED_DSCB] = {0};
+  int dscbLength = 0;
+  if (getDSCB(datasetPath, dscb, &dscbLength) == 0) {
+    if (isPartionedDataset(dscb)) {
+      char *overwriteParam = getQueryParam(response->request,"overwrite");
+      int overwrite = !strcmp(overwriteParam, "true") ? TRUE : FALSE;
+      char fullPath[DATASET_MEMBER_MAXLEN]; //Max dataset path length #define them
+      int apostrophePos = lastIndexOf(datasetPath, DATASET_PATH_MAX, '\''); //index of apostrophe
+      datasetPath[apostrophePos] = 0; // remove last apostrophe by null termination
```
jordanfilteau1995: We have a function in utils.c called nullTerminate. Please use that.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+      //concatenates dataset name with member name
+      snprintf(fullPath, DATASET_MEMBER_MAXLEN + 1, "%s(%s)'", datasetPath, memberName);
```
jordanfilteau1995: This is not safe. Please look at comment above about fullPath. You are adding DATASET_MEMBER_MAXLEN + 1, but there isn't enough space for that.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+}
+
+void newDatasetMember(HttpResponse* response, char* datasetPath, char* memberName) {
+  char dscb[INDEXED_DSCB] = {0};
+  int dscbLength = 0;
+  if (getDSCB(datasetPath, dscb, &dscbLength) == 0) {
+    if (isPartionedDataset(dscb)) {
+      char *overwriteParam = getQueryParam(response->request,"overwrite");
+      int overwrite = !strcmp(overwriteParam, "true") ? TRUE : FALSE;
+      char fullPath[DATASET_MEMBER_MAXLEN]; //Max dataset path length #define them
```
jordanfilteau1995: char fullPath[DATASET_MEMBER_MAXLEN + 1]; for the inevitable null terminator from snprintf.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+      int apostrophePos = lastIndexOf(datasetPath, DATASET_PATH_MAX, '\''); //index of apostrophe
+      datasetPath[apostrophePos] = 0; // remove last apostrophe by null termination
+      //concatenates dataset name with member name
+      snprintf(fullPath, DATASET_MEMBER_MAXLEN + 1, "%s(%s)'", datasetPath, memberName);
+      FILE* datasetExists = fopen(fullPath,"r");
+      if (datasetExists && overwrite != TRUE) {//Member already exists and overwrite wasn't specified
+        fclose(datasetExists);
+        respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Member already exists and overwrite wasn't specified");
```
jordanfilteau1995: respondWithJsonError is better. Please use that.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+      }
+      else { // Member doesn't exist
+        if (datasetExists) {
+          fclose(datasetExists);
+        }
+        FILE* newMember = fopen(fullPath, "w");
+        if (!newMember){
+          respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Bad dataset name");
```
jordanfilteau1995: See comment above about respondWithJsonError.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+          return;
+        }
+        if (fclose(newMember) == 0){
+          response200WithMessage(response, "Successfully created member");
+        }
+        else {
+          respondWithError(response, HTTP_STATUS_BAD_REQUEST, "Could not close dataset");
```
jordanfilteau1995: This requires a zowelog of some sort. This is really bad.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+        }
+      }
+    }
+    else {
```
jordanfilteau1995: I usually do something like:

bool isPDS = isPartionedDataset(dscb);
if (!isPDS) {
  return error
}

It's more obvious about bad conditions than finding them at the very end. That comes down to preference, but I thought I'd give my opinion on readability.

c/datasetjson.c
```diff
@@ -1671,7 +1669,117 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
   finishResponse(response);     
 #endif /* __ZOWE_OS_ZOS */
 }
+//new function
+
+static int getDSCB(char* datasetName, char* dscb, int* dscbLength){
+  Volser volser = {0};
+  DatasetName dsn = {0};
+  memset(dsn.value, ' ', 44);
```
jordanfilteau1995: That 44 can be the #define now.

c/datasetjson.c
```diff
@@ -1225,7 +1224,6 @@ void respondWithVSAMDataset(HttpResponse* response, char* absolutePath, hashtabl
     }
     safeFree(dsnUidPair, 44+8+1);
   } else { /* need to open an ACB */
-    printf("about to dynalloc a DD to %s\n", dsn);
```
jordanfilteau1995: It looks like you might be on the wrong branch or didnt do a git pull?

ifakhrutdinov:
@chcaron01 can you please squash the commits and put a description in the commit message?

c/datasetjson.c
jordanfilteau1995: Nevermind. Blue is partial commit I guess.

jordanfilteau1995:
Do not merge without @ifakhrutdinov approval.

chcaron01:
@ifakhrutdinov Hello, I think I've added commits I didn't mean to due to my inexperience with git. I'm going to create a new pull request with just the changes I want.

PR #49, opened 2019-06-28
Adds dynamic creation of logging id's for plugins

elizabethkirschner:
changes to zowe-common-c are only in data structures, see PR for zss for more details.


elizabethkirschner:
see zss PR #53
https://github.com/zowe/zss/pull/53

elizabethkirschner:
@jordanfilteau1995

h/dataservice.h
```diff
@@ -19,7 +19,12 @@
  *
  */
 
-
+//hash stuff
```
jordanfilteau1995: See comment about struct name on other PR.

h/httpserver.h
```diff
@@ -212,6 +212,7 @@ typedef struct HttpServer_tag{
   char             *defaultProductURLPrefix;
   uint64           serverInstanceUID;   /* may be something smart at some point. Now just startup STCK */
   void             *sharedServiceMem; /* address shared by all HttpServices */
+  hashtable        *loggingTable;
```
jordanfilteau1995: That should really be pluginLoggingTable. Variable names should be self documenting as much as possible.

jordanfilteau1995:
Also, this cannot be merged until @ifakhrutdinov approves.

h/httpserver.h
```diff
@@ -212,6 +212,7 @@ typedef struct HttpServer_tag{
   char             *defaultProductURLPrefix;
   uint64           serverInstanceUID;   /* may be something smart at some point. Now just startup STCK */
   void             *sharedServiceMem; /* address shared by all HttpServices */
+  hashtable        *PluginLoggingTable;
```
ifakhrutdinov: * A variable name should not start with a capital letter
* This doesn't belong in the http server. A plug-in is a ZSS feature, not a feature of the http server. You can make it a dynamic property in the http server. So don't need to make any changes in httpserver.c/h. See ```privilegedServerName``` in https://github.com/zowe/zss as a reference.

h/dataservice.h
ifakhrutdinov: What is ```//hash stuff``` for?

jordanfilteau1995: Yeah, the comment should be removed.

h/httpserver.h
jordanfilteau1995: As of now, the data service would take that hashtable and use it to get the logging component id associated with the plugin identifier. This can be moved to a dynamic property without an problems.

ifakhrutdinov:
@elizabethkirschner Can you please squash the commits?

PR #50, opened 2019-07-01
PDS-E member creation (squashed commits)

chcaron01:
Created functions that add/remove PDS members. Functions take valid PDS paths and member names as parameters and will add or remove a member to/from the function using fopen() and remove().

@ifakhrutdinov @jordanfilteau1995


c/datasetjson.c
```diff
@@ -1671,7 +1668,130 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+    }
+    else {
+      char *overwriteParam = getQueryParam(response->request,"overwrite");
+      int overwrite = !strcmp(overwriteParam, "true") ? TRUE : FALSE;
+      char fullPath[DATASET_MEMBER_MAXLEN + 1]; //Max dataset path length
+      int apostrophePos = lastIndexOf(datasetPath, DATASET_PATH_MAX, '\''); //index of apostrophe
+      nullTerminate(datasetPath, apostrophePos); // remove last apostrophe by null termination
+      //concatenates dataset name with member name
+      snprintf(fullPath, DATASET_MEMBER_MAXLEN, "%s(%s)'", datasetPath, memberName);
+      FILE* datasetExists = fopen(fullPath,"r");
```
jordanfilteau1995: You are really checking for the member here, right? We know the dataset exists because it has a dscb. This boolean-like name is a bit confusing considering what you are actually trying to achieve here.

I’m still concerned about the implications of using c standard functions “fopen”, “fclose”, etc. I don’t think they will compile in METAL C. Sean said it was OK to use them though.

c/datasetjson.c
```diff
@@ -1671,7 +1668,130 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+  if (getDSCB(datasetPath, dscb, bufferSize) != 0) {
+    respondWithJsonError(response, "Error decoding dataset", 400, "Bad Request");  
+  }
+  else {
+    if (!isPartionedDataset(dscb)) {
+      respondWithJsonError(response, "Dataset must be PDS/E", 400, "Bad Request");
+    }
+    else {
+      char fullPath[DATASET_MEMBER_MAXLEN + 1];
+      int apostrophePos = lastIndexOf(datasetPath, DATASET_PATH_MAX, '\''); //index of apostrophe
```
jordanfilteau1995: Same as above.

h/datasetjson.h
```diff
@@ -23,6 +21,9 @@
 
 #define SAF_AUTHORIZATION_READ 0x04
 #define SAF_AUTHORIZATION_UPDATE 0x08
+#define LOG_COMP_ID_MVD_SERVER 0x008F000300010000
```
jordanfilteau1995: This should not be here. This should be in “logging.h”. It should be something like, “LOG_COMP_ID_DATASETJSON” if it doesnt already exist.

h/datasetjson.h
```diff
@@ -23,6 +21,9 @@
...
+#define DATASET_PATH_MAX 44
+#define DATASET_MEMBER_MAXLEN 58
```
jordanfilteau1995: How’d you get 58? A dataset member has a max length of 8.

44+8 = 52

Even if you include the (), I’m still off by 4.

Regardless, if you need a max length of something made up by two different values, just add them together.

For example,

#define DS_NAME_MAX_LEN 44
#define DS_MEMBER_NAME_MAX_LEN 8

char fullPath[DS_NAME_MAX_LEN + DS_MEMBER_NAME_MAX_LEN + 2 + 1] /* +2 for () and +1 for null

If anyone wanted to use that constant, it would just be wrong.

You could also do 

#define DS_NAME_PDS_MAX_LEN DS_NAME_MAX_LEN + DS_MEMBER_NAME_MAX_LEN + 2

chcaron01: 58 come from 44 + 8 + ( ) + //' + '

h/logging.h
```diff
@@ -89,6 +89,7 @@ ZOWE_PRAGMA_PACK_RESET
 #define LOG_COMP_DATASERVICE   0x008F0001000B0000LLU
 #define LOG_COMP_CMS           0x008F0001000C0000LLU
 #define LOG_COMP_LPA           0x008F0001000D0000LLU
+#define LOG_COMP_ID_MVD_SERVER 0x008F000300010000
```
jordanfilteau1995: Use a different one.

PR #54, opened 2019-07-09
Make zowe-common-c compatible with new dataset API

chcaron01:
@jordanfilteau1995

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -53,9 +53,13 @@ static int defaultVSAMCSIFieldCount = 4;
 
 static char getRecordLengthType(char *dscb);
 static int getMaxRecordLength(char *dscb);
+static void respondWithNonVSAMDataset(HttpResponse* response, char* absolutePath, int jsonMode);
+static void respondWithVSAMDataset(HttpResponse* response, char* absolutePath, hashtable *acbTable, int jsonMode);
+static void updateNonVSAMDataset(HttpResponse* response, char* absolutePath, int jsonMode);
+static void updateVSAMDataset(HttpResponse* response, char* absolutePath, hashtable *acbTable, int jsonMode);
 
 typedef struct DatasetName_tag {
-  char value[44]; /* space-padded */
+  char value[DATASET_NAME_LEN]; /* space-padded */
 } DatasetName;
 
 typedef struct Volser_tag {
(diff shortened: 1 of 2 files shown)


chcaron01:
Fixed dataset metadata so it can handle new, longer endpoint

c/datasetjson.c
```diff
@@ -335,6 +339,28 @@ static int isPartionedDataset(char *dscb) {
 }
 #endif
 
+static int checkForVSAM(char* dsPath) {
```
jordanfilteau1995: Please, use bool instead of int, since it returns TRUE or FALSE. Also, the function name should ideally be boolean-esque such as, isDatasetVSAM().

c/datasetjson.c
```diff
@@ -335,6 +339,28 @@ static int isPartionedDataset(char *dscb) {
...
+  char dscb[INDEXED_DSCB] = {0};
+  Volser volser = {0};
+  DatasetName dsn = {0};
+  memset(dsn.value, ' ', DATASET_PATH_MAX);
```
jordanfilteau1995: Just with padWithSpaces()

c/datasetjson.c
```diff
@@ -335,6 +339,28 @@ static int isPartionedDataset(char *dscb) {
...
+  memcpy(dsn.value,dsPath+3,strlen(dsPath)-4);
+  int volserSuccess = getVolserForDataset(&dsn, &volser);
+
+  if(!volserSuccess){
+
+    int rc = obtainDSCB1(dsn.value, sizeof(dsn.value),
```
jordanfilteau1995: Check the return code.

c/datasetjson.c
```diff
@@ -964,8 +990,30 @@ void updateVSAMDataset(HttpResponse* response, char* absolutePath, hashtable *ac
   read = openSAM(name,OPEN_CLOSE_INPUT,TRUE,recfm,lrecl,blksize);
 
  */
+void respondWithVSAMorDataset(HttpResponse* response, char* dsName, int jsonMode, HttpService* service){
```
jordanfilteau1995: VSAM is a dataset. This name is misleading. Please find an alternative.

jordanfilteau1995:
API is fully functional. I see no reason why it can't be merged.

PR #57, opened 2019-07-17
Multipart string support in the json printer

fkovinAtRocket:
Signed-off-by: Fyodor Kovin <fkovin@rocketsoftware.com>

### DIFF
--- a/c/json.c
+++ b/c/json.c
@@ -607,6 +607,52 @@ void jsonAddJSONString(jsonPrinter *p, char *keyOrNull, char *validJsonValue) {
   jsonWrite(p, validJsonValue, false, p->inputCCSID);
 }
 
+void jsonStartMultipartString(jsonPrinter *p, char *keyOrNull) {
+  if (jsonShouldStopWriting(p)) {
+    return;
+  }
+  if (p->isFirstLine) {
+    p->isFirstLine = FALSE;
+  } else {
+    jsonNewLine(p);
+  }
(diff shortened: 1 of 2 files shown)


fkovinAtRocket:
@lchudinov  can you please take a look?

lchudinov:
Looks good to me.

1000TurquoisePogs:
What is the intent of this?
The new functions do a sanity check of if you are in the multi-part string state or not, but all old functions do not do this check. Obviously the json printer needs to be used deliberately but I was just wondering if there was more stupid-proofing that could be done here or if it was intentional that the other functions become part of the multi-part string.

fkovinAtRocket:
The intent is actually forming a single JSON string out of separate string pieces that  come from a complex data structure returned by a callable service: some of them return big arrays of length-prefixed strings, each correspoinding to a line of output, and we just need to gather the entire response in one plain string. There can be lots of data, so it's nice to do this without extra copying.

Adding that sanity check to every function would be too massive a change. I thought that this is going to be an exotic feature, so most users won't invoke these function anyway, and thus the existing functions shouldn't be changed for the sake of this new feature. Also, the error reporing capabilities of the printer API are very limited: all functions are `void something()`, so there's little we can do anyway, like `return -1`. Most callers will just continue printing stuff, even when things are clearly wrong.

Also, a subtle detail: if any of the existing functions is changed, any plugin that's been compiled without  the side file yet will still use the old functions anyway.

1000TurquoisePogs:
Fine to me.

PR #58, opened 2019-07-18
Dynaically grow the JSON tokenizer buffer

fkovinAtRocket:
The buffer used to be fixed size. Now it grows dynamically up to a
compile time specified limit.

### DIFF
--- a/c/json.c
+++ b/c/json.c
@@ -182,20 +182,6 @@ void jsonWriteBufferInternal(jsonPrinter *p, char *text, int len) {
   }
 }
 
-void *safeRealloc(void *ptr, size_t size, size_t oldSize, char *site) {
-  void *new;
-
-  new = safeMalloc(size, site);
-  if (new == NULL) {
-    return NULL;
-  }
-  if ((ptr != NULL) && (oldSize > 0)) {
-    memmove(new, ptr, oldSize);
-    safeFree(ptr, oldSize);
-  }
-  return new;
-}
-
 #define MAX($a, $b) ((($a) > ($b))? ($a) : ($b))
 
 #define ESCAPE_LEN 6 /* \u0123 */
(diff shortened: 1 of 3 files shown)


fkovinAtRocket:
@lchudinov please review :)

c/json.c
```diff
@@ -953,6 +957,25 @@ void jsonParseFail(JsonParser *parser, char *formatString, ...) {
...
+static int
+jsonTokenizerGrowBuffer(JsonTokenizer *t) {
+  if (t->bufferSize >= INT32_MAX / 2) {
+    return -1;
+  }
+  int newSize = 2 * t->bufferSize;
+  if (newSize >= JSON_TOKEN_BUFFER_SIZE_LIMIT) {
+    return -1;
+  }
+  void *newBuffer = safeRealloc(t->buffer, newSize,
```
1000TurquoisePogs: Surprisingly safeRealloc is within this file. It seems to be completely generic so it should be moved to alloc.c

c/json.c
```diff
@@ -1020,36 +1042,46 @@ JsonToken *getStringToken(JsonTokenizer *tokenizer) {
...
+      }
     } else {
       int c = jsonTokenizerRead(tokenizer);
-      if (isprint(c) && pos < tokenizer->bufferSize) {
-        buffer[pos++] = c;
+      if (isprint(c)) {
+        if (pos >= tokenizer->bufferSize) {
+          int growRc = jsonTokenizerGrowBuffer(tokenizer);
+          if (growRc < 0) {
+            break;
```
lchudinov: I would ask you to introduce a new token type denoting a too long string. The token would help JSON parser to issue a correct error message.

c/json.c
fkovinAtRocket: done

c/json.c
fkovinAtRocket: done

PR #65, opened 2019-07-29
Remove dataset

chcaron01:
Functionality for removing datasets.

Signed-off-by: Charlie Caron <ccaron@rocketsoftware.com>


c/dynalloc.c
```diff
@@ -959,7 +959,60 @@ int unallocDataset(DynallocInputParms *inputParms, int *reasonCode) {
...
+    rc = -1;
+
+    below2G->textUnits[0] = createSimpleTextUnit2(DALDDNAM,
+                                                  inputParms->ddName,
+                                                  sizeof(inputParms->ddName));
+    if (below2G->textUnits[0] == NULL) {
+      break;
+    }
+    
+    below2G->textUnits[1] = createCharTextUnit(DALNDISP, 4);
```
chcaron01: Change 4 to DISP_DELETE

c/dynalloc.c
```diff
@@ -959,7 +959,60 @@ int unallocDataset(DynallocInputParms *inputParms, int *reasonCode) {
 
 }
 
+int freeDataset(DynallocInputParms *inputParms, int *reasonCode) {
```
ifakhrutdinov: This is a bad name. It unallocates a dataset and then because ```DISP_DELETE``` has been used, the dataset gets removed. It'd be better to add a flag value to the ```dynallocUnallocDatasetByDDName``` function.

PR #67, opened 2019-07-30
JSON and base64 improvements

fkovinAtRocket:
- the ability to parse a UTF8 buffer as JSON
- printing JSON to a buffer
- support for base64url
- small refactoring of the old base64 code


c/utils.c
```diff
@@ -1070,17 +1078,83 @@ char *encodeBase64(ShortLivedHeap *slh, char *buf, int size, int *resultSize, in
...
+    case '+':
+      s[i] = '-';
+      break;
+    case '/':
+      s[i] = '_';
+      break;
+    }
+    i++;
+  }
+  while (s[--i] == '=') {
```
fkovinAtRocket: self-found bug

PR #79, opened 2019-08-19
Implemented dynamic logging id generation for data services

jordanfilteau1995:


jordanfilteau1995:
https://github.com/zowe/zss/pull/78

ifakhrutdinov:
@jordanfilteau1995 @1000TurquoisePogs this has been a breaking change.

c/dataservice.c
```diff
@@ -298,13 +305,17 @@ WebPlugin *makeWebPlugin(char *pluginLocation, JsonObject *pluginDefintion, Inte
       char *type = jsonObjectGetString(serviceDef, "type");
 
       if (!type || !strcmp(type, "service")) {
-        plugin->dataServices[k++] = makeDataService(plugin, jsonArrayGetObject(dataServices, i), NULL, internalAPIMap);
+        plugin->dataServices[k++] = makeDataService(plugin, jsonArrayGetObject(dataServices, i), NULL, internalAPIMap,
+                                                    idMultiplier, pluginLogLevel);
+        (*idMultiplier++);
```
ifakhrutdinov: @jordanfilteau1995 , @1000TurquoisePogs  This has no effect - ```*``` and ```++``` have the same precedence but since they get processed right-to-left, pointer ```idMultiplier``` gets incremented first and then dereferenced.

jordanfilteau1995: Neat. Thanks!

h/dataservice.h
```diff
@@ -64,10 +64,12 @@ typedef struct DataService_tag {
   ExternalAPI *externalAPI;  /* a function call to some lower level module to get data */
   void        *extension;    /* another slot to stash things */
   JsonObject *serviceDefinition;
+  uint64 loggingIdentifier;
```
ifakhrutdinov: This breaks all plugins that haven't been rebuilt after this commit - they will assume that the ```plugin``` field is at the old offset, whereas it's moved.

PR #83, opened 2019-09-17
Fix the include path in the header

fkovinAtRocket:
Signed-off-by: Fyodor Kovin <fkovin@rocketsoftware.com>

### DIFF
--- a/jwt/jwt/jwt.h
+++ b/jwt/jwt/jwt.h
@@ -14,7 +14,7 @@
 #include "zowetypes.h"
 #include "json.h"
 #include "utils.h"
-#include "rs_icsfp11.h"
+#include "../rscrypto/rs_icsfp11.h"
 
 #define RC_JWT_OK                0
 #define RC_JWT_MEMORY_ERROR      1


jwt/jwt/jwt.h
```diff
@@ -14,7 +14,7 @@
 #include "zowetypes.h"
 #include "json.h"
 #include "utils.h"
-#include "rs_icsfp11.h"
```
ifakhrutdinov: What about ```jwt/rscrypto/rs_icsfp11.h```? With the solution you have you may include something by accident (of course with the name rs_icsfp11.h it's unlikely), i.e. the compiler will scan something you didn't intend it to. Plus, you clearly know this is from jwt.

jwt/jwt/jwt.h
```diff
@@ -14,7 +14,7 @@
...
+#include "../rscrypto/rs_icsfp11.h"
```
ifakhrutdinov: What about ```jwt/rscrypto/rs_icsfp11.h```? With the solution you have you may include something by accident (of course with the name rs_icsfp11.h it's unlikely), i.e. the compiler will scan something you didn't intend it to. Plus, you clearly know this is from jwt.

ifakhrutdinov: Or alternatively, ```rscrypto/rs_icsfp11.h``` if rscrypto is a library on its own.

fkovinAtRocket: This is for ZSS plugins that are unaware of this change and therefore can't (or don't want to) change their include path in the build script.

The plugin includes `httpserver.h`
`httpeserver.h` includes `jwt/jwt/jwt.h`
`jwt.h` includes `../rscrypto/rs_icsfp11.h`: this won't break the plugin 

`rscrypto/rs_icsfp11.h` won't work for that reason.

`jwt/rscrypto/rs_icsfp11.h` can also work,  but I don't think it's any better.  What can be at `jwt/jwt/../rscrypto/rs_icsfp11.h` "by accident" other than what's needed?

ifakhrutdinov: > Do you think it will try to include ../rscrypto/rs_icsfp11.h from every folder on the -I list?
Possibly. I guess it's not a big issue with such unique names.

Anyway, if that's the best option, lgtm.

trudeaun:
The IBM XL C/C++ on z/OS compiler will search all paths specified in all `-I` options until it finds the header file being referenced in the `#include` directive and will stop looking once the header file is first found. Since the header file in your case is enclosed in double quotes, the compiler will start looking in the directory containing the source with the `#include` directive.

PR #86, opened 2019-09-27
PC, paulse elements, A/S and shared memory objects wrappers for the ZIS AUX projects.

ifakhrutdinov:
The following wrappers have been added:
* Address space management
* Pause elements
* Program calls
* 64-bit shared memory objects

### DIFF
--- a/c/shrmem64.c
+++ b/c/shrmem64.c
@@ -12,14 +12,16 @@
 
 #ifdef METTLE
 #include <metal/metal.h>
+#include <metal/stdbool.h>
 #include <metal/stddef.h>
 #include <metal/stdint.h>
 #else
+#include "stdbool.h"
 #include "stddef.h"
 #include "stdint.h"
 #endif
 
-#include <shrmem64.h>
+#include "shrmem64.h"
 #include "zos.h"
 
 #ifndef _LP64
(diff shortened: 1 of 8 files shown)


c/shrmem64.c
```diff
@@ -0,0 +1,364 @@
...
+#error ILP32 is not supported
+#endif
+
+typedef uint64_t MemObj;
+
+#define IARV64_V4PLIST_SIZE 160
+
+#define IS_IARV64_OK($rc) (($rc) < 8)
+
+#define MAKE_RSN($functionRC, $iarv64RC, $iarv64RSN) \
```
daveyc: Prefer functions to macros

c/shrmem64.c
```diff
@@ -0,0 +1,364 @@
...
+
+#ifndef _LP64
+#error ILP32 is not supported
+#endif
+
+typedef uint64_t MemObj;
+
+#define IARV64_V4PLIST_SIZE 160
+
+#define IS_IARV64_OK($rc) (($rc) < 8)
```
daveyc: Prefer functions to macros

c/shrmem64.c
```diff
@@ -0,0 +1,364 @@
...
+#include <shrmem64.h>
+#include "zos.h"
+
+#ifndef _LP64
+#error ILP32 is not supported
+#endif
+
+typedef uint64_t MemObj;
+
+#define IARV64_V4PLIST_SIZE 160
```
daveyc: prefer "static const" to #define

c/shrmem64.c
```diff
@@ -0,0 +1,364 @@
...
+  return result.token;
+}
+
+int shrmem64Alloc(MemObjToken userToken, size_t size, void **result, int *rsn) {
+
+  uint32_t iarv64RC = 0, iarv64RSN = 0;
+
+  uint64_t segmentCount = 0;
+  if ((size & 0xFFFFF) == 0){
+    segmentCount = size >> 20;
```
daveyc: Please simplify or explain the mask with a comment or constant.

c/shrmem64.c
```diff
@@ -0,0 +1,364 @@
...
+  } else{
+    segmentCount = (size >> 20) + 1;
```
daveyc: Ditto

c/shrmem64.c
ifakhrutdinov: I'm not sure I want to use a static const. That value is used to define the size of an array. A static const would make the compiler use VLA which are optional in C11.

PR #89, opened 2019-10-09
Fixed bool function --> int function and added better response returns

DivergentEuropeans:
- Reorganized logic to use boolean better and changed a 500 to a 400 to better differentiate a probable permission problem from a general server error.

PR 2 of 2
PR 1 https://github.com/zowe/zss/pull/93

### DIFF
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -68,6 +68,9 @@ bool isDir(char *absolutePath) {
   FileInfo info = {0};
 
   status = fileInfo(absolutePath, &info, &returnCode, &reasonCode);
+  if (status == -1) {
+    return FALSE;
+  }
 
   return (fileInfoIsDirectory(&info));
 }
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -68,15 +68,12 @@ bool isDir(char *absolutePath) {
   FileInfo info = {0};
 
   status = fileInfo(absolutePath, &info, &returnCode, &reasonCode);
-  if (status == -1) {
-    return -1;
-  }
 
   return (fileInfoIsDirectory(&info));
 }
 
 /* Returns a boolean value for whether or not
- * the specified file exists.
+ * the specified file/directory exists.
  */
 bool doesFileExist(char *absolutePath) {
   int returnCode = 0, reasonCode = 0, status = 0;  
@@ -160,7 +157,7 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
     response200WithMessage(response, "Successfully deleted a directory");
   }
   else {
-    respondWithJsonError(response, "Failed to delete a directory", 500, "Internal Server Error");
+    respondWithJsonError(response, "Failed to delete a directory", 400, "Bad Request");
   }
 }
 
@@ -195,7 +192,7 @@ void deleteUnixFileAndRespond(HttpResponse *response, char *absolutePath) {
     response200WithMessage(response, "Successfully deleted a file");
   }
   else {
-    respondWithJsonError(response, "Failed to delete a file", 500, "Internal Server Error");
+    respondWithJsonError(response, "Failed to delete a file", 400, "Bad Request");
   }
 }
 
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -69,7 +69,7 @@ bool isDir(char *absolutePath) {
 
   status = fileInfo(absolutePath, &info, &returnCode, &reasonCode);
   if (status == -1) {
-    return FALSE;
+    return false;
   }
 
   return (fileInfoIsDirectory(&info));
@@ -84,10 +84,10 @@ bool doesFileExist(char *absolutePath) {
 
   status = fileInfo(absolutePath, &info, &returnCode, &reasonCode);
   if (status == -1) {
-    return FALSE;
+    return false;
   }
   
-  return TRUE;
+  return true;
 }
 
 /* Creates a new unix directory at the specified absolute


c/httpfileservice.c
```diff
@@ -60,10 +60,13 @@ void response200WithMessage(HttpResponse *response, char *msg) {
...
-/* Returns a boolean value for whether or not
- * the specified file is a directory or not.
+/* Returns an int value for whether or not the specified
+ * path is a file, folder, or it doesn't exist.
+ * -1 Does not exist
+ * 0 File exists 
+ * 1 Folder exists
  */
-bool isDir(char *absolutePath) {
+int doesFileOrDirectoryExist(char *absolutePath) {
```
ifakhrutdinov: * Why did you change an existing global function?
* Why does a file related global function have to be in ```httpfileservice.c```? It has nothing to do with http.

c/httpfileservice.c
```diff
@@ -160,7 +148,8 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
     response200WithMessage(response, "Successfully deleted a directory");
   }
   else {
-    respondWithJsonError(response, "Failed to delete a directory", 500, "Internal Server Error");
+    respondWithJsonError(response, "Failed to delete a directory. This is usually "
+	"because of permission problems", 403, "Forbidden");
```
ifakhrutdinov: 403 may be misleading here. If someone only checks for http codes, they'll think there is really a permission problem whereas the file can be just missing.

ifakhrutdinov:
@DivergentEuropeans Your ```Signed-off-by``` doesn't seem right.

c/httpfileservice.c
DivergentEuropeans: - I changed isDir because it didn't work. It was a bool that returned 3 things. I also changed the name because the name didn't make much sense compared to what it was trying to do. So either keep 2 methods that are bools, or make 1 int method return 3 conditions, I chose the second option. It was also being used in one place, and thus will not break things because I also updated where it was being used.

- I don't have answer for that, that's just where it was before, didn't want to change stuff up. I guess I could move it to unixFileService.c instead if that's useful?

c/httpfileservice.c
DivergentEuropeans: If the file is missing, we would get a 404 prior to reaching this point. 
(see https://github.com/zowe/zss/blob/5638836e4e773a45c977a40480f0526ab3d09b90/c/unixFileService.c#L714)
You can view Jordan and I's email chain for info, but here are the things we know when we get to this point in the code.

1. The server is running
2. The user is authenticated
3. The path is either a file or folder and
4. The file or folder exists, and can be proven to exist from the perspective of the server

The question is, are these conditions suffice to determine that if we can't delete the file or folder at this point, we are probably blocked by permissions or something else related.

c/httpfileservice.c
ifakhrutdinov: * ```bool isDir(const char*)``` should tell whether the specified file is a directory. If it doesn't, then there is a bug. From this name, I only expect that. If the path doesn't exist, it's not a directory. The name should drive the implementation in this case, that's an existing function.
* You can make a new static function to do what you need to do. If anyone needs that outside this file, they'll make it global.

c/httpfileservice.c
ifakhrutdinov: > The file or folder exists, and can be proven to exist from the perspective of the server

No, ```deleteUnixDirectory``` may fail due to other reasons, for example, if the file/directory doesn't exist. A more correct way would be mapping the return codes from the system services we use to the corresponding response message texts (permission denied, file missing, etc).

c/httpfileservice.c
DivergentEuropeans: Right @ifakhrutdinov and it didn't work, which is why it was a bug. So I made it work and renamed it to a name that is more appropriate for what it does. If you want instead, I could keep "isDir" and then remove the -1 part, thereby removing the third condition and make it a true "bool" if that's a good thing to do, but I don't see the point of it when 2 methods can just be 1 for simplicity's sake.

I can keep the original method and make another, whatever works best I'll incorporate

c/httpfileservice.c
DivergentEuropeans: @ifakhrutdinov That's not true, because `deleteUnixDirectory` gets called in `unixfileservice.c`, and thus if the file/directory doesn't exist, it would throw a 404 _before_ reaching this function in `httpfileservice.c` at all

So those 4 conditions _are_ true in the context of how they are used, however if the functions get called by themselves then they're not true. So I think what I'll do is to make the functions work better if they're called somewhere else to future-proof

1. keep isDir
2. fix isDir to work as a boolean
3. use isDir to check if deletion should happen for a file or dir
4. fix delete methods to do checks for if the file exists

DivergentEuropeans: Thanks @fkovinAtRocket I'll use 400 instead. I don't want to pile through response codes and texts right now, I have Editor work to finish.

DivergentEuropeans:
Git closed this after I updated this branch for some reason.

c/httpfileservice.c
```diff
@@ -68,15 +68,12 @@ bool isDir(char *absolutePath) {
   FileInfo info = {0};
 
   status = fileInfo(absolutePath, &info, &returnCode, &reasonCode);
```
ifakhrutdinov: I'd keep the check and return ```false```.
```c
if (status == -1) {
  return false;
}
```
The reason is because usually you can't rely on the result of a function if the return value doesn't indicate success. So ```fileInfoIsDirectory``` shouldn't be called in that case.

PR #90, opened 2019-10-10
500 for delete is 400 to differentiate server failure

DivergentEuropeans:
- Reorganized logic to use boolean better and changed a 500 to a 400 to better differentiate a probable permission problem from a general server error.

For some reason, Git closed my previous pull request when I pushed to the same branch. Original comments, suggestions, and then fixes were previously addressed here: https://github.com/zowe/zowe-common-c/pull/89


DivergentEuropeans:
Duplicate. Git nonesense.

PR #95, opened 2019-12-02
Fix for deleting nonexisting member

sakeerthy:
Signed-off-by: Suneeth Keerthy <skeerthy@rocketsoftware.com>

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -75,8 +75,7 @@ typedef struct Volser_tag {
 } Volser;
 
 static int getVolserForDataset(const DatasetName *dataset, Volser *volser);
-bool memberExists(char* dsName, DynallocMemberName daMemberName);
-void nullTermDSName(char* dsName, DatasetName datasetName);
+static bool memberExists(char* dsName, DynallocMemberName daMemberName);
 
 int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter *jPrinter){
 #ifdef __ZOWE_OS_ZOS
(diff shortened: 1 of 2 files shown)


c/datasetjson.c
```diff
@@ -1185,6 +1185,52 @@ void deleteDatasetOrMember(HttpResponse* response, char* absolutePath) {
...
   else {
+    char *typesArg = defaultDatasetTypesAllowed;
+    int datasetTypeCount = (typesArg == NULL) ? 3 : strlen(typesArg);
+    int workAreaSizeArg = 0;
+    int fieldCount = defaultCSIFieldCount;
+    char **csiFields = defaultCSIFields;
+
+    csi_parmblock *returnParms = (csi_parmblock*)safeMalloc(sizeof(csi_parmblock),"CSI ParmBlock");
+
+    char dsName[45];
```
jordanfilteau1995: There is a #define for this value.

c/datasetjson.c
```diff
@@ -1185,6 +1185,52 @@ void deleteDatasetOrMember(HttpResponse* response, char* absolutePath) {
...
+    for (int i = 0; i < sizeof(datasetName.value); i++) {
```
jordanfilteau1995: Make this a separate function. I really have no idea what the purpose of this is by looking at it, but it looks like it happens more than once.

c/datasetjson.c
```diff
@@ -1185,6 +1185,52 @@ void deleteDatasetOrMember(HttpResponse* response, char* absolutePath) {
...
+        dsName[i] = datasetName.value[i];
+      }
+    }
+    EntryDataSet *entrySet = returnEntries(dsName, typesArg, datasetTypeCount, 
+                                           workAreaSizeArg, csiFields, fieldCount, 
+                                           NULL, NULL, returnParms);
+                 
+    EntryData *entry = entrySet->entries[0];
+  
+    bool foundMember = false;
```
jordanfilteau1995: This logic should be in a separate function too.

PR #96, opened 2019-12-03
SRB support in the recovery facility

ifakhrutdinov:
Changes:
* Use FRR when running under an SRB
* Bypass the SDWA check under an SRB
* Use the FRR exit key when accessing SDWA field (to prevent 0C4 when
SDWA is fetch protected)
* Tell SETRP to restore R15 as this behavior is different from ESTAE
* Verify that the state structure after is valid we've retried
* Save and restore the caller's key in recoveryPush - SETRP restores the
key used at the time of SETFRR which may be different from what we want
* Use better names for the labels in the recovery push asm

### DIFF
--- a/c/recovery.c
+++ b/c/recovery.c
@@ -237,6 +237,93 @@ static ESTAEXFeedback deleteESTAEX() {
   return localFeedback;
 }
 
+static void setFRR(void * __ptr32 userExit, void *userData,
+                   bool nonInterruptible) {
+
+  int wasProblem = supervisorMode(TRUE);
+  int oldKey = setKey(0);
+
+  union {
+    void *userData;
+    char data[24];
+  } * __ptr32 parmArea = NULL;
+
+  if (nonInterruptible) {
+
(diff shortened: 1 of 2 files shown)


c/recovery.c
```diff
@@ -865,27 +971,39 @@ int recoveryEstablishRouter(int flags) {
...
+             feedback.returnCode, feedback.reasonCode);
 #endif
-    if (previousESPIEToken != 0) {
-      resetESPIE(previousESPIEToken);
-      previousESPIEToken = 0;
+      if (previousESPIEToken != 0) {
+        resetESPIE(previousESPIEToken);
+        previousESPIEToken = 0;
+      }
+      storageRelease((char *)context, sizeof(RecoveryContext));
```
daveyc: What is the purpose of the cast?

ifakhrutdinov: Can't tell. This could have been copied/movied as is (with the cast) from another function in the original code.

PR #97, opened 2019-12-13
Change mode

sbollingerAtRocket:
Adding code to change file mode.


1000TurquoisePogs:
I think this one is older and outdated.

PR #104, opened 2019-12-31
chown support

sbollingerAtRocket:
Change owner and group ID for files and/or directories.


c/httpfileservice.c
```diff
@@ -452,6 +460,76 @@ void respondWithUnixFileMetadata(HttpResponse *response, char *absolutePath) {
   }
 }
 
+void directoryChangeOwnerAndRespond(HttpResponse *response, char *path,
+        char *usrId, char *grpId, char *Recursive, char *pattern) {
```
ifakhrutdinov: What is `usrId` used for? Are we mixing UID and TSO user name?

c/zosfile.c
```diff
@@ -770,10 +783,166 @@ int symbolicFileInfo(const char *filename, BPXYSTAT *stats, int *returnCode, int
...
+    }
+  }
+
+  if (returnValue != 0) {
+    returnValue = -1;
+  }
+  return returnValue;
+}
+
+int gidGetUserInfo(const char *userName,  UserInfo * info,
```
ifakhrutdinov: It doesn't seem to belong in `zosfile`. It's used to get user info. Same for `gidGetGroupInfo`.

c/httpfileservice.c
```diff
@@ -452,6 +460,76 @@ void respondWithUnixFileMetadata(HttpResponse *response, char *absolutePath) {
...
+  else if (stringIsDigit(usrId)) {
+    userId = atoi(usrId);
+  }
+  else {
+    status = gidGetUserInfo(usrId, &userInfo, &returnCode, &reasonCode);
+    if (status == 0) {
+      userId = userInfoGetUserId ( &userInfo);
+    }
+    else {
+      sprintf(message, "Bad Input: User %s NOT found", usrId);
```
ifakhrutdinov: If the `usrId` variable is large enough, you'll get a buffer overflow.

c/httpfileservice.c
ifakhrutdinov: Why do we pass strings for group ID, user ID and Recursive? Since this is library code, it would make more sense to pass the expected types, i.e. int and bool. That way this function wouldn't have to do additional data validation and would just change the owner.

ifakhrutdinov: Can you please show an example of a pattern and how it's used with this function?

c/utils.c
```diff
@@ -2169,6 +2169,29 @@ static int incrementPlaceValues(int *placeValues,
...
+  for (int i = 0; i < strlen(str); i ++) {
+    if (!isdigit(str[i])) {
+      returnValue = FALSE;
+      break;
+    }
+  }
+  return returnValue;
+}
+
+#ifndef METTLE
```
ifakhrutdinov: Why LE only?

c/httpfileservice.c
```diff
@@ -21,7 +21,7 @@
 #else
 #include <stdio.h>
 #include <string.h>
-
+#include <le/limits.h>
```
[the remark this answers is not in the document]
sbollingerAtRocket: I can use the standard limits.h for le.  However, this file returns an error if it is built in the metal environment.  I can add meta/limits.h for this.  The following definitions will PATH_MAX defined regardless of whether either is included.

c/httpfileservice.c
```diff
@@ -452,6 +460,76 @@ void respondWithUnixFileMetadata(HttpResponse *response, char *absolutePath) {
   }
 }
 
+void directoryChangeOwnerAndRespond(HttpResponse *response, char *path,
+        char *usrId, char *grpId, char *Recursive, char *pattern) {
+# define RETURN_MESSAGE_SIZE (PATH_MAX + 50)
+
+  char message[RETURN_MESSAGE_SIZE] = {0};
+  strncpy(message,"", RETURN_MESSAGE_SIZE);
```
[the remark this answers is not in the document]
sbollingerAtRocket: Was there to setup a default message, but not being used as such.  Will eliminate.

c/httpfileservice.c
sbollingerAtRocket: The usId and the grpId follows the man pages for chown, where each could be the name of the user/group or their numeric value.

c/httpfileservice.c
sbollingerAtRocket: Changed to snprintf.

c/httpfileservice.c
sbollingerAtRocket: This is following the man page definitions where either name or numerical values may be passed.  As such, the comparison must be done somewhere.  I chose to do it in the common library where it is closer to low level functions which returns the user and group information.

sbollingerAtRocket: You can specify something like ".txt" which will select any file with this string in it, not necessarily the ending file name.  Or something like "foo" which will include any directory/file which contains the sub-string "foo".  It will not work with wildcards/

c/utils.c
sbollingerAtRocket: I can eliminate, but code is not tested for METTLE.

c/httpfileservice.c
ifakhrutdinov: Is this something that's been discussed and people agreed on such behavior?

ifakhrutdinov: Can you please provide the link?

Here's what I found from IBM, https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.bpxbd00/rtcho.htm

c/httpfileservice.c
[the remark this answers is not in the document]
ifakhrutdinov: I've successfully used `limits.h` for Metal C, see https://github.com/zowe/zowe-common-c/blob/master/c/crossmemory.c

c/httpfileservice.c
sbollingerAtRocket: This has not been wildly discussed.  It has been seen in other code reviews.  In each instance, the pattern check has been localized to 1 function.  I figured that we may need to modify this check sometime in the future.

sbollingerAtRocket: https://linuxize.com/post/linux-chown-command/

And in the linux implementation, the uid parameter may be the user id or the name of the user.
So the question is, what standard are we following.
I ask Sean this and he requested that we allow the user name to be passed (also the group).  Note the current implementation allows you to pass in either.

c/utils.c
sbollingerAtRocket: My code change includes <limits.h> for LE and <metal/limits.h> for METAL.  I cannot include just one file for both LE and METAL.

c/httpfileservice.c
daveyc: On z/OS the definitive answer is sometimes in the low-level UNIX assembler services manual [chown](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.bpxb100/cho.htm)

daveyc:
```-qnosearch -I/usr/include/metal -I/usr/include``` takes out LE search order for Metal/C compiles

File: c/httpfileservice.c (hunk 1 above)
Comment (ifakhrutdinov) [reply c/httpfileservice.c]: I think more logical would be:
* The ZSS service accepts both user name and UID (i.e. behaves like `chown` program)
* The low level API in zowe-common-c accepts the UID and GID types (i.e. behaves as the `chown` library function, like [UNIX chown](http://man7.org/linux/man-pages/man2/chown.2.html) or  [IBM chown](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.bpxbd00/rtcho.htm) do)

File: c/utils.c (hunk 4 above)
Comment (ifakhrutdinov) [reply c/utils.c]: I understand. You don't have to include just one. Just make sure that Metal C still compiles at least.

File: c/httpfileservice.c (hunk 1 above)
Comment (ifakhrutdinov) [reply c/httpfileservice.c]: @sbollingerAtRocket what do you think about our proposal?

Comment (sbollingerAtRocket) [reply c/httpfileservice.c]: I already have a functions that is almost this: 
fileChangeOwner(char *fileName, int *returnCode, int *reasonCode, int usrId, int grpID)
I can easily create function chown function that ignores the return and reason code.

Currently, the zss server accepts a user and group variable as either an integer value or the name of a user/group.  The real question is where the conversion to integer values occurs.
Currently, it is being done in the zowe-common and not in zss.    Even if the conversion was done in the zss, it still would not be calling the low level function directly.  This interface supports recursion and pattern checking and that would still be done in the zowe-common.

Comment (1000TurquoisePogs) [reply c/httpfileservice.c]: At this level of the logic, it is the best user experience to support both IDs and names, as zos as well as every other OS I have ever seen that has chown does have a way to specify names. Certainly, there may be a low-level call that only uses IDs, but it's a convenience here that we support both and convert for compatibility when calling down below.

File: c/zosfile.c (hunk 2 above)
Comment (1000TurquoisePogs) [reply c/zosfile.c]: zosaccounts.c now

**Comment by 1000TurquoisePogs** [comment]
I'm just going to run a compile test here, but Jordan & Steve & I met and we're all happy with the organization of the code at this point, all previous conversations seem resolved.

**Comment by ifakhrutdinov** [comment]
@1000TurquoisePogs @sbollingerAtRocket how did you compile test go? I can see new warning messages now.

```
WARNING CCN3236 ./../../deps/zowe-common-c/c/httpfileservice.c:58    Macro name PATH_MAX has been redefined.
FSUM3065 The COMPILE step ended with return code 4.
```

What is the correct value of `PATH_MAX`?

sbollingerAtRocket:
I am still trying to merge my sources and rebuild.  I am having problems with this.

PR #106, opened 2020-01-06
Unixfile/contents:  Allow ISO8859-1 encoding

sbollingerAtRocket:
Allow ISO8859-1 along with ISO-8859-1.
Make all file encoding case insensitive.

### DIFF
--- a/c/charsets.c
+++ b/c/charsets.c
@@ -212,8 +212,7 @@ int getCharsetCode(const char *charsetName) {
   strupcase (localArray);
 
   if ((!strcmp(localArray, "ISO-8859-1"))  ||
-      (!strcmp(localArray, "ISO8859-1")))
-     {
+      (!strcmp(localArray, "ISO8859-1"))) {
     return CCSID_ISO_8859_1;
   }
   else if (!strcmp(localArray, "IBM-1047")) {
--- a/c/charsets.c
+++ b/c/charsets.c
@@ -200,22 +200,24 @@ int convertCharset(char *input,
 }
 
 int getCharsetCode(char *charsetName) {
-  if (!strcmp(charsetName, "ISO-8859-1")) {
+  if ((!strcmp(strupcase(charsetName), "ISO-8859-1"))  ||
+      (!strcmp(strupcase(charsetName), "ISO8859-1")))
+     {
     return CCSID_ISO_8859_1;
   }
-  else if (!strcmp(charsetName, "IBM-1047")) {
+  else if (!strcmp(strupcase(charsetName), "IBM-1047")) {
     return CCSID_IBM1047;
   }
-  else if (!strcmp(charsetName, "UTF-8")) {
+  else if (!strcmp(strupcase(charsetName), "UTF-8")) {
     return CCSID_UTF_8;
   }
-  else if (!strcmp(charsetName, "UTF-16")) {
+  else if (!strcmp(strupcase(charsetName), "UTF-16")) {
     return CCSID_UTF_16;
   }
-  else if (!strcmp(charsetName, "UTF-16BE")) {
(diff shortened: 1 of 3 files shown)


c/charsets.c
```diff
@@ -200,22 +200,24 @@ int convertCharset(char *input,
 }
 
 int getCharsetCode(char *charsetName) {
-  if (!strcmp(charsetName, "ISO-8859-1")) {
+  if ((!strcmp(strupcase(charsetName), "ISO-8859-1"))  ||
```
ifakhrutdinov: What if someone passes a string literal? That would cause an S0C4 ABEND.

For example,
```c
getCharsetCode("IBM_1047");
```

I suggest making the argument `const` and using a local buffer.

As for the uppercasing, is passing mixed case names valid? Wouldn't it be better to uppercase before passing the name?

sbollingerAtRocket: Code was modified to use local array in comparison.

The other functions that we have modified/created allows for these encoding to be case insensitive.  The strupcase is keeping with this policy.

c/charsets.c
```diff
@@ -199,23 +199,37 @@ int convertCharset(char *input,
...
+int getCharsetCode(const char *charsetName) {
+  char localArray[CHARSETNAME_SIZE] = {0};
+
+  /* Make sure last element is 0 */
+  if (strlen(charsetName) > CHARSETNAME_SIZE) {
+    return -1;
+  }
+# undef CHARSETNAME_SIZE
+
+  strncpy( localArray, charsetName, sizeof (localArray) -1);
```
ifakhrutdinov: There are arbitrary spaces here, can you please use a consistent style?

c/charsets.c
```diff
@@ -199,23 +199,37 @@ int convertCharset(char *input,
...
+  strupcase (localArray);
+
+  if ((!strcmp(localArray, "ISO-8859-1"))  ||
+      (!strcmp(localArray, "ISO8859-1")))
+     {
```
ifakhrutdinov: Bad indentation.

c/charsets.c
```diff
@@ -199,23 +199,37 @@ int convertCharset(char *input,
...
 }
 
-int getCharsetCode(char *charsetName) {
-  if (!strcmp(charsetName, "ISO-8859-1")) {
+#define CHARSETNAME_SIZE 15
+int getCharsetCode(const char *charsetName) {
+  char localArray[CHARSETNAME_SIZE] = {0};
+
+  /* Make sure last element is 0 */
+  if (strlen(charsetName) > CHARSETNAME_SIZE) {
```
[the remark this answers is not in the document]
sbollingerAtRocket: Change to second suggestion.  Make local array 1 size longer.  Change strncpy to strcpy.

c/charsets.c
```diff
@@ -199,23 +199,35 @@ int convertCharset(char *input,
   }
 }
 
-int getCharsetCode(char *charsetName) {
-  if (!strcmp(charsetName, "ISO-8859-1")) {
+#define CHARSETNAME_SIZE 15
+int getCharsetCode(const char *charsetName) {
```
ifakhrutdinov: @sbollingerAtRocket please correct the declaration too.

ifakhrutdinov: @1000TurquoisePogs @sbollingerAtRocket this hasn't been addressed.

sbollingerAtRocket: What part are you referring.  I created a local array with an additional character at the end and have it initialized to guarantee that the last element is null.  The code then test to verify the the input name will not overwrite it before coping it into the local array which then the code does a convert to upper case.

ifakhrutdinov: > @sbollingerAtRocket please correct the declaration too.

PR #109, opened 2020-01-16
Create Qsam data set

sbollingerAtRocket:


c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+  int status, value;
+  char *ptr;
+
+  /* Must supply daName */
+  if (request->daName == NULL) {
+    return -1;
+  }
+
+  /* Allocate structure in A31 space */
+  EXTF *bpxwdyn=(EXTF *)fetch("BPXWDYN");
```
daveyc: There is no matching ```release()``` for this fetch.

sbollingerAtRocket: Nothing that I have read has mentioned anything about releasing an item that has been fetched.  What is your source for this?

daveyc: The source is in the [manual](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.bpxbd00/xrelease.htm#xrelease). It won't leak resources because the CDE has a use count but it's always good to clean up. You could stash the ```bpxwdyn``` pointer in a static variable and fetch it once. It will get released when the task ends.

ifakhrutdinov: `fetch` may return NULL. It needs to be checked.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+  ALLOC_STRUCT31(
+    STRUCT31_NAME(below2G),
+    STRUCT31_FIELDS(
+      char  request[REQUEST_MAX_LENGTH];
+    )
+  );
+
+  /* setup initial request string */
+  strncpy( below2G->request, baseRequest, sizeof(below2G->request));
```
ifakhrutdinov: @sbollingerAtRocket this is not the first PR I mentioned this - please do not insert arbitrary blanks or omit blanks where they're needed.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+
+  /* Initialize the DD/FI field */
+  ptr = (request->ddName == NULL ?  QSAM_DEFAULT_DDNAME: request->ddName);
```
ifakhrutdinov: We already have code for dynalloc in https://github.com/zowe/zowe-common-c/blob/master/c/dynalloc.c, you should re-use/enhance code from there.

c/datasetjson.c
sbollingerAtRocket: Thanks for the info.  None of the examples on bpxwdyn has the release function shown.

daveyc: In fact, I'm not sure that it won't leak memory because there is a fetch control block associated with fetch used by LE. The underlying LOAD macro used by fetch is fine but I would still always call DELETE when I use a LOAD.

daveyc: It's a fetch issue not related to BPXWDYN. fetch is is just a callable service.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+  if (request->spaceUnit != NULL) {
+    snprintf(field, sizeof(field)-1, "SPACE(%s) ", request->spaceUnit);
+    strncat( below2G->request, field,
+             sizeof (below2G->request) - strlen(below2G->request));
+  }
+
+  /****************************/
+  /* Create the qsam data set */
+  /****************************/
+  printf (" REQUEST: %s\n",  below2G->request);
```
ifakhrutdinov: Naked printf.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+    /* status between -20 and -9999 represents Key errors. */
+    /* The key index is (-status -20)                      */
+    /* status below -10000 has the IEFDB476 return code in */
+    /* lower 2 digits (decimal) of the status.             */
+    /* Status code between  -1610612737 and -2147483648    */
+    /* contains the S99ERROR code.                         */
+    reason = strtok(below2G->request, " ");
+    sprintf(field,"", "");  
+    if ((status <= -20) && (status >= -9999)) {
+      cause = (-status - 20) %100;
```
ifakhrutdinov: What are all these magic numbers? The comments are not enough, please use `#define`s.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
...
+  snprintf(field, sizeof(field)-1, "da(%s) ", request->daName);
+  strncat( below2G->request, field,
+             sizeof (below2G->request) - strlen(below2G->request));
+  status = bpxwdyn(below2G->request);
+
+  /* free the structure */
+freeStruct:
+  FREE_STRUCT31(
+    STRUCT31_NAME(below2G)
+    );
```
ifakhrutdinov: Bad indentation. Please review all the code.

h/qsam.h
```diff
@@ -359,7 +359,45 @@ int bpamFind(char * __ptr32 dcb, char * __ptr32 memberName, int *reasonCode);
...
+  int     secondary;
+  int     fileSize;      /* Size of file*/
+  int     numGenerations;
+  int     avrBlockLength;
+  char   *dataClass;
+  char   *averageRecord;   /* M, K or U */
+  char   *expiration;
+  char   *spaceUnit;      /* BLKS, TRKS, CYCLS, KB, MB, BYTES */
+  int     dirBlk;
+  } qsamFileRequest;
```
ifakhrutdinov: This doesn't belong in qsam.h.

h/datasetjson.h
```diff
@@ -62,6 +63,7 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
 void respondWithVSAMDataset(HttpResponse* response, char* absolutePath, hashtable *acbTable, int jsonMode);
 void respondWithDatasetMetadata(HttpResponse *response);
 void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQueryCache);
+int qsamAllocFile( qsamFileRequest *request, char* message, int messageLength);
```
ifakhrutdinov: There are no QSAM files or datasets, QSAM is an access method.

ifakhrutdinov (changes requested):

c/datasetjson.c
sbollingerAtRocket: But I would assume that any example showing how to use bpxwdyn would also show the release of the function since it already show the fetch of it.

c/datasetjson.c
```diff
@@ -2251,9 +2254,238 @@ void respondWithHLQNames(HttpResponse *response, MetadataQueryCache *metadataQue
 #endif /* __ZOWE_OS_ZOS */
 }
 
+/*************************/
+#define FIELD_MAX_LENGTH    1024
+#define REQUEST_MAX_LENGTH  2048
 
-#endif /* not METTLE - the whole module */
+int qsamAllocFile( qsamFileRequest *request, char* message,
```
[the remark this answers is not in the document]
sbollingerAtRocket: Yes.  That was noted yesterday.  The jira ticket called the file "qsam" and we have kept up with that naming convention.  This interface may create many types of data sets.  If you have a better name then please suggest it.  I will gladly change it.  Also, Jordan has asked that these functions be moved to the alloc.c or dynallo.c files.

c/datasetjson.c
sbollingerAtRocket: Thanks, fixed

sbollingerAtRocket: Thanks, fixed

c/datasetjson.c
sbollingerAtRocket: This is a coding convention, can you show me the spec.  I put in these blank lines for readability.

c/datasetjson.c
sbollingerAtRocket: This was left in for debugging.  It has been removed.

c/datasetjson.c
ifakhrutdinov: I'm talking about spaces between C constructs.

For example there is a blank between `strncpy(` and `below2G`, but there is none between the closing brackets of the `sizeof` operator and `strncpy`:
```c
strncpy( below2G->request, baseRequest, sizeof(below2G->request));
```

c/datasetjson.c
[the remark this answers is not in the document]
ifakhrutdinov: This is public code that has nothing to do with internal Jira ticket, we shouldn't mentioned them here. I don't have a name for this function right...

c/datasetjson.c
sbollingerAtRocket: Using #defines will not make the code any more understandable then the comments.  
Have modified.

h/datasetjson.h
sbollingerAtRocket: From earlier comment.  I am open to any name change from qsam.  Following names in the jira ticket.

h/qsam.h
sbollingerAtRocket: Yes I know.  When we decide where these functions should be move to, then this will also be moved.

c/datasetjson.c
daveyc: I think the comments are great but I agree with @ifakhrutdinov that constants would make the code more readable. I would use both.

h/datasetjson.h
daveyc: @ifakhrutdinov is correct. I would name it ```allocDataSet``` which covers both PS and PDS(E) data sets.

1000TurquoisePogs:
Steve noted that strtok used here is not thread safe, so tokenizing alternatives may be needed here.

h/dynalloc.h
```diff
@@ -782,6 +782,37 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+  int     fileSize;      /* Size of file*/
+  int     numGenerations;
+  int     avrBlockLength;
+  char   *dataClass;
+  char   *averageRecord;   /* M, K or U */
+  char   *expiration;
+  char   *spaceUnit;      /* BLKS, TRKS, CYCLS, KB, MB, BYTES */
+  int     dirBlk;
+  } dataSetRequest;
+int allocDataSet( dataSetRequest *request, char* message, int messageLength);
```
ifakhrutdinov: * Please remove the blank char between `allocDataSet(` and `dataSetRequest *request,`.
* Is the content of message changed inside allocDataSet? If not, please use const.
* Please use the dynalloc prefix for new functions.
* I believe `size_t` would be more appropriate to represent a size.

c/dynalloc.c
```diff
@@ -1164,6 +1165,274 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+  int  status, value;
+  char *ptr;
+
+  /* Must supply daName */
+  if (request->daName == NULL) {
+    return -1;
+  }
+
+  /* Allocate structure in A31 space */
+  EXTF *bpxwdyn=(EXTF *)fetch("BPXWDYN");
```
ifakhrutdinov: This is not going to work in Metal C. Why not use the `invokeSVC99` function?

c/dynalloc.c
```diff
@@ -29,6 +29,7 @@
 #include "zowetypes.h"
 #include "alloc.h"
 #include "dynalloc.h"
+#include "logging.h"
```
ifakhrutdinov: We're introducing a new dependency. Can we work out a different way of returning debug info? Perhaps return and reason codes.

c/dynalloc.c
```diff
@@ -1164,6 +1165,274 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+  char field[FIELD_MAX_LENGTH];
+  int  status = 0, value;
+  char *baseRequest = "alloc new catalog msg(2) ";
+  char tempString[20] = {0};
+
+  /* setup initial request string */
+  strncpy(buffer, baseRequest, bufferSize);
+
+  /* Initialize the DD/FI field */
+  ptr = (request->ddName == NULL ?  DATASET_DEFAULT_DDNAME: request->ddName);
```
ifakhrutdinov: Please take a look at the existing functions in this file, they use a programmatic way to pass this information.

c/dynalloc.c
```diff
@@ -1164,6 +1165,274 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+
+  /* Initialize expiration data  */
+  if (request->spaceUnit != NULL) {
+    snprintf(field, sizeof(field)-1, "SPACE(%s) ", request->spaceUnit);
+    strncat(buffer, field, bufferSize - strlen(buffer));
+  }
+  return status;
+}
+
+static void  allocDataSetCloseString(dataSetRequest *request, char *buffer,
```
ifakhrutdinov: What about reusing `dynallocUnallocDatasetByDDName` and the other unalloc functions?

ifakhrutdinov (changes requested):
I suggest we use SVC99 instead of passing commands and then parsing the results.

c/dynalloc.c
```diff
@@ -1164,6 +1165,274 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+
+  /* Initialize data set type */
+  if ((ptr = request->dsnType) != NULL) {
+    snprintf(field, sizeof(field)-1, "DSNTYPE(%s) ", ptr);
+    strncat(buffer, field, bufferSize - strlen(buffer));
+  }
+
+  /* Initialize organization */
+  if (request->organization != NULL) {
+    snprintf(field, sizeof(field)-1, "DSORG(%s) ", request->organization);
```
ifakhrutdinov: * `snprintf` may fail
* `snprintf` limit doesn't require a `-1`

c/dynalloc.c
```diff
@@ -1164,6 +1165,274 @@ int dynallocUnallocDatasetByDDName(const DynallocDDName *ddName,
...
+    strncat(buffer, field, bufferSize - strlen(buffer));
+  }
+
+  /* Initialize record format */
+  if (request->recordFormat != NULL) {
+    snprintf(field, sizeof(field)-1, "RECFM(%s) ", request->recordFormat);
+    strncat(buffer, field, bufferSize - strlen(buffer));
```
ifakhrutdinov: if `strlen(buffer)` is greater than `bufferSize`, the result will wrap around, you'll get a huge positive number and a buffer overflow.

ifakhrutdinov (changes requested):

PR #111, opened 2020-01-24
Update z/OS check

ifakhrutdinov:


ifakhrutdinov:
Adding @1000TurquoisePogs . We need someone with write access to approve to merge this.

PR #114, opened 2020-01-27
Log format control

sakeerthy:


c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+static void getLocationData(char *path, int line, char **locationInfo, uint64 compID, LoggingComponent *component) {
+
+  char prefix[128]; 
+  char suffix[128]; 
+
+  unsigned int id = compID & 0xFFFFFFFF;
+  if(compID >= LOG_PROD_COMMON && compID < LOG_PROD_ZIS) {
+    snprintf(prefix,5,"_zcc");
+    switch (id) { // most of these don't actally log, but are defined in logging.h
+      case 0x10000: snprintf(suffix,6,"alloc");
```
daveyc: Please use constants

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+
+   getSTCK(&stck);
+   int64 unixTime = stckToUnix(stck); // gets unixTime
+   stckToUnixSecondsAndMicros(stck, &out_secs, &out_micros); // calculates microseconds
+   out_micros = out_micros/1000;
+   unixToTimestamp((uint64) unixTime, timestamp);
+
+   // format: 2019-03-19 11:23:57.776
+   snprintf(date,24, "%.4s-%.2s-%.2s %.2s:%.2s:%.2s.%.3d"
+      ,timestamp, timestamp+4,timestamp+6,timestamp+8,timestamp+10,timestamp+12,(unsigned int) out_micros);
```
daveyc: It's bad practice to hard code offsets or lengths. If you need reformat a data format I suggest using a struct and/or a new function to do that formatting like `strftime()`.

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+    snprintf(suffix,strlen(component->name),"%s",(strrchr(component->name, '/')+1)); // given more characters than it writes
+    snprintf(prefix,strlen(component->name)-strlen(suffix),"%s",component->name);
+  }
+
+  char *filename;
+  filename = strrchr(path, '/'); // returns a pointer to the last occurence of '/'
+  filename+=1; 
+  //               formatting + prefix + suffix + filename + line number
+  int locationSize =  7  + strlen(prefix) + strlen(suffix) + strlen(filename) + 5; 
+  *locationInfo = (char*) safeMalloc(locationSize,"locationInfo");
```
daveyc: Memory leak!

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+  else if (compID > LOG_PROD_PLUGINS) {
+    snprintf(suffix,strlen(component->name),"%s",(strrchr(component->name, '/')+1)); // given more characters than it writes
+    snprintf(prefix,strlen(component->name)-strlen(suffix),"%s",component->name);
+  }
+
+  char *filename;
+  filename = strrchr(path, '/'); // returns a pointer to the last occurence of '/'
+  filename+=1; 
+  //               formatting + prefix + suffix + filename + line number
+  int locationSize =  7  + strlen(prefix) + strlen(suffix) + strlen(filename) + 5;
```
daveyc: Don't use hard coded lengths

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+    snprintf(prefix,strlen(component->name)-strlen(suffix),"%s",component->name);
+  }
+
+  char *filename;
+  filename = strrchr(path, '/'); // returns a pointer to the last occurence of '/'
+  filename+=1; 
+  //               formatting + prefix + suffix + filename + line number
+  int locationSize =  7  + strlen(prefix) + strlen(suffix) + strlen(filename) + 5; 
+  *locationInfo = (char*) safeMalloc(locationSize,"locationInfo");
+  snprintf(*locationInfo,locationSize+1," (%s:%s,%s:%d) ",prefix,suffix,filename,line);
```
daveyc: `sprintf` is aware of the null terminatior.

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+
+    char timestamp[24] = { 0 }; // formatted date is 24 characters
+    getLogTimestamp(timestamp); // UTC time
+
+    pthread_t threadID = pthread_self();
+    char thread[10];
+    snprintf(thread,10,"%d",threadID);
+
+    //    timestamp + " <ZWESA1:thread> " + user + log level + locationInfo + " " + format string
+    int newSize = 24 + 11 + 10 + 7 + strlen(logLevel) + strlen(locationInfo) + 1 + strlen(formatString); //accounts for formatting I know it's garbage
```
daveyc: Hard coded lengths. This code is very confusing even with the comment.

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+
+    char logLevel[8] = { 0 }; // logLevel will be a max of 8 characters
+    getLevel(level, logLevel);
+
+    char timestamp[24] = { 0 }; // formatted date is 24 characters
+    getLogTimestamp(timestamp); // UTC time
+
+    pthread_t threadID = pthread_self();
+    char thread[10];
+    snprintf(thread,10,"%d",threadID);
```
daveyc: Use sizeof thread

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+}
+
+static void getLocationData(char *path, int line, char **locationInfo, uint64 compID, LoggingComponent *component) {
+
+  char prefix[128]; 
+  char suffix[128]; 
+
+  unsigned int id = compID & 0xFFFFFFFF;
+  if(compID >= LOG_PROD_COMMON && compID < LOG_PROD_ZIS) {
+    snprintf(prefix,5,"_zcc");
```
daveyc: Hard coded length

c/logging.c
```diff
@@ -647,8 +652,118 @@ int logGetLevel(LoggingContext *context, uint64 compID){
...
+      case 0x20000: snprintf(suffix,5,"ctds");
+                    break;
+      case 0x30000: snprintf(suffix,9,"security");
+                    break;
+      case 0x40000: snprintf(suffix,9, "unixfile");
+                    break;
+    }
+  }
+  else if (compID > LOG_PROD_PLUGINS) {
+    snprintf(suffix,strlen(component->name),"%s",(strrchr(component->name, '/')+1)); // given more characters than it writes
```
daveyc: I think this should this be `sizeof(component->name)`

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+    char timestamp[24] = { 0 }; // formatted date is 24 characters
+    getLogTimestamp(timestamp); // UTC time
+
+    pthread_t threadID = pthread_self();
+    char thread[10];
+    snprintf(thread,10,"%d",threadID);
+
+    //    timestamp + " <ZWESA1:thread> " + user + log level + locationInfo + " " + format string
+    int newSize = 24 + 11 + 10 + 7 + strlen(logLevel) + strlen(locationInfo) + 1 + strlen(formatString); //accounts for formatting I know it's garbage
+    char *newFormatString = (char *) safeMalloc(sizeof(char) * newSize,"newFormatString");
```
daveyc: What is the point of `sizeof(char) * newSize` when sizeof(char) == 1? newSize is sufficient.

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
       lastResortLog(message);
       return;
     }
+
+    ACEE *acee;
+    acee = getAddressSpaceAcee();
+    char user[7] = { 0 }; // wil this always be 7?
```
daveyc: This length does not include the null terminator.

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+    snprintf(user,7,acee->aceeuser+1);
```
daveyc: Use sizeof(user).

c/logging.c
daveyc: Typo in this comment. Should be `will`

c/logging.c
```diff
@@ -685,14 +799,41 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+    char *locationInfo; // largest possible variation in size
+    getLocationData(path,line,&locationInfo,compID,component); // location info is allocated in getLocationData
+
+    char logLevel[8] = { 0 }; // logLevel will be a max of 8 characters
+    getLevel(level, logLevel);
+
+    char timestamp[24] = { 0 }; // formatted date is 24 characters
+    getLogTimestamp(timestamp); // UTC time
+
+    pthread_t threadID = pthread_self();
```
daveyc: It's my understanding that zowelog is used by both LE and Metal/C runtime environments. pthread_self() is not valid for Metal/C so you will need to handle both LE and Metal/C (using the current TCB address).

PR #115, opened 2020-01-29
Add helper functions to test caller's environment

ifakhrutdinov:
The following new functions has been added:
* A function to test whether the caller is running in SRB
* A function to test whether the caller is in cross-memory mode
* A function to test whether the caller is holding a CPU, CMS, CML or local lock

### DIFF
--- a/c/zos.c
+++ b/c/zos.c
@@ -1532,9 +1532,12 @@ bool isCallerCrossMemory(void) {
       /* assume not cross-memory */
       "         LA    15,0                                                     \n"
       /* get HASN */
-      "         LA    1,0                                                      \n"
-      "         USING PSA,1                                                    \n"
+      "         USING PSA,0                                                    \n"
+#ifdef _LP64
+      "         LLGT  1,PSAAOLD                                                \n"
+#else
       "         L     1,PSAAOLD                                                \n"
+#endif
       "         USING ASCB,1                                                   \n"
       "         LLH   1,ASCBASID                                               \n"
       "         DROP  1                                                        \n"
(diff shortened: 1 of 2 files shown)


c/zos.c
```diff
@@ -1452,13 +1452,136 @@ char *getASCBJobname(ASCB *ascb){
...
+      "         LA    15,0                                                     \n"
+      /* get HASN */
+      "         LA    1,0                                                      \n"
+      "         USING PSA,1                                                    \n"
+      "         L     1,PSAAOLD                                                \n"
+      "         USING ASCB,1                                                   \n"
+      "         LLH   1,ASCBASID                                               \n"
+      "         DROP  1                                                        \n"
+      /* get PASN and compare with HASN */
+      "         LA    2,0                                                      \n"
```
gadyck: Not necessary to clear R2, EPAR sets a 32 bit value.

c/zos.c
```diff
@@ -1452,13 +1452,136 @@ char *getASCBJobname(ASCB *ascb){
...
+      "         EPAR  2                                                        \n"
+      "         CLR   1,2                                                      \n"
+      "         JNE   &LXMEM                                                   \n"
+      /* get SASN and compare with HASN */
+      "         LA    2,0                                                      \n"
```
gadyck: Not necessary to clear R2, ESAR sets a 32 bit value.

c/zos.c
```diff
@@ -1452,13 +1452,136 @@ char *getASCBJobname(ASCB *ascb){
...
+      ASM_PREFIX
+      /* generate unique branch labels */
+      "&LX      SETA  &LX+1                                                    \n"
+      "&LXMEM   SETC  'LXMEM&LX'                                               \n"
+      "&LEXIT   SETC  'LEXIT&LX'                                               \n"
+      "         PUSH  USING                                                    \n"
+      /* assume not cross-memory */
+      "         LA    15,0                                                     \n"
+      /* get HASN */
+      "         LA    1,0                                                      \n"
```
gadyck: Do not reference the PSA this way.  Remove the LA and use "USING PSA,0"

PR #116, opened 2020-01-29
Expose wrappers for CPOOL

ifakhrutdinov:
**Overview**

The purpose of this pull-request is to expose the CPOOL wrappers and additionally, ensure they work in SRB and cross-memory. The code is almost identical to the code in https://github.com/zowe/zowe-common-c/blob/master/c/cmutils.c

Differences with the existing code in cmutils:
* SRB/xmem support for `cellpoolBuild`
* New names
* Doxygen documentation
* A new helper function for cell size calculation

### DIFF
--- a/c/cellpool.c
+++ b/c/cellpool.c
@@ -349,14 +349,24 @@ zos.o \
 #define CELLPOOL_TEST_STATUS_OK        0
 #define CELLPOOL_TEST_STATUS_FAILURE   8
 
+#define CELLPOOL_TEST_PRIMARY_SIZE      10
+#define CELLPOOL_TEST_SECONDARY_SIZE    2
+#define CELLPOOL_TEST_CELL_SIZE         512
+#define CELLPOOL_TEST_SUBPOOL           132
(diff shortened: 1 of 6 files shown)


c/cellpool.c
```diff
@@ -0,0 +1,516 @@
...
+    if (cells[i] == NULL) {
+      printf("error: cellpoolFree test failed (alloc 1), cell #%d\n", i);
+      status = CELLPOOL_TEST_STATUS_FAILURE;
+      break;
+    }
+  }
+
+  if (status == CELLPOOL_TEST_STATUS_OK) {
+
+    for (int i = 0; i < sizeof(cells) / sizeof(cells[0]); i++) {
```
daveyc: An ARRAY_SIZE() macro would make this less tedious and error prone.

ifakhrutdinov:
I may reconsider the `NULL` value for cell pool IDs, perhaps `0` is less error prone.

```c
if (cpid) {
  cellpoolDelete(cpid);
}
```

PR #120, opened 2020-02-15
Expose additional functionality in collections and cross-memory

ifakhrutdinov:
#### Overview
This pull request add the following features:
* Ability to use the lock-free queue intrusively which allows a more flexible storage management on the user's side
* Functions to copy to/from foreign address space using destination/source keys and ALETs

### DIFF
--- a/c/collections.c
+++ b/c/collections.c
@@ -1137,16 +1137,13 @@ static int compareAndSwapTriple(long *oldCounter, long newCounter, long *counter
   return status;
 }
 
-void qInsert(Queue *q, void *newData){
+void qEnqueue(Queue *q, QueueElement *newElement) {
 
   union {
     long long alignit;
     CSTSTParms parms;
   };
 
-  QueueElement *newElement = NULL;
-  newElement = (QueueElement*)safeMalloc(sizeof(QueueElement),"Q Element");
-  newElement->data = newData;
   newElement->next = NULL;
 
   /* Note: The PLO compare value must alway be the first data fetched when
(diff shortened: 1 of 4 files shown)


h/collections.h
```diff
@@ -259,9 +259,52 @@ ZOWE_PRAGMA_PACK_RESET
...
+ */
 void destroyQueue(Queue *q) QueueAmode64;
+
+#ifdef __ZOWE_OS_ZOS
+
+/**
+ * Enqueue a queue element.
+ * @param q Queue to which the new element will be enqueued.
+ * @param newElement Queue element to be enqueued. WARINIG: it must reside on
+ * a double word boundary, otherwise a specification exception (S0C6 ABEND)
```
daveyc: I'm not sure if double word alignment is required (ABEND0C6) or recommended for performance reasons. Maybe Greg can comment?

c/collections.c
```diff
@@ -1343,11 +1350,17 @@ void *qRemove(Queue *q){
...
+
   void *result = NULL;
-  if (currentHead)
-  {
-    result = currentHead->data;
-    safeFree31((char*)currentHead,sizeof(QueueElement));
+  QueueElement *element = qDequeue(q);
+  if (element) {
+    result = element->data;
+    safeFree((char *)element, sizeof(QueueElement));
```
daveyc: Nitpick! This cast to (char *) bothers me and should be fixed in the future. It's not a breaking change and removes an irritating requirement for an unnecessary cast.

c/collections.c
```diff
@@ -1237,7 +1234,17 @@ void qInsert(Queue *q, void *newData){
   return;
 }
 
-void *qRemove(Queue *q){
+void qInsert(Queue *q, void *newData) {
```
daveyc: newData should be const

ifakhrutdinov: I'd need to think this through. If we make this const, we need to make the field const and `qRemove` will need to return a const value which will break existing code. I don't see how we can solve this without casting at some point.

c/collections.c
ifakhrutdinov: I agree.

c/collections.c
daveyc: I think a cast is fine. I would  cast to (void *) when assigning the data element and leave everything else as is. `qRemove` should remain as is. That's pretty standard. Making newData const signifies the intent of the function in that it doesn't mutate newData.

PR #118, opened 2020-02-02
Locked environment support in recovery

ifakhrutdinov:
**WARNING**
The pull-request required the changes from #116. When reviewing the code, please disregard the following files:
* `zowe-common-c/c/cellpool.c`
* `zowe-common-c/c/cmutils.c`
* `zowe-common-c/h/cellpool.h`

**Overview**
This pull request adds the support of locked environments in recovery. In order to achieve that, the recovery code will rely on cell pools when built with the corresponding compiler time option (it will also require a new dependency - `zowe-common-c/c/cellpool.c`)

**Major changes**
* Users have the ability to pass their own recovery context storage and cell pools for the recovery states
* For backward compatibility the code detects if a state has been allocated using cell pools or storage obtain (for backward compatibility)
* The code detects if there is an FRR already in place before choosing between ESTAEX and FRR

**Testing**
* The code has been tested using unit tests from https://github.com/ifakhrutdinov/zowe-common-c/blob/feature/zos-tests/test/recoverytest.c
* The code has been tested with the current ZIS and ZSS versions

**Potential discussion items**
* Default subpool for the state cell pool (the current one is 0)
* Time to refactor `establishRouterInternal` (aka formal `recoveryEstablishRouter`)?

PR #121, opened 2020-02-15
Support of locked callers in cross-memory

ifakhrutdinov:
### WARNING
The pull-request required the changes from #118 and #120 . When reviewing the code, please disregard the following files:

* zowe-common-c/c/cmutils.c
* zowe-common-c/c/collections.c
* zowe-common-c/c/recovery.c
* zowe-common-c/h/cellpool.h
* zowe-common-c/h/cmutils.h
* zowe-common-c/h/collections.h
* zowe-common-c/h/recovery.h

**Files to be reviewed in this pull-request**:
* zowe-common-c/c/crossmemory.c
* zowe-common-c/h/crossmemory.h

#### Overview
Due to the need to support locked callers in PC-ss, the cross-memory code has been changed to use cell pool based recovery and cell pool based PC-ss stack storage. Additionally, the message service has also been changed to use cell pool for its message queue.

#### Details
* To support cell pool stack storage, the cross memory server allocates a cell pool upon start up. The cell pool size can be adjusted from outside (via a PARMLIB parameter in ZIS). The PC-ss prolog/epilog code performs a conditional GET request to ensure that the cell pool doesn't get expanded which allows it to be used in locked environments. Since the GET request is conditional, normal callers will not experience any ABENDs either in case there is no region storage left. The PC routine will always terminate gracefully with a non zero return code.
* The recovery states are now shared among all PC-ss calls. The cross-memory server allocates a recovery pool at start up, this size of the pool can be adjusted via a parameter.
* The message queue relies on 2 cell pools. The main cell pool is used unconditionally if the caller is unlocked and conditionally if the caller is locked. If the unconditional request fails, the recovery will handle that and the PC call will return a bad RC. If the conditional call fails (locked callers), the code will conditionally attempt to use the fall back pool. This increases the chances of having enough storage for locked callers.

#### Minor changes:
* The PC-ss and PC-cp prolog/epilog HLASM code has details comments now.

### DIFF
--- a/c/crossmemory.c
+++ b/c/crossmemory.c
@@ -29,6 +29,7 @@
 
 #include "zowetypes.h"
 #include "alloc.h"
+#include "cellpool.h"
 #include "cmutils.h"
 #include "crossmemory.h"
 #include "lpa.h"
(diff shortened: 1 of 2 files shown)


c/crossmemory.c
```diff
@@ -1192,204 +1200,245 @@ typedef struct PCHandlerParmList_tag {
...
-LSSRET   DS    0H                                 \n\
-         DROP                                     \n\
+"        PUSH  USING                                                    \n\
+LSSPRLG  LARL  10,LSSPRLG          establish addressability using r10   \n\
+         USING LSSPRLG,10                                               \n\
+         LGR   5,4                 save latent parm as CPOOL uses r4    \n\
+         USING PCLPARM,5                                                \n\
+         LLGT  2,PCLPPRM1          load global area from latent parm    \n\
+         USING GLA,2                                                    \n\
+         SAM31                                                          \n\
```
gadyck: While not documented, CPOOL GET and FREE *support* AMODE(64) callers as long as SYSSTATE AMODE64=YES has been specified.

c/crossmemory.c
```diff
@@ -1192,204 +1200,245 @@ typedef struct PCHandlerParmList_tag {
...
+         SYSSTATE AMODE64=NO                                            \n\
+         CPOOL GET,C,CPID=GLASTCP,REGS=USE   allocate a stack cell      \n\
+         SAM64                                                          \n\
+         SYSSTATE AMODE64=YES                                           \n\
+         DROP  2                   drop, r2 is changed by CPOOL         \n\
+         LTGFR 1,1                 check if we have a cell              \n\
```
gadyck: Why not do LTGFR 13,1 here, and delete the LGR following label LSSINIT?

c/crossmemory.c
```diff
@@ -1192,204 +1200,245 @@ typedef struct PCHandlerParmList_tag {
...
+         LGR   3,13                put stack to r3 for CPOOL            \n\
+         LGR   4,15                save RC as CPOOL uses r15            \n\
+         LARL  10,&CCN_BEGIN       establish addressability             \n\
+         USING &CCN_BEGIN,10                                            \n\
+         USING PCHSTACK,13                                              \n\
+         LG    5,PCHPLLPL          load latent parm from PC parm        \n\
+         USING PCLPARM,5                                                \n\
+         LLGT  2,PCLPPRM1          load global area from latent parm    \n\
+         USING GLA,2                                                    \n\
+         SAM31                                                          \n\
```
gadyck: Same comment as CPOOL GET, above

c/crossmemory.c
```diff
@@ -1464,24 +1523,78 @@ void post(ECB * __ptr32 ecb, int code) {
...
+
+  if (recoveryRC == RC_RCV_OK) {
+    recoveryPop();
+  }
+
+  if (element) {
+    memset(element, 0, sizeof(*element));
+    element->cellPool = cpid;
+  } else {
+    return RC_CMS_NO_STORAGE_FOR_MSG;
```
gadyck: Add *result = NULL;

ifakhrutdinov:
@gadyck thank you for the review and suggestions. I will add them in follow up PR as they don't seem to be bugs (at least serious one).

PR #125, opened 2020-02-21
printfs to zowelogs

sakeerthy:
Still needs work on some files compIds

WIP


c/alloc.c
```diff
@@ -139,7 +140,7 @@ char *allocECSA(int size, int key){
   case 2:
     return getmain31Key2(size,241);
   default:
-    printf("PANIC cannot allocate ECSA in key=%d\n",key);
+    zowelog(NULL, LOG_COMP_ALLOC, ZOWE_LOG_SEVERE, "PANIC cannot allocate ECSA in key=%d\n", key);
```
ifakhrutdinov: I think it's a bad idea to call zowelog in alloc and possibly other library function. They shouldn't depend on logging. Plus, from what I've seen in another related PR, you call `safeMalloc` in your logger code. So you logger code that calls `safeMalloc` will call the logger code again and so on, so forth until you run our of stack and ABEND.

PR #126, opened 2020-02-27
zowelog zss format handler

sakeerthy:
Dependent on https://github.com/zowe/zss/pull/160


c/crossmemory.c
```diff
@@ -2808,12 +2808,13 @@ static void printDisplayConfigCommandResponse(CrossMemoryServer *server, CMSWTOR
...
-  zowelog(NULL, LOG_COMP_ID_CMS, ZOWE_LOG_INFO,
 #ifdef CROSS_MEMORY_SERVER_DEBUG
-      CMS_LOG_DISP_CMD_RESULT_MSG"Server name - \'%16.16s\' (debug mode)\n"
+  char* formatStringPrefix = CMS_LOG_DISP_CMD_RESULT_MSG"Server name - \'%16.16s\' (debug mode)\n";
 #else
-      CMS_LOG_DISP_CMD_RESULT_MSG"Server name - \'%16.16s\'\n"
+  char* formatStringPrefix = CMS_LOG_DISP_CMD_RESULT_MSG"Server name - \'%16.16s\'\n";
 #endif
+
+  zowelog(NULL, LOG_COMP_ID_CMS, ZOWE_LOG_INFO, strcat(formatStringPrefix,
```
ifakhrutdinov: Do you `strcat` into a string literal? Have you tested this code? I would expect an S0C4.

ifakhrutdinov (changes requested):
I'll be reviewing this code next week but what I've already seen is concerning.

c/crossmemory.c
daveyc: Agreed! it's a buffer overflow into the literal pool! `strcat` is a code smell and should never be used.

h/logging.h
```diff
@@ -93,14 +93,34 @@ ZOWE_PRAGMA_PACK_RESET
...
+
 #define LOG_DEST_DEV_NULL      0x008F0000
 #define LOG_DEST_PRINTF_STDOUT 0x008F0001
 #define LOG_DEST_PRINTF_STDERR 0x008F0002
 
 struct LoggingContext_tag;
 
 typedef void (*LogHandler)(struct LoggingContext_tag *context,
-                           LoggingComponent *component, 
+                           LoggingComponent *component,
```
ifakhrutdinov: This breaks all the code that uses logging, we shouldn't make such changes in minor releases. If you kept compatibility you wouldn't have to change the cross memory code. Imagine that someone uses this API, do they now need to change their code? 

What about binaries? You will break existing plugins because you'll call theirs handler assuming it's something else.

Please reconsider this design. You may want to introduce another handler type, allow new code to set it and check at runtime whether you need to caller the old handler type or the new one. That way you don't have to change so much code and break old existing code.

h/logging.h
```diff
@@ -93,14 +93,34 @@ ZOWE_PRAGMA_PACK_RESET
...
+						   char* path, int line,
```
ifakhrutdinov: Do not mix tabs and spaces. This applies to all the changes in this PR.

h/logging.h
```diff
@@ -332,8 +360,8 @@ int logGetLevel(LoggingContext *context, uint64 compID);
 extern int logSetExternalContext(LoggingContext *context);
 extern LoggingContext *logGetExternalContext();
 
-void printStdout(LoggingContext *context, LoggingComponent *component, void *data, char *formatString, va_list argList);
-void printStderr(LoggingContext *context, LoggingComponent *component, void *data, char *formatString, va_list argList);
+void printStdout(LoggingContext *context, LoggingComponent *component, char* path, int line, int level, uint64 compID, void *data, char *formatString, va_list argList);
```
ifakhrutdinov: Do not change existing APIs.

h/logging.h
```diff
@@ -297,15 +317,23 @@ bool logShouldTraceInternal(LoggingContext *context, uint64 componentID, int lev
...
+char *standardDumperFunction(char *workBuffer, int workBufferSize,
+                                    void *data, int dataSize, int lineNumber);
 #define LOGCHECK(context,component,level) \
   ((component > MAX_LOGGING_COMPONENTS) ? \
    (context->applicationComponents[component].level >= level) : \
    (context->coreComponents[component].level >= level) )
 
+
+#define zowelog(context, compID, level, formatString, ...) \
+        _zowelog(context, compID, __FILE__, __LINE__, level, formatString, ##__VA_ARGS__);
```
ifakhrutdinov: By introducing a macro we will break existing plugins because they assume that zowelog has fewer args. It may be better to have `zowelog2` to keep the backward compatibility.

PR #127, opened 2020-03-03
Use __troo to perform fast EBCDIC1047-UTF8 character conversion

ifakhrutdinov:
#### Overview
Use the TROO instruction with a "EBCDIC 1047 to ISO/IEC 8859-1" translation table to do a fast EBCDIC to UTF8 character translation when possible.

#### Details
* The translation table had to be coped from xlate_tables.c since the tables defined in that file don't guarantee a double word alignment

#### Potential future changes
* Dynamic translation tables generated using CUNLCNV
* Refactoring of `convertCharset` - the function seems to be pretty heave at the moment


c/charsets.c
```diff
@@ -116,6 +116,33 @@ int convertCharset(char *input,
 
 #elif defined(__ZOWE_OS_ZOS)
 
+static const union {
+  char table[256];
+  uint64_t alignment;
```
daveyc: You can align storage using the [aligned variable attribute ](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.cbclx01/var_attrib_aligned.htm). It annoyingly emits a warning message which I suppress. It would be interesting to measure if there is any performance difference between 4 byte, 8 byte and 16 byte alignment in 64-bit code. 

I have anecdotal evidence that a double world aligned `ex` instruction was much quicker but that was a long time ago and is probably obsolete.

ifakhrutdinov: I did use it initially, then spent lots of time trying to suppress that message (turns out you can't use pragmas to do that) and decided to just use a union.

gadyck: Rather than duplicate the table I would have suggested you align the existing tables in xlate_tables.c.  Now the same table exists twice.

FYI, I usually use a union, like you did, to force alignment of storage.

ifakhrutdinov: @gadyck , I agree, that would be a better approach. I'm thinking, if possible, how we could do that without having to change any existing code that uses these tables.

PR #128, opened 2020-03-09
Password Reset

sakeerthy:
Goes along with:
https://github.com/zowe/zss/pull/161
https://github.com/zowe/zowe-common-c/pull/128
https://github.com/zowe/zlux-app-server/pull/106
https://github.com/zowe/zss-auth/pull/29
https://github.com/zowe/zlux-server-framework/pull/181
https://github.com/zowe/zlux-app-manager/pull/193


c/httpserver.c
```diff
@@ -3194,7 +3200,13 @@ static int handleHttpService(HttpServer *server,
 #endif
   if (request->authenticated == FALSE){
     /* could make this parameterizable */
-    respondWithError(response,401,"Not Authorized");
+    switch (racfRC) {
+      case SERVICE_AUTH_PASSWORD_EXPIRED:
+        respondWithError(response, 428, "Password Expired");
```
1000TurquoisePogs: Define this number

c/httpserver.c
```diff
@@ -3194,7 +3199,13 @@ static int handleHttpService(HttpServer *server,
 #endif
   if (request->authenticated == FALSE){
     /* could make this parameterizable */
-    respondWithError(response,401,"Not Authorized");
+    switch (racfRC) {
```
ifakhrutdinov: It's best to check both SAF RC and RACF RC to determine the "expired" condition. If the RACF RC matches the value under a different SAF RC you will report the wrong error.

c/httpserver.c
```diff
@@ -3184,7 +3189,7 @@ static int handleHttpService(HttpServer *server,
         break;
       } /* else fall through */
     case SERVICE_AUTH_TOKEN_TYPE_LEGACY:
-      request->authenticated = serviceAuthNativeWithSessionToken(service,request,response,&clearSessionToken);
+      request->authenticated = serviceAuthNativeWithSessionToken(service,request,response,&clearSessionToken, &safRC, &racfRC, &racfRSN);
```
ifakhrutdinov: This will need refactoring at some point. All those functions now have to have the safRC, racfRC, racfRSN parameters which don't directly relate to the functions.

We're leaking abstraction here, I don't think we should expose details like SAF, RACF. Perhaps returning some internal reason code and the translating it into the corresponding HTTP response would be better. This would also apply to the `safAuthenticate` call.

c/httpserver.c
```diff
@@ -83,6 +83,17 @@ typedef int time_t;
 #define APF_AUTHORIZED  1
 #endif
 
+#define SAF_RC_PASSWORD_EXPIRED    8
+#define RACF_RC_PASSWORD_EXPIRED   12
+
+typedef struct AuthResponse_tag {
```
1000TurquoisePogs: This struct and defines are going to see a lot of use in the future I think, so this should be in a .h file rather than embedded in this .c file.
I'm not sure which .h file is best: is it one related to the httpserver, or one related to auth

sakeerthy: irek said to leave it in the .c file since its only used here

ifakhrutdinov: @1000TurquoisePogs I think `AuthResponse` is very specific to `httpserver.c`. We should move it when we need to reuse it, right now there is only one user.

ifakhrutdinov:
@sakeerthy can you please go over the comments and resolve the ones which are resolved and reply to the ones which aren't resolved?

PR #130, opened 2020-03-18
Warn less frequently to not fill up log quickly

1000TurquoisePogs:
Previous code prints this line whenever attls is not set up:
`*** WARNING: Connection is insecure! TLS needed but not found on socket. ***`
Let's print it only once every 25 times.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -209,6 +209,14 @@ static char crlf[] ={ 0x0d, 0x0a};
 #define DEFAULT_UMASK 0022
 #endif
 
+//No need to fill logs with the same message, warn based on a desired frequency
+#ifdef __ZOWE_OS_ZOS
+#ifndef TLS_WARN_FREQUENCY
+#define TLS_WARN_FREQUENCY 25
+#endif
+static int tlsWarnCounter = 0;
+#endif
+
 static int64 getFineGrainedTime();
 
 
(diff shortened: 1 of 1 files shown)


c/httpserver.c
```diff
@@ -209,6 +209,14 @@ static char crlf[] ={ 0x0d, 0x0a};
 #define DEFAULT_UMASK 0022
 #endif
 
+//No need to fill logs with the same message, warn based on a desired frequency
+#ifdef __ZOWE_OS_ZOS
```
1000TurquoisePogs: Should this be `#if defined(__ZOWE_OS_ZOS) || defined(USE_RS_SSL)` instead?

ifakhrutdinov: Yes, based on the code, we need to use these `#define`s because we won't be able to compile if `__ZOWE_OS_ZOS` is not defined.

c/httpserver.c
```diff
@@ -209,6 +209,14 @@ static char crlf[] ={ 0x0d, 0x0a};
...
+#ifndef TLS_WARN_FREQUENCY
+#define TLS_WARN_FREQUENCY 25
+#endif
+static int tlsWarnCounter = 0;
```
ifakhrutdinov: Please use `unsigned int`, otherwise if we ever hit integer overflow, it will result in undefined behavior.

PR #133, opened 2020-04-03
Use ASCII mapping for character conversion; fix integer overflow in json

ifakhrutdinov:
#### Overview
The pull request fixes the following items:
* All fast path conversion are done to ASCII only, the rest of the characters will require the slow path.
  * `z/OS DFSMS/MVS Using Magnetic Tapes, Appendix D` has been used as a reference for the EBCDIC-ASCII translation table
* Unsigned integer overflow is fixed in the json code. This overflow was causing ZSS crashes when non-UTF8 characters were encountered.

#### Testing:
* The result JSON of reading a data set containing all 0x00 - 0xFF characters in column 10 has been compared with the non-TROO ZSS results
image

* Additionally, converted buffers were dumped in hex and compared to make sure they are identical before going to JSON
 
Here's a comparison of the new table (on the right) with `iconv` generated table (on the left), we're only suppressing some conversions (`0x1A`), no other codes have been changed:
image

Fixes zowe/zss#169


ifakhrutdinov:
@daveyc @1000TurquoisePogs 

* There are characters which are considered irregular (e.g. `[` and `]`), would that be safe to use the output of CUNLCNV to define mapping for them or could they change from system to system?
* Since LF is returned as 0xC385 from CUNLCNV  but the table defines it as 0x0A, I chose to make it "undefined" and let CUNLCNV  do the work. Do you think it'd be safe to use 0x0A since it's listed as compatible with UTF8?

daveyc:
> 
> 
> @daveyc @1000TurquoisePogs
> 
>     * There are characters which are considered irregular (e.g. `[` and `]`), would that be safe to use the output of CUNLCNV to define mapping for them or could they change from system to system?
> 
>     * Since LF is returned as 0xC385 from CUNLCNV  but the table defines it as 0x0A, I chose to make it "undefined" and let CUNLCNV  do the work. Do you think it'd be safe to use 0x0A since it's listed as compatible with UTF8?

If we are only supporting IBM-1047 then I don't see a problem. A one-to-one translation is simple so square brackets are not an issue.

ifakhrutdinov:
Being even more conservative, perhaps we should also exclude `^`, `¬` and other symbols from the "gray" area, see [EBCDIC code page layout](https://en.wikipedia.org/wiki/EBCDIC).

@daveyc @gadyck what do you think?

daveyc:
I don't see a valid reason for not doing a full conversion from EBCDIC to ASCII. I don't get the grey area!

daveyc:
Slow path I really don't know

daveyc:
The conversion tables should handle this if they are correct

daveyc:
Does the JSON code not escape non-UTF8 code?

ifakhrutdinov:
> 
> 
> Does the JSON code not escape non-UTF8 code?

well, it didn't seem to do that before the fix and it doesn't do it now, I only fixed the crash and infinite loop

1000TurquoisePogs:
When you translate a `[` or `]` I believe they sometimes show up as an accented Y such as `Ý`
But, I don't recall the circumstance. Maybe its 37 vs 1047, or some other common codepage.
There's a question of performance vs accuracy here, since I've kind of become accustomed to seeing the accented Ys and knowing what was meant. Should we keep `[` and `]` and even `^` in for performance? I think yes, since if the cp really is 1047, then the mapping will be correct.

1000TurquoisePogs:
Yeah: it is the case of 1047 vs 37. This page shows it: https://en.wikipedia.org/wiki/EBCDIC_1047

daveyc:
The only difference between IBM-037 and IBM-1047 are square brackets and the not character which is translated to a ^. I've spoken to Irek about this and in a future release it may be possible to create dynamic translation tables to support most EBCDIC character sets when converting to ISO8859-1 in the fast path.

ifakhrutdinov:
Here's ISO8859-i - UTF8 translation table comparison, I think we can see which ones will be safe to enable (the ones occupying 1 byte)
image

@1000TurquoisePogs what do you think?

PR #137, opened 2020-04-24
added a new option for saf auth to hide logs

jordanfilteau1995:
https://github.com/zowe/zss/pull/175

### DIFF
--- a/h/zos.h
+++ b/h/zos.h
@@ -1197,6 +1197,7 @@ int safVerify5(int options,
 
 #define AUTH_DATASET_VSAM 0x80
 #define AUTH_DATASET_TAPE 0x40
+#define AUTH_WITHOUT_LOG  0x02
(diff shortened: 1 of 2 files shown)


h/zos.h
```diff
@@ -1195,6 +1195,7 @@ int safVerify5(int options,
 #define SAF_AUTH_ATTR_UPDATE      0x04
 #define SAF_AUTH_ATTR_READ        0x02
 
+#define AUTH_WITHOUT_LOG  0x10
```
ifakhrutdinov, line 1198: @jordanfilteau1995 I think it's better to place this between `AUTH_DATASET_TAPE` and `AUTH_SUPERVISOR` and give it value 0x02. Otherwise it falls into the range of the "DATASET" options. Other than that the PR looks good to me.

PR #140, opened 2020-05-19
Control LPA dev mode with a run-time option

ifakhrutdinov:
**Overview**

This pull-request adds a new control flag. The flag allows users to enable the LPA dev mode without recompiling the module. In the LPA dev mode, the ZWESIS01 module will always be refreshed at start up and removed from LPA upon termination. This latter should make the memory footprint smaller in a development environment.

Since removing modules from LPA is considered dangerous, a new warning message will be issued to the SYSLOG and SYSPRINT if the new mode is enabled:
```
ZWES0247W Development mode is enabled
```

### DIFF
--- a/h/crossmemory.h
+++ b/h/crossmemory.h
@@ -393,7 +393,8 @@ ZOWE_PRAGMA_PACK_RESET
 #define CMS_SERVER_FLAG_COLD_START            0x00000001
 #define CMS_SERVER_FLAG_DEBUG                 0x00000002
 #define CMS_SERVER_FLAG_CHECKAUTH             0x00000004
-#define CMS_SERVER_FLAG_CLEAN_LPA             0x00000008
+#define CMS_SERVER_FLAG_DEV_MODE              0x00000008
+#define CMS_SERVER_FLAG_DEV_MODE_LPA          0x00000010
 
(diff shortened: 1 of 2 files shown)


h/crossmemory.h
```diff
@@ -848,6 +850,12 @@ CrossMemoryServerName cmsMakeServerName(const char *nameNullTerm);
 #define CMS_LOG_SRVC_ENTRY_OCCUPIED_MSG_TEXT    "Service entry %d is occupied"
 #define CMS_LOG_SRVC_ENTRY_OCCUPIED_MSG         CMS_LOG_SRVC_ENTRY_OCCUPIED_MSG_ID" "CMS_LOG_SRVC_ENTRY_OCCUPIED_MSG_TEXT
 
+#ifndef CMS_LOG_CLEAN_LPA_MSG_ID
+#define CMS_LOG_CLEAN_LPA_MSG_ID                CMS_MSG_PRFX"0247W"
+#endif
+#define CMS_LOG_CLEAN_LPA_MSG_TEXT              "Unsafe LPA development mode is enabled"
```
daveyc, line 856: What exactly is unsafe? Are we using some kind of brute force method that may result in system integrity problems?

ifakhrutdinov: The messages has been changed. Please take a look.

PR #149, opened 2020-07-16
Verify that the STC user has an OMVS segment

ifakhrutdinov:
### Overview
There's been a lot of confusion when users would start ZIS under a user without an OMVS segment and ZIS would silently crash. This pull-request adds an explicit BPX call which should verify if the STC user has an OMVS segment ans is able to perform BPX calls later in the program.

Fixes zowe/zss#189

### DIFF
--- a/c/crossmemory.c
+++ b/c/crossmemory.c
@@ -4068,7 +4068,7 @@ static bool isDubStatusOk(int *status, int *bpxRC, int *bpxRSN) {
           CMS_LOG_DEBUG_MSG_ID" BPXnQDB RV = %d, RC = %d, RSN = 0x%08X\n",
           *status, *bpxRC, *bpxRSN);
 
-  if (*status == dubFailRC){
+  if (*status == dubFailRC) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


c/crossmemory.c
```diff
@@ -4050,6 +4050,31 @@ static bool isModulePrivate(void) {
...
+static int checkOMVS(int *bpxRC, int *bpxRSN) {
+
+  int returnValue = 0;
+
+  const int timerRealType = 0;  /* ITIMER_REAL */
+  char result[32] = {0};        /* BPXYITIM */
+  void *resultAddr = result;
+
+  /* Use a getitimer() call to check OMVS. */
+  BPXGTR(&timerRealType, &resultAddr, &returnValue, bpxRC, bpxRSN);
```
daveyc, line 4070: There is a callable service that is designed with a query option to test if the user has access to OMVS [BPXnQDB - query dub](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxb100/qdb.htm)

PR #154, opened 2020-07-21
Add service to create tar file.

Akayeshmantha:
This PR will give the ability to end users to create tar files from folder content.
This is a requirement which was provided as a part of file-transfer app.

### DIFF
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -197,7 +197,10 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
   }
 }
 
-void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
+void createArchiveFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
+#ifdef METTLE
+  printf("create archive from directory unimplemented in metal\n");
+#else
   if (isDir(absolutePath)) {
 
       char fileNameBuffer[USS_MAX_PATH_LENGTH + 1] = {0};
@@ -231,6 +234,7 @@ void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolut
     respondWithJsonError(response, "Failed to identify a directory with the given name", 400, "Bad Request");
     return;
   }
+#endif
 }
 
 /* Modifies the mode of files/directories */
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -216,8 +216,8 @@ void createArchiveFromUnixDirectoryAndRespond(HttpResponse *response, char *abso
       char *tarFileName = (slashPos == -1) ? absolutePath : absolutePath + slashPos + 1;
       snprintf(fileNameBuffer, sizeof(fileNameBuffer), "%s%s", tarFileName, ".tar");
 
-      char *tarCommand = "/bin/sh -c 'tar -cf ";
-      snprintf(commandBuffer, sizeof(commandBuffer), "%s%s%s%s%s", tarCommand, fileNameBuffer, " ", absolutePath, "'");
+      char *tarCommand = "/bin/tar -cf ";
+      snprintf(commandBuffer, sizeof(commandBuffer), "%s%s%s%s", tarCommand, fileNameBuffer, " ", absolutePath);
 
       // system command will create the tar.
       system(commandBuffer);
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -199,24 +199,27 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
 
 void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
   if (isDir(absolutePath)) {
+
+      char fileNameBuffer[USS_MAX_PATH_LENGTH + 1] = {0};
+      char commandBuffer[USS_MAX_PATH_LENGTH + 1] = {0};
+
       int folderNameLen = strlen(absolutePath);
       if(folderNameLen == 0 || absolutePath == NULL) {
          respondWithJsonError(response, "Failed to idenity the folder pointed", 400, "Bad Request");
       }
+
       int slashPos = lastIndexOf(absolutePath, folderNameLen, '/');
       char *tarFileName = (slashPos == -1) ? absolutePath : absolutePath + slashPos + 1;
-      char finalFileName[strlen(tarFileName) + strlen(".tar")];
-      strcpy(finalFileName, tarFileName);
-      strcat(finalFileName,".tar");
-      char finalCommand[strlen("/bin/sh -c 'tar -cf ") + strlen(finalFileName) + strlen(absolutePath) + 2];
-      strcpy(finalCommand, "/bin/sh -c 'tar -cf ");
-      strcat(finalCommand,finalFileName);
-      strcat(finalCommand," ");
-      strcat(finalCommand,absolutePath);
-      strcat(finalCommand,"'");
-      system(finalCommand);
-      if(doesFileExist(finalFileName)){
-        response200WithMessage(response, "Successfully created a file");
+      snprintf(fileNameBuffer, sizeof(fileNameBuffer), "%s%s", tarFileName, ".tar");
+
+      char *tarCommand = "/bin/sh -c 'tar -cf ";
+      snprintf(commandBuffer, sizeof(commandBuffer), "%s%s%s%s%s", tarCommand, fileNameBuffer, " ", absolutePath, "'");
+
+      // system command will create the tar.
+      system(commandBuffer);
+      if(doesFileExist(fileNameBuffer)){
+        respondWithUnixFileContentsWithAutocvtMode(NULL, response, fileNameBuffer, TRUE, 0);
+        deleteUnixFile(fileNameBuffer);
       }
       else{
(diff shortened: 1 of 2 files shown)


c/httpfileservice.c
```diff
@@ -197,6 +197,27 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
+void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
+  if (isDir(absolutePath)) {
+      int folderNameLen = strlen(absolutePath);
+      int slashPos = lastIndexOf(absolutePath, folderNameLen, '/');
+      char *tarFileName = (slashPos == -1) ? "NULL" : absolutePath + slashPos + 1;
+      char finalFileName[strlen(tarFileName) + strlen(".tar.gz")];
+      strcpy(finalFileName, tarFileName);
+      strcat(finalFileName,".tar.gz");
+      char *arguments[] = { "tar", "-zcvf", finalFileName , absolutePath };
+      execvp("/bin/sh", arguments);
```
ifakhrutdinov, line 209: Is this going to compile with Metal C?

c/httpfileservice.c
```diff
@@ -197,6 +197,27 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
   }
 }
 
+void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
+  if (isDir(absolutePath)) {
+      int folderNameLen = strlen(absolutePath);
+      int slashPos = lastIndexOf(absolutePath, folderNameLen, '/');
+      char *tarFileName = (slashPos == -1) ? "NULL" : absolutePath + slashPos + 1;
```
ifakhrutdinov, line 204: Why `"NULL"`?

c/httpfileservice.c
```diff
@@ -197,6 +197,27 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
+      char finalFileName[strlen(tarFileName) + strlen(".tar.gz")];
```
ifakhrutdinov, line 205: You don't have room for the null-terminator, this is a buffer overflow.

ifakhrutdinov, line 205: I'd advise not to use VLA. They became an optional feature in C11.

c/httpfileservice.c
```diff
@@ -197,6 +197,27 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
   }
 }
 
+void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
```
ifakhrutdinov, line 200: The function doesn't really create a file from a directory, it archives that directory.

1000TurquoisePogs:
This pull request is actually somewhat broken, because Ayesh has newer code that he showed me, but it is not appearing in this PR.
Anyway, why do we need such code: we have an existing REST API which serves files as raw content, for the purpose of efficient large file transfer. When you think of FTP, you can download an entire folder recursively. Unfortunately, you can't do exactly the same in a browser, because the API of browsers dont let you create folders and go inside of them. So the only way to respect the hierarchy of folders is to create a tar archive, and allow the browser to download that. I don't see a better way to invoke tar than using an `exec` or `system` call, but it seems `exec` is superior to `system` due to some security and environment variable concerns.

c/httpfileservice.c
```diff
@@ -200,13 +200,21 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
 void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
   if (isDir(absolutePath)) {
       int folderNameLen = strlen(absolutePath);
+      if(folderNameLen == 0 || absolutePath == NULL) {
+         respondWithJsonError(response, "Failed to idenity the folder pointed", 400, "Bad Request");
```
ifakhrutdinov, line 204: Don't you need to return from here?

c/httpfileservice.c
```diff
@@ -197,6 +197,35 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
 
+void createFileFromUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
+  if (isDir(absolutePath)) {
+      int folderNameLen = strlen(absolutePath);
+      if(folderNameLen == 0 || absolutePath == NULL) {
+         respondWithJsonError(response, "Failed to idenity the folder pointed", 400, "Bad Request");
+      }
+      int slashPos = lastIndexOf(absolutePath, folderNameLen, '/');
+      char *tarFileName = (slashPos == -1) ? absolutePath : absolutePath + slashPos + 1;
+      char finalFileName[strlen(tarFileName) + strlen(".tar")];
```
ifakhrutdinov, line 208: This buffer is not big enough. You'll get a buffer overflow.

ifakhrutdinov, line 208: I suggest you avoid using VLA. There is a file name limit, you can use that.

c/httpfileservice.c
```diff
@@ -197,6 +197,35 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
+      strcpy(finalFileName, tarFileName);
+      strcat(finalFileName,".tar");
+      char finalCommand[strlen("/bin/sh -c 'tar -cf ") + strlen(finalFileName) + strlen(absolutePath) + 2];
```
ifakhrutdinov, line 211: Same issue here. You don't have room for the null-terminator.

c/httpfileservice.c
```diff
@@ -197,6 +197,35 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
+      strcpy(finalCommand, "/bin/sh -c 'tar -cf ");
```
ifakhrutdinov, line 212: You can define a single variable with the command and in case the command changes you don't have to update several places.
```c
const char *command = "/bin/sh -c 'tar -cf ";
```

c/httpfileservice.c
```diff
@@ -197,6 +197,35 @@ void deleteUnixDirectoryAndRespond(HttpResponse *response, char *absolutePath) {
...
+      strcat(finalCommand,finalFileName);
+      strcat(finalCommand," ");
+      strcat(finalCommand,absolutePath);
+      strcat(finalCommand,"'");
+      system(finalCommand);
+      if(doesFileExist(finalFileName)){
```
ifakhrutdinov, line 218: Is it possible that:
* File already exists when you run the command
* The command fails (e.g. no write permissions to overwrite the file)
* You return "Success"

The caller will assume that the file has been successfully created whereas the file will still contain the old data.

ifakhrutdinov:
@1000TurquoisePogs what I meant is, can we actually make it part of zss and not zowe-common-c since it's very specific?

c/httpfileservice.c
Akayeshmantha: @ifakhrutdinov  so after serving the file we will delete it.

Akayeshmantha: so it will not exists anyways

c/httpfileservice.c
Akayeshmantha: @ifakhrutdinov  am not familiar with C what is VLA  ?

Akayeshmantha: so  @ifakhrutdinov  what would be the suggestion here ?

1000TurquoisePogs:
I don't think it's that specific... the only place this wont work is windows, and it's the only solution to downloading folders I can think of. As long as the serve file code exists in httpserver.c, this seems like the place to put it.
Maybe file serving code should be moved, but where? If we remove it from zowe-common-c, then there wouldn't be a file server library anymore.

c/httpfileservice.c
1000TurquoisePogs: A simple thing to do is use `sprintf` of the values you want into a buffer of length https://github.com/zowe/zowe-common-c/blob/master/h/unixfile.h#L472

c/httpfileservice.c
Akayeshmantha: hmm it is a void function ?

c/httpfileservice.c
Akayeshmantha: sure

c/httpfileservice.c
Akayeshmantha: used same suggetion

c/httpfileservice.c
Akayeshmantha: sure

c/httpfileservice.c
ifakhrutdinov: I understand that. If you continue execution after you finished the response and then try finishing it again, bad things will happen.

ifakhrutdinov: You can still return from a void function. You can't return anything, but you can exit the function.

Akayeshmantha: sure @ifakhrutdinov

Akayeshmantha:
@ifakhrutdinov  pushed the changes.

c/httpfileservice.c
ifakhrutdinov: I see it's marked resolved. Can you comment on that?

Akayeshmantha: @ifakhrutdinov  so the name of the function should be changed right ?

Akayeshmantha (commented):
moving to zss.

PR #160, opened 2020-08-05
Enable force encoding with provided encoding types.

Akayeshmantha:
This PR will enable force encoding for files.

For a text file to be readable on the client computer, it must be ASCII or UTF8. But a text file on the mainframe might be EBCDIC instead. So it must be converted, and the current code converts AUTOMATICALLY, and sometimes it is wrong.
When it's wrong, we want to tell it to do something we ask for, rather than automatic conversion.
So, we need a parameter to /unixfile API to allow us to specify the source and target encoding. Source=what is on the mainframe, target=what should the user download.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -4053,11 +4053,23 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
 #error Unknown OS
 #endif
         ;
-      if (ccsid == 0) {
-        streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
-      } else {
-        streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, ccsid, webCodePage, asB64);
-      }
+    char *forceEnabled = getQueryParam(response->request, "force");
+    if (ccsid == 0 && strcmp(forceEnabled, "enable")) {
+        char *sourceEncoding = getQueryParam(response->request, "source");
+        char *targetEncoding = getQueryParam(response->request, "target");
+        if(sourceEncoding != NULL && targetEncoding != NULL) {
+           streamTextForFile2(response->socket, in, ENCODING_CHUNKED, sourceEncoding, targetEncoding, asB64);
+        }
+        else {
+           streamTextForFile2(response->socket, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
+        }
+    }
+    else if(ccsid == 0 && strcmp(forceEnabled, "disable")) {
+      streamTextForFile2(response->socket, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
+    }
+    else {
+      streamTextForFile2(response->socket, in, ENCODING_CHUNKED, ccsid, webCodePage, asB64);
+    }
 
 #ifdef USE_CONTINUE_RESPONSE_HACK
       /* See comments above */
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -4064,7 +4064,7 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
            streamTextForFile2(response->socket, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
         }
     }
-    else if(ccsid == 0 && strcmp(forceEnabled, "disable")) {
+    else if(ccsid == 0)) {
       streamTextForFile2(response->socket, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
     }
     else {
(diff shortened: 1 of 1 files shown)


Akayeshmantha:
@ifakhrutdinov pushed the changes. Added a description as well.

c/httpserver.c
```diff
@@ -4053,11 +4053,23 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+        char *sourceEncoding = getQueryParam(response->request, "source");
+        char *targetEncoding = getQueryParam(response->request, "target");
+        if(sourceEncoding != NULL && targetEncoding != NULL) {
+           streamTextForFile2(response->socket, in, ENCODING_CHUNKED, sourceEncoding, targetEncoding, asB64);
+        }
+        else {
+           streamTextForFile2(response->socket, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
+        }
+    }
+    else if(ccsid == 0 && strcmp(forceEnabled, "disable")) {
```
1000TurquoisePogs, line 4067: this should not explicitly check for disable, as it should be considered a default in case people do not specify it.

c/httpserver.c
```diff
@@ -4053,11 +4053,23 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
 #error Unknown OS
 #endif
         ;
-      if (ccsid == 0) {
-        streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
-      } else {
-        streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, ccsid, webCodePage, asB64);
-      }
+    char *forceEnabled = getQueryParam(response->request, "force");
```
ifakhrutdinov, line 4056: `forceEnabled` can be NULL.

c/httpserver.c
```diff
@@ -4053,11 +4053,23 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+    if (ccsid == 0 && strcmp(forceEnabled, "enable")) {
+        char *sourceEncoding = getQueryParam(response->request, "source");
+        char *targetEncoding = getQueryParam(response->request, "target");
+        if(sourceEncoding != NULL && targetEncoding != NULL) {
+           streamTextForFile2(response->socket, in, ENCODING_CHUNKED, sourceEncoding, targetEncoding, asB64);
```
ifakhrutdinov, line 4061: Have you built and tested this code? `streamTextForFile2` has different arguments:
```c
static int streamTextForFile2(HttpResponse *response, Socket *socket, UnixFile *in, int encoding,
                      int sourceCCSID, int targetCCSID, bool asB64);
```

c/httpserver.c
Akayeshmantha: yes it can be null

c/httpserver.c
Akayeshmantha: pushed changes.

Akayeshmantha:
@ifakhrutdinov  it's tested

c/httpserver.c
```diff
@@ -4053,11 +4053,31 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+           int sscanfSource = sscanf(sourceEncoding, "%d", &sEncoding);
+           int sscanfTarget = sscanf(targetEncoding, "%d", &tEncoding);
+           if (sscanfSource != 1 || sscanfTarget != 1) {
+             respondWithError(response, HTTP_STATUS_BAD_REQUEST, "source/target encoding value parsing error");
+             return;
+           }
+           streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, sEncoding, tEncoding, asB64);
+        }
+        else {
+           streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
```
ifakhrutdinov, line 4072: Is this expected behavior to continue with this call implicitly when either of the encodings hasn't been specified? Would it be better to return `HTTP_STATUS_BAD_REQUEST` in the following case
```c
       if(sourceEncoding != NULL && targetEncoding != NULL) {
           int sscanfSource = sscanf(sourceEncoding, "%d", &sEncoding);
           int sscanfTarget = sscanf(targetEncoding, "%d", &tEncoding);
           if (sscanfSource != 1 || sscanfTarget != 1) {
             respondWithError(response, HTTP_STATUS_BAD_REQUEST, "source/target encoding value parsing error");
             return;
           }
           streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, sEncoding, tEncoding, asB64);
        }
        else {
          respondWithError(response, HTTP_STATUS_BAD_REQUEST, "source and target encodings are required");
          return;
        }
```

Akayeshmantha: Hm I hope so becuase the request doesn't content required params ? 
@ifakhrutdinov

ifakhrutdinov: I think if there was an intent to force the encoding but not enough parameters have been provided, it'd be better to return a bad request error. Otherwise we'll do something that wasn't intended.

Akayeshmantha: sure I will do it.

daveyc: You should consider `strtol()` instead of `sscanf()`. It has better error checking and is the canonical way of converting a string to an integer.

c/httpserver.c
```diff
@@ -4053,11 +4053,31 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+           int sscanfSource = sscanf(sourceEncoding, "%d", &sEncoding);
+           int sscanfTarget = sscanf(targetEncoding, "%d", &tEncoding);
+           if (sscanfSource != 1 || sscanfTarget != 1) {
+             respondWithError(response, HTTP_STATUS_BAD_REQUEST, "source/target encoding value parsing error.");
+             return;
+           }
+           streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, sEncoding, tEncoding, asB64);
+        }
+        else {
+          respondWithError(response, HTTP_STATUS_BAD_REQUEST, "force encoding enabled make sure to pass all the requried params");
```
ifakhrutdinov, line 4072: You must return after calling `respondWithError`.

ifakhrutdinov (changes requested):

ifakhrutdinov (approved):

PR #162, opened 2020-08-26
Optional authentication for services

lchudinov:
This PR allows the service to decide how to handle HTTP requests if the user is not authenticated. Example:

```c
void installServerStatusService(HttpServer *server) {
  HttpService *httpService = makeGeneratedService("Simple service", "/simple/service/**");
  httpService->authType = SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN;
  httpService->authFlags |= SERVICE_AUTH_FLAG_OPTIONAL;
  httpService->serviceFunction = serve;
  registerHttpService(server, httpService);
}

int serve(HttpService *service, HttpResponse *response) {
  int isAuthenticated = response->request->authenticated;
  // check if the user is authenticated
  //...
}
```

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3258,8 +3258,12 @@ static int handleHttpService(HttpServer *server,
   printf("service=%s authenticated=%d\n",service->name,request->authenticated);
 #endif
   if (request->authenticated == FALSE){
-    /* could make this parameterizable */
-    respondWithAuthError(response, &authResponse);
+    if (service->authFlags & SERVICE_AUTH_FLAG_OPTIONAL == SERVICE_AUTH_FLAG_OPTIONAL) {
+      // Allow the service to decide when to respond with HTTP 401
+      serveRequest(service, response, request);
+    } else {
+      respondWithAuthError(response, &authResponse);
+    }
     // Response is finished on return
   } else {
 
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3258,7 +3258,7 @@ static int handleHttpService(HttpServer *server,
   printf("service=%s authenticated=%d\n",service->name,request->authenticated);
 #endif
   if (request->authenticated == FALSE){
-    if (service->authFlags & SERVICE_AUTH_FLAG_OPTIONAL == SERVICE_AUTH_FLAG_OPTIONAL) {
+    if (service->authFlags & SERVICE_AUTH_FLAG_OPTIONAL) {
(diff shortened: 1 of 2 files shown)


ifakhrutdinov (commented):
The changes look fine to me.

The purpose of the changes is to allow a new service to serve some data without authentication and, if requested, serve more data if the user is authenticated. One problem I see is that we still call `safAuthenticate` regardless of whether the privileged data has been requested or not. So the code may potentially generate SAF warning messages/events even if the service is going to succeed.

Perhaps a better solution would be to expose a call back function which could be used to check whether the user is authenticated or not. The service could call it only when the requested parameters require authentication. The auth type of the service could be `SERVICE_AUTH_CUSTOM`.

@rocketjared, @1000TurquoisePogs what do you think?

1000TurquoisePogs:
I believe the difference between really authenticated and not authenticated at all, is if the service contains "username" when called, such as here https://github.com/zowe/zss/blob/4681e82825092f142b5b2ac59132469a210788ac/c/unixFileService.c#L606. 
So, even optionally-authed services can determine whether or not auth happened.
This is how it works in the app-server as well, so at least the behavior is unified.

1000TurquoisePogs:
I'd like to merge this as-is and then begin the refactoring effort shortly, so that the app-server and desktop code that wants to utilize this can get worked on with a functional zss independent of refactoring.

Talked with Irek and here is what we learned:

we're basically making it valid to produce SAF warning messages about failed authentications if we don't distinguish users that couldn't authenticate and users that didn't even attempt


the app-server security is a little different to ZSS but similar and it's been working well for 2 years:
1. request arrives
2. app-server delegates auth to security plugin (such as zss)
3. if response is false (auth failed), then it will return 403 if auth is mandatory, or continue rest api if auth is optionalthis is sort of what zss does, except 2 is SAF on z/os, and [something else] on another platform

> Is that specifically a security service or any service can make such decision?

1 & 2 & 3 dont happen within a service, in either the app-server or zss today

but for both there is a step 0 where the service declares what type of security it would like, on, off (and now optional)
step 3 takes the type of security into consideration when deciding whether to continue or return 403

the only time the service is involved is when security=optional, the service must decide what data gets returned, since zss doesnt know the data structure.

So, to help the zss core determine whether to process or reject the request, perhaps this is best:

`authentcated == False && username != NULL`  should return an error to the caller
but `authenticated == False && username == NULL` should allow returning partial data
and `authenticated == True` would allow all the data

PR #169, opened 2020-09-07
Refactor HTTP Server Authentication code

lchudinov:
This is a follow up for https://github.com/zowe/zowe-common-c/pull/162. 

1. This PR makes `request->authenticated` field always actual. 
2. In the case when `SERVICE_AUTH_FLAG_OPTIONAL` is set. The intent is to distinguish 
    1. the case when SAF authentication was attempted because some auth data was provided on the request(in HTTP headers, 
cookies). In this case:
        * if SAF authentication was successful the HTTP server calls the service handler 
        * otherwise, the HTTP server responds with HTTP 401 immediately 
    2. the case when no auth data was provided on the request. In this case 
        * the HTTP server calls the service handle

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -2486,7 +2486,7 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
 
     authResponse->type = AUTH_TYPE_RACF;
     authResponse->responseDetails.safStatus = status.safStatus;
-    authResponse->authDataFound = (bool)authDataFound;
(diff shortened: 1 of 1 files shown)


ifakhrutdinov (dismissed):
Technically, the changes look good to me. I would still other people to take a look, especially since this is related to security. @rocketjared, @1000TurquoisePogs, what do you think?

1000TurquoisePogs (approved):
I looked at it and it seems good to me as well.

PR #170, opened 2020-09-14
Include owner and group into Unixfile metadata response and directory listing

lchudinov:
This PR includes `owner` and `group` fields into the Unixfile metadata and directory listing(JSON).

### DIFF
--- a/c/zosaccounts.c
+++ b/c/zosaccounts.c
@@ -38,20 +38,28 @@ static int accountTrace = FALSE;
 
 #ifdef _LP64
 # pragma linkage(BPX4GGN,OS)
+# pragma linkage(BPX4GGI,OS)
 # pragma linkage(BPX4GPN,OS)
+# pragma linkage(BPX4GPU,OS)
 # pragma linkage(BPX4PWD,OS)
 
 # define BPXGGN BPX4GGN
+# define BPXGGI BPX4GGI
 # define BPXGPN BPX4GPN
+# define BPXGPU BPX4GPU
 # define BPXPWD BPX4PWD
 
 #else
 # pragma linkage(BPX1GGN,OS)
+# pragma linkage(BPX1GGI,OS)
 # pragma linkage(BPX1GPN,OS)
+# pragma linkage(BPX1GPU,OS)
 # pragma linkage(BPX1PWD,OS)
 
 # define BPXGGN BPX1GGN
+# define BPXGGI BPX1GGI
 # define BPXGPN BPX1GPN
+# define BPXGPU BPX1GPU
 # define BPXPWD BPX1PWD
 #endif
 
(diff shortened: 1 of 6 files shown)


c/zosaccounts.c
```diff
@@ -211,6 +308,11 @@ int userInfoGetUserId (UserInfo *info) {
   return userId;
 }
 
+/* Copy user name from User Info structure into userNameBuffer */
+void userInfoGetUserName (UserInfo *info, char *userNameBuffer) {
```
DivergentEuropeans, line 312: Necessary to have its own method?

c/zosaccounts.c
```diff
@@ -220,6 +322,11 @@ int groupInfoGetGroupId (GroupInfo *info) {
   return groupId;
 }
 
+/* Copy group name from Group Info structure into groupNameBuffer*/
+void groupInfoGetGroupName (GroupInfo *info, char *groupNameBuffer) {
```
DivergentEuropeans, line 326: Necessary to have its own method?

c/zosaccounts.c
```diff
@@ -211,6 +308,11 @@ int userInfoGetUserId (UserInfo *info) {
   return userId;
 }
 
+/* Copy user name from User Info structure into userNameBuffer */
+void userInfoGetUserName (UserInfo *info, char *userNameBuffer) {
+  memcpy(userNameBuffer, (const char*)&info->GIDN_U_NAME, info->GIDN_U_LEN);
```
DivergentEuropeans, line 313: Necessary to have its own method?

c/zosaccounts.c
```diff
@@ -266,6 +384,17 @@ int groupIdGet (char *string, int *returnCode, int *reasonCode) {
   return groupId;
 }
 
+/* Obtain group name by GID, groupNameBuffer must be at least 8 chars long */
+int groupGetName(int gid, char *groupNameBuffer, int *returnCode, int *reasonCode) {
+  int status = 0;
+  GroupInfo groupInfo = {0};
+  status = getGroupInfo(gid, &groupInfo, returnCode, reasonCode);
+  if (status == 0) {
```
DivergentEuropeans, line 392: Should we check if groupNameBuffer is at least 8 chars long before getting here?

DivergentEuropeans (approved):
This works for me!

c/zosaccounts.c
lchudinov: It's just for the sake of consistency with the existing code that provides convenient access methods e.g. `groupInfo GetGroupId`

c/zosaccounts.c
lchudinov: It's just for the sake of consistency with the existing code that provides convenient access methods e.g. `userInfoGetUserId`

c/zosaccounts.c
lchudinov: We are unable to check if `groupNameBuffer` because the caller is responsible for allocating the buffer that is large enough.

1000TurquoisePogs (approved):
Looks correct to me. I noticed the BPXY structs for user and group have username and groupname as a char and not a pointer, which I think could be improved, but this PR does not disturb existing code in that regard so it's good to merge.

daveyc:
PMFJI, but the C structure mapping for `BPXYGIDS` and `BPXYGIDN` are not correct. The assembler DSECT `GIDS_G_NAME` is defined as `DS 0C` which has a zero duplication factor. This signifies that the the field has a variable length. To parse the results for the system calls will require pointer arithmetic using `GIDS_G_LEN`. This seems particularily brain-damaged as the maximum length of the field is only 8 bytes but every byte is sacred in HLASM land.

PR #176, opened 2020-10-09
Allow to change tag for all files in the directory excluding subdirectories

lchudinov:
This PR allows to change tag for all files in the directory excluding subdirectories.
For example `POST /unixfile/chtag/u/user/tmp?codeset=1047&type=text&recursive=false` should change tags only for files in `u/user/tmp` without subdirectories


DivergentEuropeans (approved):
This works

PR #180, opened 2020-10-28
Secure HTTP Client

lchudinov:
Depends on:

- https://github.com/zowe/zowe-common-c/pull/178
- https://github.com/zowe/zowe-common-c/pull/179


1000TurquoisePogs (approved):
This just adds the same tls code that was previously merged, to the http client which was previously merged, so I don't see a problem with it.

PR #183, opened 2020-11-17
Make zss storage API

lchudinov:
## ZSS Storage API

Preparation work before connecting ZSS and Cashing Service in APIML (https://docs.google.com/document/d/1Z8T1mq7MRYzxREL2pf1zTlj0tP7fqaR2ba31TZHX0Qw/edit?usp=sharing)
The PR 
- defines the shape of the API
- adds in-memory implementation for the API

Related to https://github.com/zowe/api-layer/issues/890.

Test program for Memory Storage is embedded into the source code.

### DIFF
--- a/c/storage.c
+++ b/c/storage.c
@@ -1,0 +1,139 @@
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+  
+  SPDX-License-Identifier: EPL-2.0
+  
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifdef METTLE
+#include <metal/metal.h>
+#include <metal/stddef.h>
+#include <metal/stdio.h>
+#include <metal/stdlib.h>
+#include <metal/stdbool.h>
+#include <metal/string.h>
+#include <metal/stdarg.h>  
+#include <metal/limits.h>  
+#include "metalio.h"
+#include "qsam.h"
+#else
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <stdarg.h>  
+#include <limits.h>
+#endif
+
+#include "zowetypes.h"
+#include "alloc.h"
+#include "utils.h"
+#include "storage.h"
+
+static const char *TRUE_STRING = "true";
+static const char *FALSE_STRING = "false";
+
+static int parseInteger(const char *str, int *statusOut) {
+  char *end;
+  long longValue = strtol(str, &end, 10);
+  if(*end != '\0') {
+    *statusOut = STORAGE_VALUE_NOT_INTEGER;
+    return 0;
+  }
+  if (longValue < INT_MIN || longValue > INT_MAX) {
+    *statusOut = STORAGE_INTEGER_TOO_LONG;
+    return 0;
+  }
+  *statusOut = STORAGE_OK;
+  return (int)longValue;
+}
+
+static bool parseBool(const char *str, int *statusOut) {
+  if (strcmp(str, TRUE_STRING) && strcmp(str, FALSE_STRING)) {
+    *statusOut = STORAGE_VALUE_NOT_BOOLEAN;
+    return false;
+  }
+  *statusOut = STORAGE_OK;
+  return !strcmp(str, TRUE_STRING);
+}
+
+static const char *MESSAGES[] = {
+    [STORAGE_OK] = "OK",
+    [STORAGE_KEY_NOT_FOUND] = "Key not found",
+    [STORAGE_VALUE_NOT_BOOLEAN] = "Value not boolean",
+    [STORAGE_VALUE_NOT_INTEGER] = "Value not integer",
+    [STORAGE_INTEGER_TOO_LONG] = "Integer value out of bounds",
+};
+
+#define MESSAGE_COUNT sizeof(MESSAGES)/sizeof(MESSAGES[0])
+
+const char *storageGetString(Storage *storage, const char *key, int *statusOut) {
+  return storage->get(storage->userData, key, statusOut);
+}
+
+int storageGetInt(Storage *storage, const char *key, int *statusOut) {
+  int status = 0;
+  const char *str = storageGetString(storage, key, &status);
+  if (status) {
+    *statusOut = status;
+    return 0;
+  }
+  return parseInteger(str, statusOut);
+}
+
+bool storageGetBool(Storage *storage, const char *key, int *statusOut) {
+  int status = 0;
+  const char *str = storageGetString(storage, key, &status);
+  if (status) {
+    *statusOut = status;
+    return false;
+  }
+  return parseBool(str, statusOut);
+}
+
+void storageSetString(Storage *storage, const char *key, const char *value, int *statusOut) {
+  storage->set(storage->userData, key, value, statusOut);
+}
+
+#define INT_LEN 11
+
+void storageSetInt(Storage *storage, const char *key, int value, int *statusOut) {
+  char buf[INT_LEN+1] = {0};
+  snprintf (buf, sizeof(buf), "%d", value);
+  storageSetString(storage, key, buf, statusOut);
+}
+
+void storageSetBool(Storage *storage, const char *key, bool value, int *statusOut) {
+  storageSetString(storage, key, value ? TRUE_STRING : FALSE_STRING, statusOut);
+}
+
+void storageRemove(Storage *storage, const char *key, int *statusOut) {
+  storage->remove(storage->userData, key, statusOut);
+}
+
+const char *storageGetStrStatus(Storage* storage, int status) {
+  if (status >= STORAGE_FIRST_CUSTOM_STATUS) {
+    return storage->strStatus(storage->userData, status);
+  }
+  if (status >= MESSAGE_COUNT || status < 0) {
+    return "Unknown status code";
+  }
+  const char *message = MESSAGES[status];
+  if (!message) {
+    return "Unknown status code";
+  }
+  return message;
+}
+
+/*
+  This program and the accompanying materials are
+  made available under the terms of the Eclipse Public License v2.0 which accompanies
+  this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+  
+  SPDX-License-Identifier: EPL-2.0
+  
+  Copyright Contributors to the Zowe Project.
+*/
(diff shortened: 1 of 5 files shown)


ifakhrutdinov:
@lchudinov can you describe what this API is for?

lchudinov:
It's API for caching purpose. My PR defines the shape of the API, and adds in-memory implementation. There will be another PR that utilizes Caching Service in APIML.

c/storage.c
```diff
@@ -0,0 +1,176 @@
...
+  if (value == HUGE_VAL || value == -HUGE_VAL) {
+    *statusOut = STORAGE_DOUBLE_OUT_OF_RANGE;
+    return 0.0;
+  }
+  *statusOut = STORAGE_OK;
+  return value;
+}
+
+static bool parseBool(const char *str, int *statusOut) {
+  if (strcmp(str, TRUE_STRING) && strcmp(str, FALSE_STRING)) {
```
ifakhrutdinov, line 72: You can save one `strcmp` call :)
```c
*statusOut = STORAGE_OK;

if (!strcmp(str, TRUE_STRING)) {
  return true;
}
if (!strcmp(str, FALSE_STRING)) {
  return false;
}

*statusOut = STORAGE_VALUE_NOT_BOOLEAN;

return false;
```

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+  hashtable *table;      // map: key -> value
+  pthread_mutex_t lock;  // lock to protect hashtable
+} MemStorage;
+
+static char *memStorageAlloc(int size) {
+  char *data = safeMalloc31(size + 4, "mem storage");
+  if (!data) {
+    return NULL;
+  }
+  *((int *)data) = size;
```
ifakhrutdinov, line 42: You could make this more explicit by having some sort of header.
```c
struct InMemStorageBlock {
  int size;
  char data[0];
};
```

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+}
+
+#define DEFAULT_BUCKET_COUNT 257
+
+static MemStorage *makeMemStorage(MemoryStorageOptions *options) {
+  MemStorage *storage = (MemStorage*)safeMalloc(sizeof(*storage), "MemStorage");
+  if (!storage) {
+    return NULL;
+  }
+  int bucketCount = (options && options->bucketCount) > 0 ? options->bucketCount : DEFAULT_BUCKET_COUNT;
```
ifakhrutdinov, line 66: Did you intend using the following?
```c
int bucketCount = options && options->bucketCount > 0 ? options->bucketCount : DEFAULT_BUCKET_COUNT;
```
Because `options && options->bucketCount` will be either 0 or 1, never -1.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+    safeFree((char*)storage, sizeof(*storage));
+    return NULL;
+  }
+  storage->table = table;
+  int rc = pthread_mutex_init(&storage->lock, NULL);
+  if (rc) {
+    safeFree((char*)storage, sizeof(*storage));
+    htDestroy(table);
+  }
+  return storage;
```
ifakhrutdinov, line 78: If `pthread_mutex_init` fails you'll return a dangling pointer.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+}
+
+static char *memStorageCopyString(MemStorage *storage, const char *str) {
+  const int len = strlen(str);
+  char *newString = memStorageAlloc(len + 1);
+  if (!newString) {
+    return NULL;
+  }
+  memset(newString, '\0', len + 1);
+  memcpy(newString, str, len);
```
ifakhrutdinov, line 104: `strcpy()` would do here.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+#define STOARGE_LOCK_ERROR   (STORAGE_FIRST_CUSTOM_STATUS + 0)
+#define STOARGE_UNLOCK_ERROR (STORAGE_FIRST_CUSTOM_STATUS + 1)
+#define STORAGE_ALLOC_ERROR  (STORAGE_FIRST_CUSTOM_STATUS + 2)
+
+typedef struct MemStorage_tag {
+  hashtable *table;      // map: key -> value
+  pthread_mutex_t lock;  // lock to protect hashtable
+} MemStorage;
+
+static char *memStorageAlloc(int size) {
```
ifakhrutdinov, line 37: Return `void*` so you don't have to cast, like `malloc()`.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+  char *data = safeMalloc31(size + 4, "mem storage");
+  if (!data) {
+    return NULL;
+  }
+  *((int *)data) = size;
+  return (data + 4);
+}
+
+static void memStorageFree(char *mem) {
```
ifakhrutdinov, line 46: Use `void *` so you don't have to cast.

BTW, if you're not going to use this under metal, why not just use malloc/free. You basically have `memStorageAlloc->safeMalloc->malloc`.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+    if (valueCopy) {
+      memStorageFree((char*)valueCopy);
+    }
+    *statusOut = STORAGE_ALLOC_ERROR;
+    return;
+  }
+  int status = memStorageLock(storage);
+  if (status) {
+    *statusOut = status;
+    return;
```
ifakhrutdinov, line 124: You'll leak `keyCopy` and `valueCopy ` if this fails.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
...
+  storage.c \
+  storage_mem.c \
+  utils.c \
+  timeutls.c 
+
+Run: ./storage
+
+*/
+
+int main(int argc, char *argv[]) {
```
ifakhrutdinov, line 218: Could you do something like
* Put a value to the store
* Check if the value is in the store using `assert`
* Check a non-existent value in the store using `assert` (should not be there)
* etc

That way you'll see right away if something unexpected happens.

h/dataservice.h
```diff
@@ -66,10 +67,13 @@ typedef struct DataService_tag {
   JsonObject *serviceDefinition;
   uint64 loggingIdentifier;
   WebPlugin *plugin;
+  struct Storage_tag *storage;
 } DataService;
 
 WebPlugin *makeWebPlugin(char *baseDir, struct JsonObject_tag *pluginDefintion, InternalAPIMap *internalAPIMap,
                          unsigned int *idMultiplier, int pluginLogLevel);
+WebPlugin *makeWebPlugin2(char *pluginLocation, JsonObject *pluginDefintion, InternalAPIMap *internalAPIMap,
+                         unsigned int *idMultiplier, int pluginLogLevel, struct Storage_tag *storage);
```
ifakhrutdinov, line 76: Can you document the arguments (i.e. doxygen style)? It's unclear why a web plugin needs some storage.

c/storage_mem.c
```diff
@@ -0,0 +1,286 @@
+/*
```
ifakhrutdinov, line 1: Perhaps this file belongs in the `test/` directory?

h/storage.h
```diff
@@ -0,0 +1,47 @@
+#ifndef STORAGE_H
+#define STORAGE_H
+
```
ifakhrutdinov, line 3: I think we should start documenting the APIs. For example, see https://github.com/zowe/zowe-common-c/blob/master/h/cellpool.h. I've started using doxygen doc in new files.

ifakhrutdinov (commented):
@lchudinov please see my comments. There are some minor bugs, corrections.

c/storage_mem.c
lchudinov: I would leave it here until we use a test suite.

c/storage_mem.c
```diff
@@ -110,16 +114,18 @@ static void memStorageSetString(MemStorage *storage, const char *key, const char
   const char *valueCopy = memStorageCopyString(storage, value);
   if (!keyCopy || !valueCopy) {
     if (keyCopy) {
-      memStorageFree((char*)keyCopy);
+      memStorageFree((void*)keyCopy);
```
ifakhrutdinov, line 117: That's the beauty of `void *`, you don't have to cast anymore.

ifakhrutdinov (commented):
@lchudinov, I have a couple of questions
* Do users of this `dataservice.c` have to link with `storage.c`? For example, external plugins. You can check with my bare minimum plugin, https://github.com/ifakhrutdinov/zss-sample-plugin
* Has the team considered performance implications of converting binary data into strings and back? How often will this code be used?

c/storage_mem.c
lchudinov: Still need the cast because of `const char*`. `void*` would help in case of `char *`.

lchudinov:
> * Do users of this `dataservice.c` have to link with `storage.c`? For example, external plugins. You can check with my bare minimum plugin, https://github.com/ifakhrutdinov/zss-sample-plugin
External plugins don't need to link `storage.c` for now. In the future, we must put Storage API to `plugin.API.x`
> * Has the team considered performance implications of converting binary data into strings and back? How often will this code be used?
For now, Caching API supports only strings where we can store anything, e.g. JSON. We haven't consider performance implications yet. We don't have any callers of Storage API so I can't answer to your the second part of your question.

ifakhrutdinov:
> 
> 
> > * Do users of this `dataservice.c` have to link with `storage.c`? For example, external plugins. You can check with my bare minimum plugin, https://github.com/ifakhrutdinov/zss-sample-plugin
> 
> External plugins don't need to link `storage.c` for now. In the future, we must put Storage API to `pluginAPI.x`
> 
> > * Has the team considered performance implications of converting binary data into strings and back? How often will this code be used?
> 
> For now, Caching API supports only strings where we can store anything, e.g. JSON. We haven't consider performance implications yet. We don't have any callers of Storage API so I can't answer to your the second part of your question.

ifakhrutdinov (approved):
lgtm from the technical point of view
@1000TurquoisePogs can you have a look at the API from the architectural point of view?
* Is this how caching called in web, storage API? Should we considered something more idiomatic?
* Does this really belong in zowe-common-c? That is, will anyone else potentially use this except for ZSS?

1000TurquoisePogs:
> * Is this how caching called in web, storage API? Should we considered something more idiomatic?
> * Does this really belong in zowe-common-c? That is, will anyone else potentially use this except for ZSS?
We lacked an interface for storage of service data, and while there will be a specific implementation for zss, the general concept can be reused by others, so it's good to put that interface here.

1000TurquoisePogs (approved):

PR #184, opened 2020-11-27
Dsedit#498

John-A-Davies:
These are the basic changes for the dataset-save feature described in zlux #498

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -46,6 +46,13 @@
 
 #define INDEXED_DSCB 96
 
+#include "semTable.h"
+struct sem_table_type sem_table_entry [];
+#ifndef _XOPEN_SOURCE
+#define _XOPEN_SOURCE 
+#endif
+#include <sys/sem.h>
+
 static char defaultDatasetTypesAllowed[3] = {'A','D','X'};
 static char clusterTypesAllowed[3] = {'C','D','I'}; /* TODO: support 'I' type DSNs */
 static int clusterTypesCount = 3;
@@ -1591,46 +1598,46 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
   memcpy(daMember.name, memberName.value, sizeof(daMember.name));
 
   /* new code to detect ENQ header begins */
-  zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-            "ZSS0001I %s Requested dsn=\'%44.44s\', member=\'%8.8s\'\n", __FUNCTION__,
-            daDsn.name, daMember.name);
+  // zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //           "ZSS0001I %s Requested dsn=\'%44.44s\', member=\'%8.8s\'\n", __FUNCTION__,
+  //           daDsn.name, daMember.name);
 
-  int exclusiveENQ = 0; /* FALSE */
-  /* obtain header name&value */
+  // int exclusiveENQ = 0; /* FALSE */
+  // /* obtain header name&value */
   
-  HttpHeader *enqHeader = getHeader(request, "ENQ");
+  // HttpHeader *enqHeader = getHeader(request, "ENQ");
 
-  if (enqHeader == NULL){
-    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-            "ZSS0002I %s HTTP request has no ENQ header\n", __FUNCTION__);
-    }
-    else {
-      zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-        "ZSS0003I %s HTTP request has ENQ header\n", __FUNCTION__);
-          zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-            "ZSS1621I %s Raw ENQ header value is %4.4s\n", __FUNCTION__, enqHeader->value );
+  // if (enqHeader == NULL){
+  //   zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //           "ZSS0002I %s HTTP request has no ENQ header\n", __FUNCTION__);
+  //   }
+  //   else {
+  //     zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //       "ZSS0003I %s HTTP request has ENQ header\n", __FUNCTION__);
+  //         zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //           "ZSS1621I %s Raw ENQ header value is %4.4s\n", __FUNCTION__, enqHeader->value );
           
-          /* header values are in ASCII, so convert the local EBCDIC string before comparing them. */
-          char enqValue [5] = "true";
-          if(e2a(enqValue, 4) == -1){
-            zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-              "ZSS1626I %s Failed to convert true to ASCII\n", __FUNCTION__);
-          }else{            
-            if (!strncmp(enqHeader->value, enqValue, strlen(enqHeader->value)))
-              {
-                zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-                  "ZSS0005I %s HTTP request is asking for ENQ=true\n", __FUNCTION__);
-              /* request is asking for ENQ */
-                exclusiveENQ = 1; /* TRUE */
-              }
-              else {
-                zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
-                  "ZSS0006I %s HTTP request is NOT asking for ENQ=true\n", __FUNCTION__);
-                exclusiveENQ = 0;
-              }
-            }
+  //         /* header values are in ASCII, so convert the local EBCDIC string before comparing them. */
+  //         char enqValue [5] = "true";
+  //         if(e2a(enqValue, 4) == -1){
+  //           zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //             "ZSS1626I %s Failed to convert true to ASCII\n", __FUNCTION__);
+  //         }else{            
+  //           if (!strncmp(enqHeader->value, enqValue, strlen(enqHeader->value)))
+  //             {
+  //               zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //                 "ZSS0005I %s HTTP request is asking for ENQ=true\n", __FUNCTION__);
+  //             /* request is asking for ENQ */
+  //               exclusiveENQ = 1; /* TRUE */
+  //             }
+  //             else {
+  //               zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  //                 "ZSS0006I %s HTTP request is NOT asking for ENQ=true\n", __FUNCTION__);
+  //               exclusiveENQ = 0;
+  //             }
+  //           }
 
-    } /* end ELSE request has header */
+  //   } /* end ELSE request has header */
   
   /* new code ends   */
 
@@ -1657,9 +1664,6 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
     return;
   }
 
-/* new code to enqueue dataset starts */
-
-/* new code to enqueue dataset ends */
   zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
           "debug: reading dsn=\'%44.44s\', member=\'%8.8s\', dd=\'%8.8s\'\n",
           daDsn.name, daMember.name, daDDname.name);
@@ -1686,7 +1690,13 @@ zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
 }
 
 /* new function for ENQ */
-void respondWithEnqueue(HttpResponse* response, char* absolutePath, int jsonMode) {
+void respondWithEnqueue(HttpResponse* response, char* absolutePath, int jsonMode, char *sem_table_pointer) {
+
+  zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+      "ZSS1689I Begin %s\n", __FUNCTION__); 
+      pid_t pid = getpid();
+      zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_INFO,
+        "ZSS1692I %s PID is %d\n", __FUNCTION__, pid);
 
   HttpRequest *request = response->request;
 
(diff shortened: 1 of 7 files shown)


c/datasetjson.c
```diff
@@ -1602,34 +1655,520 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
   );
 
   if (daRC != RC_DYNALLOC_OK) {
-    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
-    		    "error: ds alloc dsn=\'%44.44s\', member=\'%8.8s\', dd=\'%8.8s\',"
+    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_SEVERE,
```
ifakhrutdinov, line 1658: Do not turn these `ZOWE_LOG_DEBUG` into `ZOWE_LOG_SEVERE` and `ZOWE_LOG_INFO` because they will be printed by default. Since this is library code, we can't print anything here by default.

c/datasetjson.c
```diff
@@ -1602,34 +1655,520 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+  zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
           "debug: reading dsn=\'%44.44s\', member=\'%8.8s\', dd=\'%8.8s\'\n",
           daDsn.name, daMember.name, daDDname.name);
 
   DDName ddName;
   memcpy(&ddName.value, &daDDname.name, sizeof(ddName.value));
   respondWithDatasetInternal(response, absolutePath, &dsn, &ddName, jsonMode);
 
+zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,
+  "ZSS1778I %s wait around for manual check of ENQ ...\n", __FUNCTION__);
```
ifakhrutdinov, line 1676: Do not put message IDs, they are product specific. Again, this is common/library code.

c/datasetjson.c
```diff
@@ -1602,34 +1655,520 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+  //             }
+  //           }
+
+  //   } /* end ELSE request has header */
+  
+  /* new code ends   */
+
+  // DynallocDDName daDDname = {.name = "????????"};
+
+  // int daRC = RC_DYNALLOC_OK, daSysRC = 0, daSysRSN = 0;
```
ifakhrutdinov, line 1763: I believe you're going to remove all this commented out code eventually. Is that correct?

h/stcbase.h
```diff
@@ -78,6 +78,7 @@ typedef struct STCModule_tag {
   int  (*udpHandler)(struct STCBase_tag *stcBase, struct STCModule_tag *module, Socket *socket);
   int  (*workElementHandler)(struct STCBase_tag *stcBase, struct STCModule_tag *stcModule, WorkElementPrefix *prefix);
   int  (*backgroundHandler)(struct STCBase_tag *stcBase, struct STCModule_tag *stcModule, int selectReturnCode);
+  char *sem_table_pointer;
```
ifakhrutdinov, line 81: You're putting something very specific into a very generic header. This cannot be here.

c/httpserver.c
```diff
@@ -5876,6 +5901,17 @@ void registerHttpServerModuleWithBase(HttpServer *server, STCBase *base)
 int mainHttpLoop(HttpServer *server){
   STCBase *base = server->base;
   /* server pointer will be copied/accessible from module->data */
+
+
+  zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_INFO,
+    "ZSS5742I %s Create semaphore table for datasets\n", __FUNCTION__);
+
+  int i;
+  for(i=0; i<N_SEM_TABLE_ENTRIES; i++)
```
ifakhrutdinov, line 5910: Please use the style you see in the file you're changing. 
* Do you need `i` outside the for loop? If not, please don't declare it outside the for loop.
* Please throw in some blanks around the operators

ifakhrutdinov (changes requested):
I suggest you convert this into a draft. I haven't looked at this line by line, I've commented on the larger issues I saw. I will do a more thorough review when the pull-request is ready.

ifakhrutdinov:
Please provide a detailed description of the changes. See examples in https://github.com/zowe/zowe-common-c/pull/121, https://github.com/zowe/zowe-common-c/pull/133.

c/httpserver.c
```diff
@@ -62,6 +62,7 @@
 #include "charsets.h"
 #include "impersonation.h"
 #include "httpserver.h"
+#include "zssLogging.h"
```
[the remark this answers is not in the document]
John-A-Davies: I will remove that #include

c/datasetjson.c
John-A-Davies: That was the quickest way for me to see the message, rather than enabling ZOWE_LOG_DEBUG.
Yes, these will need tidying before MERGE.

c/datasetjson.c
John-A-Davies: Yes, accepted.  I just used line numbers.  I never intended them to be merged in that form.

c/datasetjson.c
John-A-Davies: Yes, that's part of code cleanup before MERGE.

h/stcbase.h
John-A-Davies: Yes, that was a tryout, which I have removed everywhere, but I missed that one.  I will remove it now.

c/httpserver.c
John-A-Davies: Yes, part of cleanup

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+        zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,"%c",*(rname_parm.value+i));
+    }
+    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,">\n");
+    
+/* test lock */
+    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG, 
+      "ZSS1826I %s Try exclusive Lock\n", __FUNCTION__);
+    
+/*    printf("TryExclusiveLock\n");  waits forever? CONTENTIONACT=FAIL */
+    lockRC = isgenqTryExclusiveLock(
```
timgerstel, line 1728: Does this function wait forever? I recall attempting a similar implementation and noting such behavior from isgenq functions. This could potentially cause the ZSS server to hang

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+          /* wait for semaphore to be posted */
+          
+          semaphoreRetcode = semop(semaphoreID, semaphoreOps, 1); /* 1 = one entry */
+                /* we are now waiting for our semaphore to be posted ... */
+
+          
+          if (semaphoreRetcode != 0 )
+          {
+              zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+                "ZSS1938I %s Failed to wait on semaphore, retcode = %d\n", __FUNCTION__, semaphoreRetcode);
```
timgerstel, line 1915: Should this entire logic be moved before the "respondWithMessage? If the semaphore operation is not posted then could this cause unexpected conditions?

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+        if (sem_table_entry[i].sem_ID != 0                                  /* semaphore exists */      
+            && memcmp(sem_table_entry[i].dsn, dsn_member.dsn, 44) == 0          /* DSN matches */
+            && memcmp(sem_table_entry[i].mem, dsn_member.membername, 8) == 0   /* member matches */
+            ) /* already present */
+        {
+          zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+            "ZSS1945I %s dsn already in table in row %d with semaphoreID %D\n", __FUNCTION__, i, sem_table_entry[i].sem_ID );  
+          break;
+        }
+        if (sem_table_entry[i].sem_ID == 0) /* found empty slot in table */
```
ifakhrutdinov, line 1812: How do you serialize this semaphore table itself? What if are 10 requests checking this table?

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+          arg.val = 1;
+          int     semnum = 0;    /* array index zero */
+          int semaphoreRetcode;
+
+          zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+                  "ZSS1916I %s Set semaphore %d, PID %d\n", __FUNCTION__, semaphoreID, pid);
+          semaphoreRetcode = semctl(semaphoreID, semnum, SETVAL, arg); /* set the value of our semaphore */
+          if (semaphoreRetcode == -1 )
+            {
+              printf("Failed to set semaphore\n");
```
ifakhrutdinov, line 1875: Naked printf

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+          /* define semaphore operation to be performed */
+          struct sembuf semaphoreBuffer[1]; /* just one semaphore in the array */
+          struct sembuf *semaphoreOps = &semaphoreBuffer[0];
+          semaphoreBuffer[0].sem_num = 0; /* index of first and only semaphore */
+          semaphoreBuffer[0].sem_op  = 0; /* 0 = wait */
+          semaphoreBuffer[0].sem_flg = 0; /* 0 = sychronous + don't undo */
+
+          /* wait for semaphore to be posted */
+          
+          semaphoreRetcode = semop(semaphoreID, semaphoreOps, 1); /* 1 = one entry */
```
ifakhrutdinov, line 1908: What happens if no one posts? Will this thread wait forever? I can already see this denial of service attack:
* Caller A does an enqueue on dataset A and never do dequeue 
* Caller B does >1000 enqueues on datasets A and all the threads are stuck - zss is unresponsive until caller A does a dequeue

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+        zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+            "ZSS1997I %s semaphore %d is already present \n", __FUNCTION__, semaphoreID);
+        respondWithError(response, HTTP_STATUS_RESOURCE_CONFLICT , "dataset is already enqueued"); 
+      } /* end of create semaphore */
+  /* end of semaphore section */
+
+} /* end of respondWithEnqueue */
+
+
+/* new function for DEQ */
```
ifakhrutdinov, line 1946: Rather than stating the obvious, it's better to provide a real description of the purpose and high level logic.

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+void respondWithDequeue(HttpResponse* response, char* absolutePath, int jsonMode, char *sem_table_pointer) {
```
ifakhrutdinov, line 1947: What's `sem_table_pointer`? Is it some string since it's `char *`? The purpose is totally unclear without looking at the implementation.

c/httpserver.c
```diff
@@ -5882,6 +5874,11 @@ void registerHttpServerModuleWithBase(HttpServer *server, STCBase *base)
 int mainHttpLoop(HttpServer *server){
   STCBase *base = server->base;
   /* server pointer will be copied/accessible from module->data */
+
+  /* Create semaphore table for datasets */
+  for(int i=0; i < N_SEM_TABLE_ENTRIES; i++)
```
ifakhrutdinov, line 5879: Why is this in the http server? This is service specific and an http server should not care about internal implementation of its services.

h/semTable.h
```diff
@@ -0,0 +1,7 @@
+  /* declare the table that maps datasets to semaphores */
+  struct sem_table_type {
```
ifakhrutdinov, line 2: I don't think this header should exist in the first place. But if we had to have it, it must have a some sort of a include guard.

c/datasetjson.c
ifakhrutdinov, line 1728: Who releases this lock?

ifakhrutdinov (changes requested):
There are issues, please see my comments.

I don't understand the architecture and the purpose of all the locking. Again, as I've already said:
> Please provide a detailed description of the changes. See examples in https://github.com/zowe/zowe-common-c/pull/121, https://github.com/zowe/zowe-common-c/pull/133.

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+    zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+      "ZSS1793I %s GetExclusiveLock\n", __FUNCTION__);
+
+    lockRC = isgenqGetExclusiveLock( 
+        &MajorQNAME, 
+        &rname_parm, 
+        ISGENQ_SCOPE_SYSTEMS, 
+        &lockToken, 
+        &lockRSN);   
+    /* This function WAITs in case of contention.  This should not be a problem because it's preceeded by isgenqTryExclusiveLock */
```
ifakhrutdinov, line 1777: No, this is a problem because someone can acquire that lock in between the calls.

c/datasetjson.c
John-A-Davies: Great spot, but when I looked later, I found CONTENTIONACT=FAIL on that call, so it doesn't wait.  I forgot to amend the comment, which I've now done.

c/datasetjson.c
John-A-Davies: I think the code is right as it stands.  The test `if (semaphoreRetcode != 0 )` asks "did the semop call fail?" so it would fail to start waiting.  Otherwise it waits until it's posted.

c/datasetjson.c
```diff
@@ -46,6 +46,13 @@
 
 #define INDEXED_DSCB 96
 
+#include "semTable.h"
+struct sem_table_type sem_table_entry [];
```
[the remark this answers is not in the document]
John-A-Davies: I don't see which style rule I've broken there.

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
 
 }
 
+/* new function for ENQ */
+void respondWithEnqueue(HttpResponse* response, char* absolutePath, int jsonMode) {
+
+  zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_DEBUG,
+      "ZSS1689I Begin %s\n", __FUNCTION__);
```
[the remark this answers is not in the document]
John-A-Davies: Yes, was debugging, deleted.

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+  extractDatasetAndMemberName(absolutePath, &dsn, &memberName);
+  
+  DynallocDatasetName daDsn;
+  DynallocMemberName daMember;
+  memcpy(daDsn.name, dsn.value, sizeof(daDsn.name));
+  memcpy(daMember.name, memberName.value, sizeof(daMember.name));
+
+  /* new code to detect ENQ header begins */
+  
+  /* new code ends   */
```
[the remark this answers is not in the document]
John-A-Davies: deleted

c/datasetjson.c
```diff
@@ -1630,6 +1640,395 @@ void respondWithDataset(HttpResponse* response, char* absolutePath, int jsonMode
...
+
+/* new code to enqueue dataset starts */
+#include "isgenq.h"
```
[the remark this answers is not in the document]
John-A-Davies: moved to top

c/datasetjson.c
John-A-Davies: It's a hangover from a previous tryout and so should be removed.  I did that for the rest but I missed this one.

NakulManchanda:
@lchudinov i have removed semaphores, as i was already using mutex.
They were un-needed as pointed by you.
Only refactoring is left.

NakulManchanda:
Reverted based on feedback from @lchudinov 

```
actually, the semaphores were needed to hold an ENQ
https://github.com/zowe/zowe-common-c/pull/184/commits/3025452a18d5c54d6187d50dd3c4b6d998e2a7ae#diff-9bfb521705d7475911e[…]04d06c974211df8e53ceL1714
as soon as the thread exits the ENQ automatically gets released (edited) 
but on the other hand the thread must terminate when we have sent a response to the client(browser)
```

NakulManchanda:
**options to research**
May be we can sleep thread without semaphore
Semaphore timeout -> then we have to keep sending message that we are alive
Some flag when acquiring semaphores
if kernel space - mutex exists and have better interface use that instead
or add a cleanup process on crash  and keep using semaphores

PR #185, opened 2020-12-06
maximum of 1000 requests to prevent slh from becoming full

jordanfilteau1995:
The SLH will become full eventually and crash the server. From evidence of testing, 1000 requests should be plenty and won't crash the server. At some point, we should do timeouts, but for now, at least we can do the right thing for Postman, etc.


1000TurquoisePogs (approved):
Merging after having been tested and reviewed.

PR #190, opened 2021-01-11
Make group lookup function for user

lchudinov:
The PR adds a new function that gets a list of groups for a user. The function is similar to [getgrouplist](https://man7.org/linux/man-pages/man3/getgrouplist.3.html)

Example of usage:
```C
  int count = 0;
  int rc = 0;
  int returnCode;
  int reasonCode;

  const char *user = "root";
  // obtain group count for user
  rc = getGroupList(user, NULL, &count, &returnCode, &reasonCode);
  if (rc == -1) {
    printf ("failed to obtain group count for user %s\n", user);
    exit(EXIT_FAILURE);
  }
  printf ("found %d groups\n", count);
  // allocate array for group list
  int *groups = (int*)safeMalloc(sizeof(int) * count, "groups");
  // obtain group list
  rc = getGroupList(user, groups, &count, &returnCode, &reasonCode);
  if (rc == -1) {
    printf ("unable to get groups for user %s\n", user);
    exit(EXIT_FAILURE);
  }
  for (int i = 0; i < count; i++) {
    printf ("group %d\n", groups[i]);
  }
}
```

### DIFF
--- a/c/zosaccounts.c
+++ b/c/zosaccounts.c
@@ -42,12 +42,14 @@ static int accountTrace = FALSE;
 # pragma linkage(BPX4GPN,OS)
 # pragma linkage(BPX4GPU,OS)
 # pragma linkage(BPX4PWD,OS)
+# pragma linkage(BPX4GUG,OS)
 
 # define BPXGGN BPX4GGN
 # define BPXGGI BPX4GGI
 # define BPXGPN BPX4GPN
 # define BPXGPU BPX4GPU
 # define BPXPWD BPX4PWD
+# define BPXGUG BPX4GUG
 
 #else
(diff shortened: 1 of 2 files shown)


daveyc:
For error handling should `exit(0);` be `exit(EXIT_FAILURE);`?

lchudinov:
> For error handling should `exit(0);` be `exit(EXIT_FAILURE);`?

Yes, if it were a real-life program. The exit code is not that important here.

daveyc:
I would change it to `exit(2)`

1000TurquoisePogs (approved):
Looks like exactly what I expected. Thanks!

PR #194, opened 2021-01-19
Check and return variable session timeout info

1000TurquoisePogs:
This makes use of recent PRs https://github.com/zowe/zowe-common-c/pull/192 and https://github.com/zowe/zowe-common-c/pull/190 in order to set cookie timeout according to the user or group the user is in. In the case that a user is specified in the list, that value is chosen. In the case the user is not found in the list, groups are checked. If a user is in multiple groups found, the longest session duration value is chosen. If the value is -1, that means there is no expiration. 0 is considered not found, and when not found, the default value (1 hour) will be chosen.

Because the cookie doesnt and shouldnt contain info such as expiration, and because the cookie is given during any call, a header is needed for sending expiration information.

I did not find any standard or non-standard header for sending session expiration info the way we are trying to do. The closest thing I saw was expiration info for SIP, which just seemed too unrelated to use.

So, I made a new header. Current recommendation of standards community is to NOT prefix such as "X-", so instead I chose a name that is generic and maybe will become used by others some day.

How I tested:

I made a JSON,
```
{
users: {myself: 1000},
groups: {
  one: 2000,
  two: 3000,
  three: -1
  four: 4000,
  five: 500
}

}
```

When groups is not defined, the number returned should be derived from 1000.
When groups is defined, and I'm in all the groups, I expect -1, not 4000 and not 500.

### DIFF
--- a/c/collections.c
+++ b/c/collections.c
@@ -495,12 +495,13 @@ int htRemove(hashtable *ht, void *key){
 
 int stringHash(void *key){
   char *s = (char*)key;
-  int hash = 0;
+  //See: http://www.cse.yorku.ca/~oz/hash.html
+  int hash = 5381;
   int i;
   int len = strlen(s);
 
   for (i=0; i<len; i++){
-    hash = (hash << 4) + s[i];
+    hash = ((hash << 5) + hash) + s[i];
   }
   return hash&0x7fffffff;
 }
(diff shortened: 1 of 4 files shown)


lchudinov (approved):
Works fine to me.

PR #195, opened 2021-01-27
Fix an implicit declaration

ifakhrutdinov:
#### Overview
Function `zosPost` is not properly declared in `c/stcbase.c`, this may lead to unexpected behavior where the compiler assumes that the arguments are all `int`s whereas we need to pass a 64-bit address and an int. This pull-request include `scheduling.h` to ensure the compiler knows the `zosPost` prototype.

### DIFF
--- a/c/stcbase.c
+++ b/c/stcbase.c
@@ -36,6 +36,7 @@
 #include "openprims.h"
 #include "collections.h"
 #include "stcbase.h"
+#include "scheduling.h"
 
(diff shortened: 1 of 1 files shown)


daveyc (approved):
I wish there was a compiler flag which generated errors for implicit declarations!

ifakhrutdinov:
@daveyc I think there is one, but in my case I used `xlclang` ;)

PR #196, opened 2021-01-29
Fix an incorrect if condition in the http socket write function

ifakhrutdinov:
When an additional check was introduced (EWOULDBLOCK), the WRITE_FORCE
value was not correctly applied - it should be AND'ed with both the
EAGAIN and EWOULDBLOCK values:

`WRITE_FORCE and (EAGAIN or EWOULDBLOCK)`


1000TurquoisePogs (approved):
This makes sense.

PR #198, opened 2021-02-23
added dataset heartbeat

joepun76:
In this PR - 
1. Add new endpoint POST /datasetHeartbeat
2. Record heartbeat - username: last heartbeat timestamp
3. New background task to monitor user heartbeat, and release locks if heartbeat timestamp is older than 30 sec

We are using new STC_MODULE_ZSS Module for monitoring, and created new interface 
- addZssBackgroundTask(heartbeatFunction, 'DATASET_HEARTBEAT',30) - every 30 sec we evoke this
- processZssBackgroundHandler - have a queue to support atmost 10 tasks, currently we are just using one, its called every 10 secs, but it checks time requested by task like 30 seconds equivalent to every third time processZssBackgroundHandler is evoked


c/datasetjson.c
```diff
@@ -88,6 +87,16 @@ typedef struct Volser_tag {
 static int getVolserForDataset(const DatasetName *dataset, Volser *volser);
 static bool memberExists(char* dsName, DynallocMemberName daMemberName);
 
+void initDatasetLocking(HttpServer *server) {
+  zowelog(NULL, LOG_COMP_DATASERVICE, ZOWE_LOG_INFO,"initDatasetLocking\n");  
+  // register background handler
+  addZssBackgroudTask(&heartbeatBackgroundHandler,"1dlockbg", 10);
```
DivergentEuropeans, line 93: *addZssBackgroundTask

NakulManchanda:
@ifakhrutdinov we removed all the code from httpserver.c
this is all new code to support POST/datasetheartbeat endpoint
It will be merged to John's original PR
We added generic background handler for zss as well

h/stcbase.h
```diff
@@ -42,6 +42,7 @@
 
 #define STC_MODULE_GENERIC          0x00000000
 #define STC_MODULE_JEDHTTP          0x00010000
+#define STC_MODULE_ZSS              0x00020000
```
NakulManchanda, line 45: @ifakhrutdinov does using 0x00020000 this slot for STC_MODULE_ZSS looks good to you?

c/datasetlock.c
```diff
@@ -0,0 +1,129 @@
...
+  }
+  
+  for(int j=0; j < N_HBT_TABLE_ENTRIES; j++) {
+    hbt_table_entry[j].cnt = -1;  /* initialise */      
+  }
+
+  return;
+}
+
+int addUserToHbt(char user[8]){
```
timgerstel, line 34: Should the heartbeat table also take a dataset as input rather than just a username? One scenario I pictured while testing was:
- A user has many datasets open in the editor but not all are being edited
- Because the datasets are open and with the heartbeat logic. all datasets the user has open in the editor will remain locked

Should the /datasetHeartbeat call take a dataset name as a parameter such that not all locks are refreshed? Perhaps the editor logic could be improved to only call datasetHeartbeat for the dataset currently being edited, or the most recently used datasets.

NakulManchanda: yes, i think we should 
GET /datasetContent
and POST /datasetEnqueue - only when user intends to edit file.

ifakhrutdinov (changes requested):
I'd like to tackle the background function concept first.

In this implementation there doesn't seem to be a way for a plugin to actually provide a function which would be called in background. Everything seems to be pretty much tied to the dataset services.

Here's how I see one way we could extend the framework so that you can add your background functions:
* Extend the `DataService` structure to include a new function field. For example, `intervalHandler` or `intervalFunction` (I'm bad at naming)
* Extend the `DataService` structure to include new fields for the interval value and the last call time value
* Provide a public API to invoke it. For example, a public function in `c/dataservice.c` similar to `initalizeWebPlugin()`. Maybe `webPluginOnInterval()`? It would call the functions which have expired.
* Somewhere in zss.c (`readWebPluginDefinitions()`?), you'd initialize those fields from the plugin JSONs (similar to how the `initializer` field is set).
* The new zss background handler would just go over the dataservices and invoke `webPluginOnInterval()`, which would invoke all the interval functions.

@NakulManchanda, @1000TurquoisePogs, what do you think?

h/zssbackground.h
```diff
@@ -0,0 +1,46 @@
...
+  int id;
+  char taskLabel[LEN_TASK_LABEL];
+  Task task;
+  int timeInterval;
+  int countInterval;
+} Background_Task; 
+
+Background_Task task_list [N_TASK_TABLE_ENTRIES];
+
+int addZssBackgroudTask(Task task, char* taskLabel, int timeInterval);
```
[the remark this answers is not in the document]
NakulManchanda: hi yes, i renamed them. STC_BACKGROUND_MODULE
- people can initialize it if they want
- there is only operation they can do addStcBackgroudTask -

c/zssbackground.c
```diff
@@ -0,0 +1,48 @@
+#include "logging.h"
```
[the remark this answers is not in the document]
NakulManchanda: yes, people can reuse this, i renamed it to stcbackground{.h,.c}
this when registered runs every 10 seconds
but can be use to register any background step in multiples of 10

- eg heartbeat,30 -> will run hearbeat every 3rd time this module master background process runs

h/datasetlock.h
```diff
@@ -0,0 +1,54 @@
+/*
+  This program and the accompanying materials are
```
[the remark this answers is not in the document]
NakulManchanda: it helps implement pessimistic locking solution, i can move it.
it enables support to:
POST /datasetHeartbeat
POST /datasetEnqueue - this calls semTableEnqueue function defined in datasetlock{.h,.c}
DELETE /datasetEnqueue - this calls semTableDequeue function defined in datasetlock{.h,.c}

h/stcbase.h
NakulManchanda: changed the module name to STC_MODULE_BACKGROUND

ifakhrutdinov:
@NakulManchanda why is merged without review? What's the point of review then?

cc: @1000TurquoisePogs

PR #200, opened 2021-03-10
Fixed bugs around the value for session timeout

1000TurquoisePogs:
Various problems fixed: int initialization, using int instead of uint64, and not checking retval before acting upon returned values.


DivergentEuropeans:
Needs a final test from a full Zowe build

PR #202, opened 2021-03-11
Reinstating keep-alive

jordanfilteau1995:
Even though we have logic to close the connection, we aren't using it. I'd like to reinstate the code that was removed temporarily, unless there are any objections.

It goes along with the changes added in this pull request: https://github.com/zowe/zowe-common-c/commit/6cbd244f2147be5e653773ab9bc457cab628c141

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -1933,9 +1933,6 @@ static void addRequestHeader(HttpRequestParser *parser){
       parser->isWebSocket = TRUE;
     }
   }
-  /* Temporarily disable this while investigating and developing solutions to
-     Out of memory issues around the SLH used by each connection running out of memory
-     Due to connections not closing when they should
   else if (!compareIgnoringCase(newHeader->nativeName, "Connection",
                                   parser->headerNameLength)) {
     if (!compareIgnoringCase(newHeader->nativeValue, "Keep-Alive",
@@ -1943,8 +1940,6 @@ static void addRequestHeader(HttpRequestParser *parser){
       parser->keepAlive = TRUE;
     }
   }
-  */
-  parser->keepAlive = FALSE;
 
   HttpHeader *headerChain = parser->headerChain;
   if (headerChain){


ifakhrutdinov:
> Even though we have logic to close the connection, we aren't using it.

What logic do you mean? By not using, do you mean you don't need it?

jordanfilteau1995:
> > Even though we have logic to close the connection, we aren't using it.
> 
> What logic do you mean? By not using, do you mean you don't need it?

I added this a while back. 

https://github.com/zowe/zowe-common-c/commit/6cbd244f2147be5e653773ab9bc457cab628c141

If we don't want it, then we can remove it. response->conversation->isKeepAlive can never be true due to the commented out code. I'm not sure how I missed it. I probably checked out the file again to get rid of added logging when debugging the issue and never reverted that part. 

I was trying to give us more standard behavior according to this: 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

gadyck:
This has nothing to do with the keepalive header.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

gadyck:
Tcpip keepalive processing does not limit the duration of a connection. Rather, it periodically confirms that the connection is alive and well by sending the client a message that must be responded to. If, and only if, the client does not respond does the connection get closed.

I have additional functionality pending for merge after testing completes that times the connection idle time, and terminates the connection when no response is pending and the idle time limit is exceeded. This is close to what the keep-alive header time is trying to indicate. However, the header is only a recommendation, not a mandate, on how long the connection is maintained by the server. The server may set any limit is wishes on how long an idle connection, and associated tcpip port/socket, it held open.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jordanfilteau1995:
> Tcpip keepalive processing does not limit the duration of a connection. Rather, it periodically confirms that the connection is alive and well by sending the client a message that must be responded to. If, and only if, the client does not respond does the connection get closed. I have additional functionality pending for merge after testing completes that times the connection idle time, and terminates the connection when no response is pending and the idle time limit is exceeded. This is close to what the keep-alive header time is trying to indicate. However, the header is only a recommendation, not a mandate, on how long the connection is maintained by the server. The server may set any limit is wishes on how long an idle connection, and associated tcpip port/socket, it held open.

Hi Greg. 

Irek merged this a while back to fix a bug. https://github.com/zowe/zowe-common-c/pull/147/files.

This checks for the keep alive header and I go off of that value. You can see what I did below.

https://github.com/zowe/zowe-common-c/pull/185/files

I did this to stop the server from crashing due to the slh being full by having the same connection. There is no way for the Python code to know how many requests it takes to blow up the server, and Irek told me this issue has been in the server for a while now. It was crashing during our automated testing that attaches keep alive by default.

I don't think your future work addresses this issue because the automated tests are throwing an enormous amount of requests at zss and if we don't try to close the connection, we will just crash the server due to an unchecked slh alloc. Sending a 'connection: close' should make any reasonable REST client close the connection. If it doesn't get closed, we just send an error like we do if recovery fails on the next request. 

I was under the impression that with my pull requests, we'd be able to utilize keep alive on our testing again. I acknowledge I should have put the uncommenting of that in my original pull request...so sorry about the confusion. 

@ifakhrutdinov Do we not want your bug fix? I built on top of it because it was a bug fix that you put in. I didn't think fixing the core slh issue of the server was a reasonable thing to do at the time... I should probably just leave most of this to the Zowe team now-a-days instead of being a cowboy and implementing my own solution because they have more time to spend on it.

gadyck:
An additional thought... I added the ability to enable TCPIP Keep-Alive due to "Zombie Connections" that were causing the maximum supported socket count to be reached on the MFA Web server.  Zombie connections occur when a network or client error occurs, such that the client can not, or does not, send a proper connection close request.  There were connections that had been established weeks prior, and were long gone, that were still open from the server point of view because no close had been received.  The TCPIP Keep-alive processing addresses the issue, by detecting that the client is no longer there and causing an error to be presented for the connection, so that the server gets driven to close the connection and free the resources associated with it.

gadyck:
Enabling TCPIP Keep-alive is only the first step in addressing this.  It addresses long term Zombie connections.  As I said earlier, I have code that is pending merge to common that tracks how long a connection is idle **and a response is not pending** and terminates the connection when the idle time is exceeded (first sending the proper response to the client to say we decided to close it).  The idle time probably will be very short... under a minute.  What the header says is a recommendation, and the header isn't even supported with HTTP 2.0 protocols, so relying on it is a mistake in my opinion.  I think you need the idle timeout support.

jordanfilteau1995:
> Enabling TCPIP Keep-alive is only the first step in addressing this. It addresses long term Zombie connections. As I said earlier, I have code that is pending merge to common that tracks how long a connection is idle **and a response is not pending** and terminates the connection when the idle time is exceeded (first sending the proper response to the client to say we decided to close it). The idle time probably will be very short... under a minute. What the header says is a recommendation, and the header isn't even supported with HTTP 2.0 protocols, so relying on it is a mistake in my opinion. I think you need the idle timeout support.

The connection won't ever be idle long enough in our automated tests for it to close.

gadyck:
I don't understand your concerns.

TCPIP Keep-alive support occurs at the IP transport layer, and is transparent/invisible to the server and client application after being enabled.  *IF* there is ANY activity on the connection IT NEVER DOES ANYTHING.  TCPIP doesn't even send a keepalive message to the client if there has been a recent message received from it.  Enabling TCPIP Keepalive does NOT LIMIT the length of a connection unless the client stops responding to TCPIP messages.  There is no reason not to enable TCPIP Keep-alive processing.

The future Idle timeout support is performed at the server level.  Again, if there is activity on the connection the connection will remain open.  If the test program hangs or stops performing activity the server will close the connection.  This is something that most full function Web servers, such as Apache, already do.

Supporting HTTP 2.0 isn't actually a big deal, in my opinion.  My point is that the HTTP architects recognized that connection keep-alive was a bad idea, and thus dropped it in HTTP 2.0.  Ergo, supporting it at the HTTP 1.x level isn't something to exert energy doing.  The HTTP documentation explicitly states it is a **hint** to the server, not a mandate.

jordanfilteau1995:
> I don't understand your concerns.
> 
> TCPIP Keep-alive support occurs at the IP transport layer, and is transparent/invisible to the server and client application after being enabled. _IF_ there is ANY activity on the connection IT NEVER DOES ANYTHING. TCPIP doesn't even send a keepalive message to the client if there has been a recent message received from it. Enabling TCPIP Keepalive does NOT LIMIT the length of a connection unless the client stops responding to TCPIP messages. There is no reason not to enable TCPIP Keep-alive processing.
> 
> The future Idle timeout support is performed at the server level. Again, if there is activity on the connection the connection will remain open. If the test program hangs or stops performing activity the server will close the connection. This is something that most full function Web servers, such as Apache, already do.
> 
> Supporting HTTP 2.0 isn't actually a big deal, in my opinion. My point is that the HTTP architects recognized that connection keep-alive was a bad idea, and thus dropped it in HTTP 2.0. Ergo, supporting it at the HTTP 1.x level isn't something to exert energy doing. The HTTP documentation explicitly states it is a **hint** to the server, not a mandate.

I'm not doing anything with timeouts. I'm just limiting the maximum number of requests per connection if the Keep-Alive header is present. I understand that if it's not present, it won't work, but is it so bad for my requests to purposely attach the Keep-Alive to prevent me from having any issues in cases where I know I'm going to have a long session? I get that it's a **hint**, but if I, as the client follow that hint purposefully and correctly, I'd get the ability to not abend ZSS without a large change to httpserver. All of this is conditional on the Connection header being sent. It's a choice to send that header, so those who don't send it should see the exact same behavior as before. I'm not suggesting we expose any timeout details or even implement it with keep-alive. 

Connection: Keep-Alive
Keep-Alive: max=1000

PR #206, opened 2021-03-18
add background module

NakulManchanda:
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>


ifakhrutdinov:
@1000TurquoisePogs would we want this to be integrated with data services?

h/stcbackground.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef __STCBACKGROUND__
+#define __STCBACKGROUND__
+
+#include "stcbase.h"
+
+#define N_TASK_TABLE_ENTRIES 101
+#define STC_BACKGROUND_INTERVAL 10
+#define LEN_TASK_LABEL 30
+
+typedef void (*Task)(void* server, void* taskInput);
```
ifakhrutdinov, line 21: We need to design this prototype carefully because we won't be able to change it without breaking backward compatibility. 
So, perhaps:
```c
typedef int (STCIntervalCallback)(struct STCBase_tag *stcBase,
                                  struct STCModule_tag *module, 
                                  struct STCIntervalCallbackData_tag *callbackData, 
                                  void *userData);
```

We should be able to derive all we need from these arguments.

All access to the fields in `callbackData` could be done via public functions, so we can change the structure without breaking old plugins - they all access the fields using those functions which are exported from `zssServer` dynamically at runtime.

For example, in your concrete interval callback you could do something like this even if we shuffle the fields in `struct STCIntervalCallbackData_tag`:
```c
static int checkDatasets(struct STCBase_tag *stcBase,
                         struct STCModule_tag *module, 
                         struct STCIntervalCallbackData_tag *callbackData, 
                         void *userData) {

  . . .
  if (stcIsIntervalTimedOut(callbackData)) {
    // issue a warning
  }

}
```

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
...
+#include "stcbackground.h"
+
+static int processStcBackgroundHandler(STCBase *base, STCModule *module, int selectStatus) {
+  BackgroundModuleData* moduleData = (BackgroundModuleData*)(module->data);
+  BackgroundTask* taskList = moduleData->taskList;
+
+  for (int i = 1; i < N_TASK_TABLE_ENTRIES; i++) {
+    BackgroundTask* t = &taskList[i];
+    if (t->id != 0)  /* initialise */ {
+      t->countInterval+=STC_BACKGROUND_INTERVAL;
```
ifakhrutdinov, line 11: You're assuming that the interval is always equal to `STC_BACKGROUND_INTERVAL`, which is not true.

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
...
+      if (t->countInterval == 0) {
```
ifakhrutdinov, line 12: * What if `countInterval` is not a multiple of `STC_BACKGROUND_INTERVAL`? Wouldn't you end up never calling the callback?

Let's say your interval is 17 and `STC_BACKGROUND_INTERVAL` is 9. According to the logic you'll get the following values:
|Invocation #|countInterval|
| -------------|-----:|
|0|-17|
|1|-8|
|2|1|
|3|10|
|4|19|

* What if `countInterval` is smaller than `STC_BACKGROUND_INTERVAL`? I don't think the callback will be invoked either.

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
...
+    }
+    memcpy(taskList[slotId].taskLabel, taskLabel, len);
+    taskList[slotId].task = task;
+    taskList[slotId].timeInterval = timeInterval;
+    taskList[slotId].countInterval = -timeInterval;
+    taskList[slotId].taskInput = taskInput;
+    return 0;
+  }
+
+  return slotId;
```
ifakhrutdinov, line 72: Why do you return `slotId` here? Is just because it happens to be -1 and you also want to return -1? If so, please return -1 so that your intention is clear.

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
...
+      }
+    } else {
+      break;
+    }
+  }
+  return 0;
+};
+
+static int nextSlot(BackgroundTask* taskList) {
+  for (int i = 1; i < N_TASK_TABLE_ENTRIES; i++) {
```
ifakhrutdinov, line 25: Why do we start from 1? Is it because `id = 0` indicates a uninitialized entry? Could we then just use -1 instead of 0?

ifakhrutdinov (changes requested):
@NakulManchanda thank you for the pull-request. It's really helpful to have these changes as a separate pull-request.

Please see my comment. I believe this needs some more work. After all, this is going to be a pretty generic component so we need to ensure the interfaces are well-defined from the get-go.

c/stcbackground.c
NakulManchanda: its part of logic when to process registered handler, when this counter goes to zero, we execute its associated background handler, and reset it to desired multiple again.
In short, this is what we needed.

c/stcbackground.c
NakulManchanda: you are correct here, there are no check around it, just an assumption that it should always be equal to or greater and is a multiple of STC_BACKGROUND_INTERVAL.
Let me see what error handling i can put around this

c/stcbackground.c
NakulManchanda: ok, chaning

c/stcbackground.c
NakulManchanda: changed to if (t->countInterval <= 0) instead, if its negative or turned negative it will execute

c/stcbackground.c
NakulManchanda: this is fine

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
+#include "logging.h"
+#include "stcbackground.h"
+
+static int processStcBackgroundHandler(STCBase *base, STCModule *module, int selectStatus) {
```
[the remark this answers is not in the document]
NakulManchanda: changed

c/stcbackground.c
```diff
@@ -0,0 +1,73 @@
...
+  for (int i = 1; i < N_TASK_TABLE_ENTRIES; i++) {
+    if (taskList[i].id == 0)  /* initialise */ {
+      return i;
+    }
+  }
+  return -1;
+};
+
+STCModule* initBackgroundModule(void* server, STCBase *base) {
+  BackgroundModuleData* moduleData = (BackgroundModuleData*)safeMalloc(sizeof(BackgroundModuleData),"BackgroundModuleData");
```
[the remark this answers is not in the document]
NakulManchanda: checking and returning NULL now after both safeMalloc

h/stcbackground.h
```diff
@@ -0,0 +1,48 @@
...
+
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifndef __STCBACKGROUND__
+#define __STCBACKGROUND__
+
+#include "stcbase.h"
+
+#define N_TASK_TABLE_ENTRIES 101
```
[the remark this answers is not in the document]
NakulManchanda: new name -> STC_BG_ENTRIES

h/stcbackground.h
```diff
@@ -0,0 +1,48 @@
...
+  int timeInterval;
+  int countInterval;
+} BackgroundTask; 
+
+typedef struct BackgroundModuleData_type {
+  void* server;
+  void* taskList;
+} BackgroundModuleData;
+
+STCModule* initBackgroundModule(void* server, STCBase *base);
```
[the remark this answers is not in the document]
NakulManchanda: done

h/stcbase.h
```diff
@@ -42,7 +42,7 @@
 
 #define STC_MODULE_GENERIC          0x00000000
 #define STC_MODULE_JEDHTTP          0x00010000
-
+#define STC_MODULE_BACKGROUND       0x00020000
```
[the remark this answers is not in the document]
NakulManchanda: done

h/stcbackground.h
```diff
@@ -0,0 +1,48 @@
...
+
+typedef void (*Task)(void* server, void* taskInput); 
+typedef struct BackgroundTask_type {
+  int id;
+  char taskLabel[LEN_TASK_LABEL];
+  Task task;
+  void* taskInput;
+  int timeInterval;
+  int countInterval;
+} BackgroundTask;
```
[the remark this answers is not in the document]
NakulManchanda: its all callback now, not task

h/stcbackground.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef __STCBACKGROUND__
+#define __STCBACKGROUND__
+
+#include "stcbase.h"
+
+#define N_TASK_TABLE_ENTRIES 101
+#define STC_BACKGROUND_INTERVAL 10
+#define LEN_TASK_LABEL 30
+
```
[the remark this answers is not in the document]
NakulManchanda: moved except LEN_TASK_LABEL

h/stcbackground.h
NakulManchanda: modified

NakulManchanda:
@ifakhrutdinov addressed all feedback, let me know if it looks better now
- added one new function `stcModifyInterval`, just to change timeInterval, as we have to keep in sync countInterval=-timeInterval,
- also if timeInterval<0 we disable the job

All changes in this last commit
https://github.com/zowe/zowe-common-c/pull/206/commits/34eb713ccdc3e10604e03cabf9a55f9974919dc7

h/stcbackground.h
NakulManchanda: just added one more extra argument in signature, `void* server`, as server may have context too for callbacks

```
typedef int (*STCIntervalCallback)(void* server,
                                  struct STCBase_tag *stcBase,
                                  struct STCModule_tag *module, 
                                  STCIntervalCallbackData* callbackData, 
                                  void *userData);
```

NakulManchanda:
all changes are tested after refactoring, seems to work as expected.

h/stcbackground.h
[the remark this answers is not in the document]
ifakhrutdinov: What about `STCIntervalCallbackData` and `STCCallbackList`?

h/stcbackground.h
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+#ifndef __STCBACKGROUND__
+#define __STCBACKGROUND__
+
+#include "stcbase.h"
+#define STC_BG_CB_LABEL_LEN 30
+
+struct STCIntervalCallbackData_tag;
+typedef int (*STCIntervalCallback)(void* server,
```
ifakhrutdinov, line 19: What's `server` for? I don't think we need it here. We already have `userData` for passing the data needed by the callbacks.

h/stcbackground.h
```diff
@@ -0,0 +1,54 @@
...
+  SPDX-License-Identifier: EPL-2.0
+
+  Copyright Contributors to the Zowe Project.
+*/
+
+#ifndef __STCBACKGROUND__
+#define __STCBACKGROUND__
+
+#include "stcbase.h"
+#define STC_BG_CB_LABEL_LEN 30
```
ifakhrutdinov, line 16: Same as the structures below, this doesn't belong here,

h/stcbackground.h
```diff
@@ -0,0 +1,54 @@
...
+  int countInterval;
+} STCIntervalCallbackData; 
+
+typedef struct STCCallbackList_tag {
+  void* server;
+  void* callbackList;
+} STCCallbackList;
+
+STCModule* stcInitBackgroundModule(void* server, STCBase *base);
+int stcAddIntervalCallback(STCModule *module, STCIntervalCallback callback, char* callbackLabel, int timeInterval, void* userData);
```
ifakhrutdinov, line 41: What unit is `timeInterval` in? Is it in milliseconds?

c/stcbackground.c
```diff
@@ -0,0 +1,98 @@
...
+
+  for (int i = 0; i < STC_BG_ENTRIES; i++) {
+    STCIntervalCallbackData* t = &callbackList[i];
+    if (t->id != -1)  /* initialise */ {
+      
+      // giving ability to disable job by making timeInterval less than zero
+      // as we return callbackdata, user can make timeInterval zero
+      if(t->timeInterval<0) continue;
+
+      t->countInterval+=STC_BG_INTERVAL;
```
ifakhrutdinov, line 20: The STC base can have a variable timeout (see the `selectTimeoutMillis` argument in `stcBaseMainLoop()`)? How does this code take this into account? I don't think `STC_BG_INTERVAL` will work.

c/stcbackground.c
```diff
@@ -0,0 +1,98 @@
...
+      break;
+    }
+  }
+  return 0;
+};
+
+static int nextSlot(STCIntervalCallbackData* callbackList) {
+  for (int i = 0; i < STC_BG_ENTRIES; i++) {
+    if (callbackList[i].id == -1)  /* initialise */ {
+      return i;
```
ifakhrutdinov, line 36: Don't we need to return `id`? If not, why is the field called `id`? Should it be `bool isInitialized`?

h/stcbackground.h
ifakhrutdinov, line 41: Use `const` with the argument you're not going to mutate. `char* callbackLabel` should be const.

h/stcbackground.h
```diff
@@ -0,0 +1,54 @@
...
+int stcModifyInterval(STCIntervalCallbackData* callback, int newInterval);
```
ifakhrutdinov, line 42: Let's not add public functions if we don't need them. 

This function exposes `STCIntervalCallbackData` which shouldn't be exposed. You've already identified your callback using a label, so we'd want to keep using it to address the callback.

c/stcbackground.c
```diff
@@ -0,0 +1,98 @@
...
+};
+
+STCModule* stcInitBackgroundModule(void* server, STCBase *stcBase) {
+  STCCallbackList* moduleData = (STCCallbackList*)safeMalloc(sizeof(STCCallbackList),"STCCallbackList");
+  if (moduleData == NULL) {
+    return NULL;
+  }
+
+  STCIntervalCallbackData* callbackList = (STCIntervalCallbackData*)safeMalloc(sizeof(STCIntervalCallbackData)*STC_BG_ENTRIES,"STCIntervalCallbackData");
+  if (callbackList == NULL) {
```
ifakhrutdinov, line 49: You're leaking `moduleData`. We'd also want to free both `callbackList` and `moduleData` if `stcRegisterModule` returns NULL.

c/stcbackground.c
```diff
@@ -0,0 +1,98 @@
...
+int stcAddIntervalCallback(STCModule *module, STCIntervalCallback callback, char* callbackLabel, int timeInterval, void* userData) {
+  STCCallbackList* moduleData = (STCCallbackList*)(module->data);
+  STCIntervalCallbackData* callbackList = moduleData->callbackList;
+
+  int len;
+  int slotId = nextSlot(callbackList);
+  if (slotId >= 0 && slotId<STC_BG_ENTRIES) {
+    callbackList[slotId].id = slotId;
+    len = strlen(callbackLabel);
+    if (len > STC_BG_CB_LABEL_LEN-1) {
```
ifakhrutdinov, line 80: Is a label supposed to uniquely identify a callback? If yes, what happens if a label is too big, it gets truncated and you potentially have 2 identical labels? If no, what does uniquely identifies callbacks then? A label seems like a good choice for such purpose, but then we need to make sure there are no duplicate labels and if there is a duplicate we need to return an error code.

ifakhrutdinov (changes requested):
@NakulManchanda, please see my comments. I've only identified larger issues and have not mentioned smaller ones like ambiguous names (for example, `t`) and inconsistent style (some operands have spaces around them and some don't).

c/stcbackground.c
NakulManchanda: label can be used for logging and debugging, just human readable string.
should be unique, but wont harm anything if duplicated
Will check for uniqueness while iterating for empty slot

c/stcbackground.c
NakulManchanda: yes, ok

h/stcbackground.h
NakulManchanda: i see, good choice

just one question though:
what about exposing it here
```
typedef int (*STCIntervalCallback)(void* server,
                                  struct STCBase_tag *stcBase,
                                  struct STCModule_tag *module, 
                                  STCIntervalCallbackData* callbackData, 
                                  void *userData);
```

h/stcbackground.h
NakulManchanda: ok

h/stcbackground.h
ifakhrutdinov: I think we should but we shouldn't expose the field in the header. That way we can change the internal implementation and provide public functions to access them. This will ensure backward/forward compatibility.

h/stcbackground.h
NakulManchanda: removed server

h/stcbackground.h
NakulManchanda: renamed it to interval seconds

h/stcbackground.h
NakulManchanda: i have moved one structure, but as stccallbackdata is part of callback signature, so kept that in .h

we are exposing   struct STCIntervalCallbackData_tag* callbackData, 
so kept the definition in .h and related constant
```
typedef int (*STCIntervalCallback)(STCBase* stcBase, 
                                    STCModule* module, 
                                    struct STCIntervalCallbackData_tag* callbackData, 
                                    void* userData
                                   );
```

c/stcbackground.c
NakulManchanda: yes, you are right, we dont have have way to configure this value, changed name, hope it helps

```
#define MAIN_WAIT_MILLIS 10000
#define STC_BG_MIN_INTERVAL_SECS MAIN_WAIT_MILLIS/1000
```

c/stcbackground.c
NakulManchanda: `id` value is same as index location in array, which makes it id
you are right not using it for any purpose other than like boolean.
still kept it, can remove it if you want.

h/stcbackground.h
NakulManchanda: changed

c/stcbackground.c
NakulManchanda: doing allocation together now, so should be good now

c/stcbackground.c
NakulManchanda: add label duplication check, and improved on truncation logic, thanks!

h/stcbackground.h
NakulManchanda: moved struct definition to .c

h/stcbackground.h
NakulManchanda: moved struct definition to .c

h/stcbackground.h
NakulManchanda: hi @ifakhrutdinov , i believe then we should remove it from signature as well, as public functions can use callback label, so reference is not needed.

Moved all structure definitions to .c

NakulManchanda: its in alignment with your comment above
>You've already identified your callback using a label, so we'd want to keep using it to address the callback.

NakulManchanda:
Hi @ifakhrutdinov ,

Its ready for review again,
Unit tested various parts, **truncating label**, **checking duplicate label**, **modifying time interval**, and **original PR**
**All working good after this refactor.**
Let me know if any more suggestions, would like to merge `dataset locking` as well, once this is done.
Thanks!

NakulManchanda:
Hi @ifakhrutdinov , can we get review on latest refactoring. Thanks!

c/stcbackground.c
```diff
@@ -0,0 +1,153 @@
...
+  return stcBackgroundModule;
+};
+
+const char* truncateLabel(const char* callbackLabel) {
+  int len = strlen(callbackLabel);
+  if (len > STC_BG_CB_LABEL_LEN-1) {
+    len = STC_BG_CB_LABEL_LEN-1;
+  }
+
+  char* labelCopy = (char*)safeMalloc((len+1)*sizeof(char),"char");
```
ifakhrutdinov, line 102: Allocations fails. You can avoid allocations.

c/stcbackground.c
```diff
@@ -0,0 +1,153 @@
...
+
+int stcAddIntervalCallback(STCModule *module, STCIntervalCallback callback, const char* callbackLabel, int intervalSeconds, void* userData) {
+  STCCallbackList* moduleData = (STCCallbackList*)(module->data);
+  STCIntervalCallbackData* callbackList = moduleData->callbackList;
+
+  int len;
+  int slotId = nextSlot(callbackList);
+  if (slotId >= 0 && slotId < STC_BG_ENTRIES) {
+
+    const char* labelTruncated = truncateLabel(callbackLabel);
```
ifakhrutdinov, line 117: Check if the label is too long and return an error if it is. Then proceed with your logic. So no need to alloc/free anything.

c/stcbackground.c
ifakhrutdinov: Again, your value is not tied to `selectTimeoutMillis` in any way. This will not work correctly.

c/stcbackground.c
ifakhrutdinov: Well, my point is the code shouldn't be ambiguous. If someone unfamiliar looks at it and they'll probably ask, what does id have to do with initialization? Anyway, since this is in the C file and we can change it without breaking backward compatibility. At this point, I just want this code to have the decent API and work correctly. It's up to you. 

@lchudinov how does this look to you?

h/stcbackground.h
ifakhrutdinov: You can keep it in the signature of `STCIntervalCallback` by using forward declaration. Concrete callbacks can use it to access internal stuff via public functions. Concrete callbacks don't usually know their label. The label is useful for functions like "removeCallback".

ifakhrutdinov (commented):
@1000TurquoisePogs , @lchudinov, what concerns me in this PR is handling of intervals. It's not aware of the interval passed to `stcBaseMainLoop` so we can't reliably invoke callback on their interval. 

Alternatively this code would measure time between calls before deciding whether to invoke the callbacks, right now it's counting and I don't think it will work correctly without taking into account the `selectTimeoutMillis`.

Can we just check last time we were invoked, calculate elapsed time and decide if a certain callback needs to be called?

c/stcbackground.c
lchudinov: `bool isInitialized` is a field with an apt name and type. `id` looks misleading to me.

lchudinov:
> Can we just check last time we were invoked, calculate elapsed time and decide if a certain callback needs to be called?
@ifakhrutdinov I agree with you. It would be easy and reliable approach.

NakulManchanda:
Hi @ifakhrutdinov @lchudinov 
Thanks for review.

- We can use time elapsed for callback instead of interval logic.
- will throw error is label too long
- will review index vs IsInitialized as well.

need to check what `selectTimeoutMillis` is, do we relinquish control after that value?

c/stcbackground.c
NakulManchanda: modified

c/stcbackground.c
NakulManchanda: removed as mentioned below

c/stcbackground.c
NakulManchanda: added new error

NakulManchanda:
@lchudinov @ifakhrutdinov addressed all concerns, should look good now

c/stcbackground.c
NakulManchanda: @ifakhrutdinov  should look good now, i am using time of last run to compute, whether to run again or not, thanks!

c/stcbackground.c
```diff
@@ -0,0 +1,160 @@
...
+  Copyright Contributors to the Zowe Project.
+*/
+
+#include <time.h>
+#include "logging.h"
+#include "stcbackground.h"
+
+#define MAIN_WAIT_MILLIS 10000
+#define STC_BG_ENTRIES 101
+#define STC_BG_MIN_INTERVAL_SECS MAIN_WAIT_MILLIS/1000
```
lchudinov, line 16: STC_BG_MIN_INTERVAL_SECS is unused

c/stcbackground.c
```diff
@@ -0,0 +1,160 @@
...
+    callbackList[slotId].intervalSeconds = intervalSeconds;
+    callbackList[slotId].timeOfLastRun = (unsigned)time(NULL);
+    callbackList[slotId].userData = userData;
+    return 0;
+  }
+
+  return -1;
+};
+
+//if intervalSeconds < 0, we disable the job
```
lchudinov, line 136: @NakulManchanda what's use case for this function?

NakulManchanda: Modify interval? Change how often we need to call added job.
Additional, if user set negative interval, job is disabled

c/stcbackground.c
NakulManchanda: thanks for catching, removed

NakulManchanda:
@ifakhrutdinov @lchudinov if all good lets merge this, thanks!

c/stcbackground.c
NakulManchanda: removed modify interval

NakulManchanda:
if all looks good lets merge this @lchudinov

lchudinov (approved):

1000TurquoisePogs:
OK, I'm going to merge this since it appears that all comments have been addressed, but if there are still things to be done then please note it so that we can create followup PRs.

PR #208, opened 2021-04-01
HTTP Client: Handle empty status string

lchudinov:
## Fix for HTTP Client
This PR allows to handle correctly statuses with empty status message e.g. `HTTP/1.1 404` (compare with `HTTP/1.1 404 Not Found` which has non empty status message)


1000TurquoisePogs (approved):

rocketjared (commented):
Good change; thank you Leonty

PR #209, opened 2021-04-05
Fix for dataset write when recfm=FB

lchudinov:
Fix for dataset write when recfm=FB. The code mistakenly called `safeFree()` for memory allocated in `ShortLivedHeap`.

PR for testing: https://github.com/zowe/zss/pull/276


1000TurquoisePogs (approved):
This looks good to me.

PR #213, opened 2021-04-19
Handle files with unknown extension

lchudinov:
Signed-off-by: Leonty Chudinov <lchudinov@rocketsoftware.com>

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -3851,13 +3851,14 @@ HttpRequestParam *getCheckedParam(HttpRequest *request, char *paramName){
   return NULL;
 }
 
-static char *getMimeType2(char *extension, int *isBinary, int dotPos);
+static char *getMimeType2(char *extension, int *isBinary, int dotPos, int ccsid);
 
 char *getMimeType(char *extension, int *isBinary) {
-  getMimeType2(extension, isBinary, FALSE);
+  getMimeType2(extension, isBinary, FALSE, -1);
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs (changes requested):
This does solve handling of unknown extensions but continues to assume binary status of known extensions.
In this code, if a user has a .bin file that is text, and tagged it as text, zss would still send it as binary because of ignoring the ccsid when finding certain extensions.

the isBinary boolean shouldn't be modified in the the getMimeType2 extension cases if the function starts with a ccsid.

c/httpserver.c
```diff
@@ -3851,116 +3851,87 @@ HttpRequestParam *getCheckedParam(HttpRequest *request, char *paramName){
...
+static MimeType MIME_TYPES[] = {
+  {"js", "text/javascript", FALSE},
+  {"json", "application/json", FALSE},
+  {"ts", "text/typescript", FALSE},
+  {"c", "text/plain", FALSE},
+  {"py", "text/plain", FALSE},
+  {"cbl", "text/plain", FALSE},
+  {"asm", "text/plain", FALSE},
+  {"cpp", "text/plain", FALSE},
+  {"csv", "text/cvs", FALSE},
```
1000TurquoisePogs, line 3875: typo: text/csv

1000TurquoisePogs (approved):

PR #217, opened 2021-05-19
Unix-file contents: Properly handle empty files

lchudinov:
This PR fixes a bug when downloading hangs for empty file with binary tag or without tags.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -4419,7 +4419,7 @@ static int streamBinaryForFile2(HttpResponse *response, Socket *socket, UnixFile
       zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG2,
               "Binary streaming has ended. (return = 0x%x, reason = 0x%x)\n",
               returnCode, reasonCode);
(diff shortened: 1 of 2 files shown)


c/httpserver.c
```diff
@@ -4419,7 +4419,7 @@ static int streamBinaryForFile2(HttpResponse *response, Socket *socket, UnixFile
       zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG2,
               "Binary streaming has ended. (return = 0x%x, reason = 0x%x)\n",
               returnCode, reasonCode);
-      return 0;
+      break;
```
lchudinov, line 4422: Don't return too early, finish the response properly.

ifakhrutdinov (approved):
This looks good, but could we report any errors to the user if `fileRead` breaks (no just EOF) half way?

lchudinov:
> This looks good, but could we report any errors to the user if `fileRead` breaks (no just EOF) half way?

At that point the HTTP response headers have been sent to the user, so we're unable to change the response status.

PR #218, opened 2021-05-24
Introduce RBAC checking via custom authHandler function

DivergentEuropeans:
PR (2 of 2)
PR 1: https://github.com/zowe/zss/pull/281


lchudinov:
The code is questionable because it introduces a dependency on zss by calling functions defined there.

Could we introduce `int registerHttpMiddleware(HttpServer *server, HttpMiddleware *middleware);` where `HttpMiddleware` has a pointer to a callback that HttpServer would call at the beginning of `serveRequest` e.g.

`bool MiddlewareCallback(HttpServer *server, HttpRequest *request, HttpMiddleware *middleware);`

c/httpserver.c
```diff
@@ -3451,23 +3453,26 @@ static int handleHttpService(HttpServer *server,
...
-    int size = sizeof server->authHandler / sizeof server->authHandler[0];
-    for (int i = 0; i < size; i++) {
-      if (server->authHandler[i] != NULL) {
-        if (strcmp(server->authHandler[i]->type, SERVICE_AUTH_NATIVE_WITH_SESSION_TOKEN) == 0)
-        {
-          int rc = service->server->authHandler[i]->authFunction(conversation, request, service, response);
-          if (rc != 0) {
-            respondWithError(response, HTTP_STATUS_UNAUTHORIZED, "Not Authorized");
+    if (request->authenticated) {
+      int size = sizeof(server->authHandler)/sizeof(server->authHandler[0]);
```
ifakhrutdinov, line 3457: Nitpick, but is there a reason not to have blanks around `/`? The previous version had them, so does the rest of the new code (around operators).

c/httpserver.c
```diff
@@ -3476,6 +3481,8 @@ static int handleHttpService(HttpServer *server,
     } else {
       respondWithAuthError(response, &authResponse);
     }
+  } else if (!authorized) {
+    respondWithError(response, HTTP_STATUS_FORBIDDEN, "Forbidden");
```
ifakhrutdinov, line 3485: How does this affect the endpoints which are open to everyone? Do we want authorization be ignored the same way authentication is when `SERVICE_AUTH_FLAG_OPTIONAL`?

c/httpserver.c
```diff
@@ -6001,7 +6025,27 @@ int mainHttpLoop(HttpServer *server){
   return stcBaseMainLoop(base, MAIN_WAIT_MILLIS);
 }
 
-
+void registerHttpAuthorizationHandler(HttpServer *server, int authorizationType, AuthorizationHandler *authorizationHandler, void *userData) {
```
ifakhrutdinov, line 6028: Would we want to return an int to indicate success/failure? Maybe 0 is a success, and -1 (or some other non-zero) is a failure (in your case an allocation failure). There are functions in `httpserver.c` that seem to do that.

lchudinov: I've added a return code for `registerHttpAuthorizationHandler`.

h/httpserver.h
```diff
@@ -146,6 +146,7 @@ typedef int HttpServiceServe(struct HttpService_tag *service, HttpResponse *resp
 typedef int AuthExtract(struct HttpService_tag *service, HttpRequest *request);
 typedef int AuthValidate(struct HttpService_tag *service, HttpRequest *request);
 typedef int HttpServiceInsertCustomHeaders(struct HttpService_tag *service, HttpResponse *response);
+typedef int AuthorizationHandler(struct HttpService_tag *service, HttpRequest *request, HttpResponse *response, void *userData);
```
ifakhrutdinov, line 149: Can we prepend this typedef with `Http` to avoid name clashes. The reason is, `AuthorizationHandler` seems to be too generic. So perhaps, `HttpAuthorize` (the rest of the typedefs are verbs)?

ifakhrutdinov (approved):
lgtm, @lchudinov please also have a look, I think you're more familiar with the changes and the goals of those changes

lchudinov (approved):

ifakhrutdinov (approved):
lgtm, @rocketjared, perhaps you're interested in these changes.

@lchudinov, please re-review, it's always good to go over changes related to security multiple times

PR #226, opened 2021-07-28
Feature/zss64 2021

lchudinov:
## ZSS 64-bit


c/radmin.c
```diff
@@ -136,40 +134,121 @@ static void freeXTROutput(RadminXTROutput *data) {
...
 ) {
 
-  void * __ptr32 * __ptr32 resultPtr = result;
-#ifdef METTLE
-  resultPtr = (void * __ptr32 * __ptr32)((int32_t)resultPtr | 0x80000000);
-#endif
+
+  /* get the immediate data that needs to have 31 bit pointers into 31 bit area and
+      make some pointers that point in. */
+  char *__ptr32 data31 = (char *__ptr32)safeMalloc31(DATA31_LENGTH,"IRRSEQ00Data:31");
```
ifakhrutdinov, line 228: Can we require the caller to pass 31-bit data so that we don't have to allocate it here and the free it? The caller could use `ALLOC_STRUCT31` and `FREE_STRUCT31` which in most of the cases would not result in a heap allocation.

The same applies to all the new `safeMalloc31` calls. Otherwise the current code would need to handle allocation errors and the callers ( which is mainly ZIS, Metal 64-bit) would also have to do that. The corresponding ZIS services don't expect ABENDs in the radmin code.

Since the services which use radmin are executed in PC-cp, it's possible we'll run into alloc issues because we don't know which address space we're using.

c/radmin.c
```diff
@@ -180,57 +259,96 @@ static void invokeRadmin(
...
-      racfReturnCodeALET, racfReturnCode,
-      racfReasonCodeALET, racfReasonCode,
-      functionCode,
-      parmList,
-      userID,
-      acee,
-      resultSubpool,
-      resultPtr
-  );
+  __asm(ASM_PREFIX
```
ifakhrutdinov, line 262: Would it be possible to link radmin API the old way to get the entry point and then invoke that entry point via an inline HLASM? That was there won't be any questions in case the below isn't common knowledge.

c/shrmem64.c
```diff
@@ -45,8 +46,118 @@ static int makeRSN(int shrmem64RC, int iarv64RC, int iarv64RSN) {
...
+  char       unknown145[7];
+  /* offset 0x98 */
+  uint64_t   mapPageTable;
+  /* this next field is only in plist version = 5 and later */
+  /* uint64_t   units;  
+     char       flags9; 
+     char       flags10; 
+     char       flags11; 
+   */
+} IARV64_V4Parms;
```
ifakhrutdinov, line 153: This parmlist and `#define`s aren't used. Can we remove them?

ifakhrutdinov (commented):
@lchudinov, please see my comment

Regarding radmin, pe, and shrmem64. For the radmin changes the ZIS service which use it need to be tested. The AUX code will need to be tested to ensure the pause element and shrmem64 changes haven't broken anything. 

You may also find some tests in https://github.com/ifakhrutdinov/zowe-common-c/tree/dev/test. Maybe it's worth checking against them too.

Regarding zos.h. Please double check if the new data structures are available in open sources.

If I find anything else, I'll let you know.

1000TurquoisePogs (approved):

ifakhrutdinov:
@1000TurquoisePogs, @lchudinov, before this gets merged can you please have a look at my comments?

ifakhrutdinov:
@JoeNemo, I've opened https://github.com/zowe/zowe-common-c/pull/255 to simplify the shrmem64 changes. Can you please check if your tests pass?

Apologies for the formatting changes.

lchudinov:
Our tests went well for both 31 and 64 bit versions.

PR #245, opened 2021-09-29
E tag in body

joepun76:
To add etag in dataset.json.


c/httpserver.c
```diff
@@ -2840,6 +2840,7 @@ static int getUserSessionValidity(char *username, const HttpServerConfig *config
       int currentValiditySec = 0;
       for (int i = 0; i < groupCount; i++) {
         retVal = getGroupSessionValidity(POINTER_FROM_INT(groups[i]), config, &currentValiditySec, returnCode, reasonCode);
+
```
ifakhrutdinov, line 2843: Do we need this change?

c/datasetjson.c
```diff
@@ -110,6 +118,16 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter
       }
     }
     fclose(in);
+    rcEtag = icsfDigestFinish(&digest, hash);
+
+    // Convert hash text to hex.
+    char eTag[70] = "";
+    memset(eTag, '\0', strlen(eTag));
```
ifakhrutdinov, line 125: `strlen(eTag)` will return 0. You can just use the init method I mentioned above.

c/datasetjson.c
```diff
@@ -90,15 +91,22 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter
...
+  ICSFDigest digest;
+  char hash[32];
+
   int bufferSize = recordLength+1;
   char buffer[bufferSize];
   jsonStartArray(jPrinter,"records");
   int contentLength = 0;
   int bytesRead = 0;
   if (in) {
+	rcEtag = icsfDigestInit(&digest, ICSF_DIGEST_SHA1);
```
ifakhrutdinov, line 105: I believe this return code needs to be checked.

c/datasetjson.c
```diff
@@ -110,6 +118,16 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter
       }
     }
     fclose(in);
+    rcEtag = icsfDigestFinish(&digest, hash);
+
+    // Convert hash text to hex.
+    char eTag[70] = "";
```
ifakhrutdinov, line 124: Is 70 enough? The hash buffer is 32 bytes, which would be 64 chars. Then 6 chars are in the tag and a null-term. That's 71. Can you ensure that even if the text changes there is enough space? Maybe something like 

```c
#define ETAG_TEXT "etag->"
char eTag[sizeof(digest.hash) * 2 + sizeof(ETAG_TEXT)] = {ETAG_TEXT};
```
The size of `eTag` should then be compile time.

c/datasetjson.c
```diff
@@ -110,6 +118,16 @@ int streamDataset(Socket *socket, char *filename, int recordLength, jsonPrinter
...
+    memset(eTag, '\0', strlen(eTag));
+    sprintf(eTag, "%6s", "eTag->");
```
ifakhrutdinov, line 126: I haven't tested [simpleHexPrint](
https://github.com/zowe/zowe-common-c/blob/staging/c/utils.c#L549), but it seems promising. The code could something like this:
```c
simpleHexPrint(&eTag[sizeof(ETAG_TEXT) - 1], hash, digest.hashLength);
```

ifakhrutdinov (changes requested):

c/httpserver.c
joepun76: Irek,

This is not from me, I do not why it's there.  I only add code to datasetjson.c.  Would it be merged when I did the PR?

Thanks.

1000TurquoisePogs:
My misunderstanding: I asked for this PR to be created because I thought it was the whole solution, but this is just the part for hashing. Needs to be combined with locking logic and to implement the API seen in the mock server here https://github.com/zowe/zss/blob/staging/mock/server.py#L386-L418 so we'll close this PR and continue to implement that, and incorporate changes based on Irek's feedback.
Thanks!

PR #248, opened 2021-10-05
entity length check fix

adityadr-joy:
Added length check of entity
Updated CHANGELOG.md

### DIFF
--- a/c/zos.c
+++ b/c/zos.c
@@ -749,9 +749,9 @@ static char *makeCountedString(char *name,
 {
   int allocSize = 0;
   int len = string ? strlen(string) : 0;
-  if (len > maxLength){
+  if (len > maxLength || (len == 0 && isEntity)){
     if (safTrace){
-      printf("%s too long\n", name);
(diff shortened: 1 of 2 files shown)


c/zos.c
```diff
@@ -749,9 +749,9 @@ static char *makeCountedString(char *name,
 {
   int allocSize = 0;
   int len = string ? strlen(string) : 0;
-  if (len > maxLength){
+  if (len > maxLength || (len == 0 && isEntity)){
```
ifakhrutdinov, line 752: What does this accomplish? Currently `safAuth_internal` will return ABEND info from SAF, not it seems to silently ignore such calls.

Let the users of the SAF wrapper deal with that, they have more context.

ifakhrutdinov (changes requested):

c/zos.c
adityadr-joy: Yes, we can add this check in user of library but i felt it is always better to have boundary checks in library to make it robust. 
If you feel this change is not necessary then i can cancel this commit.

adityadr-joy:
These changes are not required, so closing this PR.

PR #253, opened 2021-10-12
JWT verification using JWK

lchudinov:
## Proposed changes
The PR adds an ability to provide a callback function for JWT signature verification.

This PR is a prerequisite for https://github.com/zowe/zss/pull/363

### DIFF
--- a/jwt/jwt/jwt.h
+++ b/jwt/jwt/jwt.h
@@ -30,6 +30,8 @@
 #define RC_JWT_CONTEXT_ALLOCATION_FAILED 11
 #define RC_JWT_CRYPTO_TOKEN_NOT_FOUND   12
 #define RC_JWT_KEY_NOT_FOUND     13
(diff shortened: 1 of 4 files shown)


jwt/jwt/jwt.h
```diff
@@ -108,7 +115,7 @@ int setJwtTrace(int toWhat);
 
 int jwtParse(const char *base64Text,
              bool ebcdic,
-             const ICSFP11_HANDLE_T *keyHandle,
+             const JwtContext *self,
```
ifakhrutdinov, line 118: Are we okay with changing a public API? This doesn't seem like a backward compatible change.

1000TurquoisePogs: While we could postpone this until zowe v2 as a breaking change, I don't think its necessary in that I don't know of any users other than us for this feature. I'm ok with it.

1000TurquoisePogs (approved):

PR #255, opened 2021-10-31
Simplify shrmem64 changes

ifakhrutdinov:


JoeNemo (commented):
Smart change.  Glad you improved my crummy copy-paste code.   I didn't see that you could avoid the caller-key case that way.

JoeNemo (commented):
Sorry to be sloppy.  Will try to factor better when I modify your code.

JoeNemo (approved):
Looks good to me

PR #257, opened 2021-11-08
SAF IDT support

vit-tomica:
This PR is a prereq for https://github.com/zowe/zss/pull/370

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -2440,9 +2440,16 @@ static int proxyServe(HttpService *service,
 static char *getCookieValue(HttpRequest *request, char *cookieName){
   HttpHeader *cookieHeader = getHeader(request,"Cookie");
   ShortLivedHeap *slh = request->slh;
-  char *cookieText = cookieHeader->nativeValue;
-  int cookieTextLength = strlen(cookieText);
-  int cookieNameLength = strlen(cookieName);
+  int cookieTextLength = 0;
+  int cookieNameLength = 0;
+  char *cookieText = NULL;
+  if (cookieHeader != NULL){
+    cookieText = cookieHeader->nativeValue;
+    cookieTextLength = strlen(cookieText);
+    cookieNameLength = strlen(cookieName);
+  } else{
+    return NULL;
+  }
   int pos = 0;
   while (pos<cookieTextLength){
     if (isspace(cookieText[pos])) {
@@ -2499,7 +2506,7 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
      }
   }
   if (traceAuth){
-    printf("safAuthenticate\n");
+    printf("safAutheniticate: authDataFound=%d\n",authDataFound);
   }
   if (authDataFound) {
     ACEE *acee = NULL;
@@ -2532,7 +2539,6 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
     int pwdCheckRC = 0, pwdCheckRSN = 0;
     pwdCheckRC = zisCheckUsernameAndPassword(privilegedServerName,
         request->username, request->password, &status);
-
     authResponse->type = AUTH_TYPE_RACF;
     authResponse->responseDetails.safStatus = status.safStatus;
 
@@ -2566,7 +2572,8 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
 
 static int nativeAuth(HttpService *service, HttpRequest *request, AuthResponse *authResponse){
 #ifdef __ZOWE_OS_ZOS
-  return safAuthenticate(service, request, authResponse);
+  int retValue = safAuthenticate(service, request, authResponse);
+  return retValue;
 #else
 #ifdef DEBUG_AUTH
   printf("*** ERROR *** native auth not implemented for this platform\n");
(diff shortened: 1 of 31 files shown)


h/zos.h
```diff
@@ -1157,7 +1196,7 @@ int setTaskAcee(ACEE *acee);
 
 int safVerify(int options, char *userid, char *password,
               ACEE **aceeHandle,
-              int *racfStatus, int *racfReason);
+              int *racfStatus, int *racfReason, Idta *idta);
```
ifakhrutdinov, line 1199: `safVerify` is a public function, we can't change its arguments because it'll break the users.

c/httpserver.c
```diff
@@ -2472,17 +2472,17 @@ static char *getCookieValue(HttpRequest *request, char *cookieName){
 }
 
 #ifdef __ZOWE_OS_ZOS
-static int isLowerCasePasswordAllowed(){
+int isLowerCasePasswordAllowed(){
```
ifakhrutdinov, line 2475: These functions don't seem to belong in `httpserver.h` once they've become public because they have nothing to do with the http server. They need to be moved somewhere more appropriate.

h/zos.h
```diff
@@ -1123,6 +1123,44 @@ typedef struct TIOTEntry_tag{
...
+#define IDTA_End_User_IDT       0x8000  /* Token is input from an end user */
+
+/* Idenitity Token Generation Return Code */
+#define IDTA_IDT_GEN_RC_SUCC          0
+#define IDTA_IDT_GEN_RC_UNSIGN        3
+#define IDTA_IDT_GEN_RC_ICSF_UNAVAIL  4
+#define IDTA_IDT_GEN_RC_ICSF_ERR      5
+
+
+typedef struct Idta_tag{
```
ifakhrutdinov, line 1150: Please follow the style of the source you're modifying, specifically:
* `Idta` should be `IDTA`
* No snake case in field names

h/zos.h
```diff
@@ -1123,6 +1123,44 @@ typedef struct TIOTEntry_tag{
   int       tioejfcb;      /* this evil thing is 24 high bits of SWA index, and last byte of status bits */
 } TIOTEntry;
 
+/* IDTA Version */
```
ifakhrutdinov, line 1126: Can you point to the source this comes from?

ifakhrutdinov (changes requested):

c/httpserver.c
vit-tomica: What would be a good candidate where to move the functions. Utils, zosaccounts or a new file?

h/zos.h
vit-tomica: Done

h/zos.h
vit-tomica: Fixed

h/zos.h
vit-tomica: Fixed

c/httpserver.c
ifakhrutdinov: No, I don't think so. Besides, it'd add an unnecessary dependency to the http server. Maybe `h/zos.c` since it's based on the structs there. @lchudinov what do you think?

lchudinov: `zos.c` and `zos.h` looks good to me.

c/impersonation.c
```diff
@@ -70,15 +70,15 @@ int startSafImpersonation(char *username, ACEE **newACEE) {
   }
   else {
     printf("Failed to do saf Verify\n");
-    int safStatus = safVerify(VERIFY_DELETE,NULL,NULL,newACEE,&racfStatus,&racfReason);    
+    int safStatus = safVerify(VERIFY_DELETE,NULL,NULL,newACEE,&racfStatus,&racfReason);
```
lchudinov, line 73: Please revert whitespace changes.

c/impersonation.c
```diff
@@ -70,15 +70,15 @@ int startSafImpersonation(char *username, ACEE **newACEE) {
...
   }  
   return impersonationBegan;
 }
 int endSafImpersonation(ACEE **acee) {
   int racfStatus = 0;
   int racfReason = 0;    
   int endedImpersonation;
-  int safStatus = safVerify(VERIFY_DELETE,NULL,NULL,acee,&racfStatus,&racfReason);    
+  int safStatus = safVerify(VERIFY_DELETE,NULL,NULL,acee,&racfStatus,&racfReason);
```
lchudinov, line 81: Please revert whitespace changes.

c/httpserver.c
vit-tomica: Done.

c/impersonation.c
vit-tomica: Done.

c/impersonation.c
vit-tomica: Done.

lchudinov (approved):

h/zos.h
```diff
@@ -1432,6 +1479,9 @@ bool isCallerSRB(void);
  */
 bool isCallerCrossMemory(void);
 
+int isLowerCasePasswordAllowed();
```
ifakhrutdinov, line 1482: Apologies, I just realized that `httpserver.c` is supposed to be buildable on nix/Win, so this change breaks that, because `zos.c` isn't available there.

c/httpserver.c
ifakhrutdinov: Apologies, I just realized that `httpserver.c` is supposed to be buildable on nix/Win, so this change breaks that, because `zos.c` isn't available there.

ifakhrutdinov: In that regard, `c/utils.c` may be a better place. Don't forget to add 8 char aliases under `#ifndef __LONGNAME__`, you'll see examples in `h/utils.h`.

lchudinov: @ifakhrutdinov maybe we can simply wrap the code blocks that call `isLowerCasePasswordAllowed()` into 
```C
#ifdef __ZOWE_OS_ZOS
//...
 #endif
```

vit-tomica: @ifakhrutdinov No problem. I moved it to utils.

ifakhrutdinov (approved):
lgtm

PR #261, opened 2021-12-08
jesService c1

joepun76:
Signed-off-by: joepun76 <jouphpun76@gmail.com>

### DIFF
--- a/c/datasetjson.c
+++ b/c/datasetjson.c
@@ -82,7 +82,7 @@ static int getVolserForDataset(const DatasetName *dataset, Volser *volser);
 static bool memberExists(char* dsName, DynallocMemberName daMemberName);
 
 
-static int getLreclOrRespondError(HttpResponse *response, const DatasetName *dsn, const char *ddPath) {
+int getLreclOrRespondError(HttpResponse *response, const DatasetName *dsn, const char *ddPath) {
   int lrecl = 0;
 
   FileInfo info;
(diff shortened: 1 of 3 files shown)


joepun76:
This is the zowe-common-c portion of /jes endpoint to submit job.

ifakhrutdinov:
@joepun76 I believe this should go into the zss repo. This is not common code. You may have thought that this belonged here because datasetsjson.c is in this repo, but we'd want to move it at some point in the future too.

CC: @1000TurquoisePogs

joepun76:
@ifakhrutdinov Yes, that's what I thought. Another question - I need to use below 4 functions in datasetjson.c, therefore I removed the static in front of these 4 functions.  Is it the right way? or you would suggest another method.  Thanks very much.
int getLreclOrRespondError(HttpResponse *response, const DatasetName *dsn, const char *ddPath)
bool isDatasetPathValid(const char *path)
void extractDatasetAndMemberName(const char *datasetPath,
                                        DatasetName *dsn,
                                        DatasetMemberName *memberName)
void respondWithDYNALLOCError(HttpResponse *response,
                              int rc, int sysRC, int sysRSN,
                              const DynallocDatasetName *dsn,
                              const DynallocMemberName *member,
                              const char *site)

If yes, datasetjson.c needs to be updated in this PR.

Thanks again!

ifakhrutdinov:
No, we don't need those functions to be public by exposing them via `datasetjson.c`.

I feel like, ideally, we'd want to move `datasetsjson.c` to zss, create some dsutils.c there or `zowe-common-c`, so your service and the data set services could share it. @1000TurquoisePogs perhaps this is the time to move it?

joepun76:
@ifakhrutdinov ,
As agreed by @1000TurquoisePogs to move datasetjson.c to zss, will do dsutils.c in zss and move datasetjson.c to zss and jes modules in zss. This PR will be closed. Thanks.

PR #268, opened 2022-02-02
JSON Schema

lchudinov:


build/build_cmgr_xlclang.sh
```diff
@@ -0,0 +1,137 @@
...
+
+echo "********************************************************************************"
+echo "Building configmgr..."
+
+rm -f "${COMMON}/bin/configmgr"
+
+mkdir -p "${WORKING_DIR}/tmp-configmgr" && cd "$_"
+COMMON="../.."
+
+QUICKJS="/u/zossteam/jdevlin/git2022/quickjs"
```
1000TurquoisePogs: Parameterize this

c/collections.c
```diff
@@ -1417,6 +1417,84 @@ void *qRemove(Queue *q){
...
+static void *arrayListAlloc(ArrayList *list, uint32_t size, char *location){
+  if (list->slh){
+    return (void*)SLHAlloc(list->slh,size);
+  } else{
+    return safeMalloc(size,location);
+  }
+}
+
+ArrayList *makeArrayList(){
+  ArrayList *list = (ArrayList*)safeMalloc(sizeof(ArrayList),"ArrayList");
```
ifakhrutdinov, line 1431: Do we want to have NULL checks in library code, so that the caller can decide what to do with OOM?

c/collections.c
```diff
@@ -1417,6 +1417,84 @@ void *qRemove(Queue *q){
...
+
+static void *arrayListAlloc(ArrayList *list, uint32_t size, char *location){
+  if (list->slh){
+    return (void*)SLHAlloc(list->slh,size);
+  } else{
+    return safeMalloc(size,location);
+  }
+}
+
+ArrayList *makeArrayList(){
```
ifakhrutdinov: It's better to use `void` for functions which don't have arguments in C, otherwise the caller can pass whatever and the compiler won't complain.

h/collections.h
```diff
@@ -307,8 +315,23 @@ void qInsert(Queue *q, void *newData) QueueAmode64;
  */
 void *qRemove(Queue *q) QueueAmode64;
 
-#endif
+typedef struct ArrayList_tag{
+  int capacity;
```
ifakhrutdinov, line 319: Perhaps `size_t` to support more than `2^32/2-1` elements?

c/collections.c
```diff
@@ -1417,6 +1417,84 @@ void *qRemove(Queue *q){
...
+
+void initEmbeddedArrayList(ArrayList *list,
+			   ShortLivedHeap *slh){ /* can be null for use safeMalloc - standard heap alloc */
+  list->capacity = 8;
+  list->size     = 0;
+  list->slh      = slh;
+  list->array = (void**)arrayListAlloc(list,list->capacity*sizeof(void*),"ArrayListArray");  
+}
+
+void arrayListAdd(ArrayList *list, void *thing){
```
ifakhrutdinov, line 1454: Could we use `const void*`? That way we can pass const value without the compiler producing warnings/errors.

c/collections.c
```diff
@@ -1417,6 +1417,84 @@ void *qRemove(Queue *q){
...
+    return NULL;
+  }
+}
+
+void *arrayListShallowCopy(ArrayList *source, ArrayList *target){
+  target->capacity = source->capacity;
+  target->size     = source->size;
+  target->slh      = source->slh;
+  target->array = (void**)arrayListAlloc(target,target->capacity*sizeof(void*),"ArrayListArray");  
+  memcpy(target->array,source->array,target->capacity*sizeof(void**));
```
ifakhrutdinov: We could probably copy `target->size * sizeof(void*)` only.

c/json.c
```diff
@@ -16,9 +16,11 @@
 #include <metal/stddef.h>
 #include <metal/stdio.h>
 #include <metal/stdlib.h>
+#include <metal/stdbool.h>
 #include <metal/string.h>
 #include <metal/stdarg.h>
 #include <metal/ctype.h>  
+#include <metal/stdbool.h>
```
ifakhrutdinov, line 23: Duplicate include.

c/json.c
```diff
@@ -33,8 +35,12 @@
 #include <fcntl.h>
 #include <string.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <ctype.h>  
 #include <sys/stat.h>
+#include <sys/types.h> /* JOE 1/20/22 */
+#include <stdint.h> /* JOE 1/20/22 */
+#include <stdbool.h> /* JOE 1/20/22 */
```
ifakhrutdinov: Duplicate include, and these comments are unnecessary.

c/json.c
```diff
@@ -164,7 +178,16 @@ void jsonWriteBufferInternal(jsonPrinter *p, char *text, int len) {
...
     while (bytesWritten < len) {
-#ifndef METTLE
+#if defined(__ZOWE_OS_WINDOWS)
+      int newWriteReturn = -666; /* send(p->fd,text+bytesWritten,len-bytesWritten,0); */
+      if (p->fd == _fileno(stdout)){
+        newWriteReturn = fwrite(text, 1, len, stdout);
+      } else if (p->fd == _fileno(stderr)){
+        newWriteReturn = fwrite(text, 1, len, stderr);
+      } else {
+        printf("JOE WINDOWS FAKE SOCKET WRITE bad case\n");
```
ifakhrutdinov: Should we have a proper message, like those `JSONERROR` and without personal info? :)

c/json.c
```diff
@@ -362,8 +385,8 @@ void jsonConvertAndWriteBuffer(jsonPrinter *p, char *text, size_t len,
       DUMPBUF(text, len);
       newLen = convertToUtf8(p, len, text, inputCCSID);
       if (newLen < 0) {
-        ERROR("jsonConvertAndWriteBuffer() error: newLen = %d\n",
-                newLen);
+        JSONERROR("jsonConvertAndWriteBuffer() error: newLen = %d\n",
```
ifakhrutdinov, line 385: I think `%zu` is used with `size_t`.

c/json.c
```diff
@@ -1047,6 +1102,7 @@ void freeJsonParser(JsonParser *parser) {
 
 static 
 void jsonParseFail(JsonParser *parser, char *formatString, ...) {
+  fflush(stdout);
```
ifakhrutdinov: What's the purpose of this call? There doesn't seem to be any interaction with stdout here.

c/json.c
```diff
@@ -1368,10 +1477,15 @@ JsonToken *jsonNextToken(JsonParser* parser) {
   lookahead = jsonTokenizerLookahead(tokenizer);
   tokenizer->lastTokenLineNumber = tokenizer->lineNumber;
   tokenizer->lastTokenColumnNumber = tokenizer->columnNumber;
+  /* printf("JOE jsonNextToken lookahead=0x%x '%c'\n",(int)lookahead,(char)lookahead);
```
ifakhrutdinov, line 1474: Dead code.

c/json.c
```diff
@@ -1401,12 +1515,13 @@ JsonToken *jsonNextToken(JsonParser* parser) {
...
   }
-  
+
   if (token->type == JSON_TOKEN_LINE_COMMENT || token->type == JSON_TOKEN_BLOCK_COMMENT) {
     token = jsonNextToken(parser);
   }
   jsonValidateToken(parser, token);
-
+  
+  /* printf("JOE returning next token=0x%p\n",token);fflush(stdout); */
```
ifakhrutdinov, line 1518: Ditto, and there are other lines like this below.

c/json.c
```diff
@@ -1231,6 +1287,59 @@ JsonToken *getNumberToken(JsonTokenizer *tokenizer) {
...
+    /* must have at least one number, followed by a lookahead to a non-number here */
+  } else {
+    badNumber = true;
+    int c = jsonTokenizerRead(tokenizer);
+    pos = addTokenChar(tokenizer,pos,c);
+  }
+  char *text = jsonTokenizerAlloc(tokenizer, pos + 1);
+  memcpy(text, buffer, pos);
+  if (badNumber){
+    return makeJsonToken(tokenizer,JSON_TOKEN_BAD_NUMBER,text);
```
ifakhrutdinov, line 1329: I know it's probably not that important, but we should probably look at ways to automate style/coding guideline checks (e.g. this code looks different from the existing code here). It can be done via clang-format, and it'd be easy to tell committers how things should be formatted by just asking to use clang-format with our config file.

c/json.c
ifakhrutdinov: Dead code according to Joe.

c/collections.c
JoeNemo: We are probably not consistent about this.   I would prefer safeMalloc and friends plus SLHAlloc should probably log reliably to stderr or WTO

c/collections.c
JoeNemo: You are right and I am guilty of this sin in many places.  I fixed this and a few others.  But not all

recovery.h:bool recoveryIsRouterEstablished();
recovery.h:void recoveryPop();
recovery.h:void recoveryEnableCurrentState();
recovery.h:void recoveryDisableCurrentState();
recovery.h:int recoveryRemoveRouter();
recovery.h:bool recoveryIsRouterEstablished();
recovery.h:RecoveryContext *getRecoveryContext();
recovery.h:  RecoveryContext *context = getRecoveryContext(); \
recovery.h:void recoveryPop();
recovery.h:void recoveryEnableCurrentState();
recovery.h:void recoveryDisableCurrentState();
recovery.h:extern RecoveryContext *rcvrgcxt();

h/collections.h
JoeNemo: Maybe someday.   I am scared of linear 2GB data structures.   I hope something else would be used for such a case.

c/collections.c
JoeNemo: What's an example of where this gets a warning?

c/collections.c
JoeNemo: You are correct.

c/json.c
JoeNemo: fixed in next commit.

c/json.c
JoeNemo: Does zu actually work on xlc 2.4 (not xlclang or clang)??

c/json.c
JoeNemo: fixed in next commit.

c/json.c
JoeNemo: leaving in until converted to debug traces.

c/json.c
JoeNemo: leaving in until converted to debug traces.

c/json.c
JoeNemo: I agree not in the scope of this PR

JoeNemo:
I don't know why we are hanging on a DCO here.  I signed the two commit/push-es I just did.  Confused.  I hope Sean or Irek understands.

c/json.c
```diff
@@ -1760,10 +2154,81 @@ int jsonAsBoolean(Json *json) {
 int jsonAsNumber(Json *json) {
   if (json->type == JSON_TYPE_NUMBER) {
     return json->data.number;
+  } else if (json->type == JSON_TYPE_INT64){
+    return json->data.integerValue;
```
ifakhrutdinov, line 2153: This will truncate the value if `integerValue > INT_MAX`.

c/json.c
```diff
@@ -1760,10 +2154,81 @@ int jsonAsBoolean(Json *json) {
...
+    return json->data.floatValue;
+  default:
+    return 0;
+  }
+}
+
+static int64 hashString(char *s){
+  int64 hash = 5381;
+  int i;
+  int len = strlen(s);
```
ifakhrutdinov, line 2187: Any chance the string is huge and we'd want to use `size_t`? I don't think there is any penalty, specially considering `strlen` returns `size_t`.

c/json.c
```diff
@@ -1760,10 +2154,81 @@ int jsonAsBoolean(Json *json) {
...
+
+  for (i=0; i<len; i++){
+    hash = ((hash << 5) + hash) + s[i];
+  }
+  return hash&0x7fffffffffffffffll;
+}
+
+int64_t jsonLongHash(Json *json){
+  int64_t hash = ((int64_t)json->type) << 56;
```
ifakhrutdinov, line 2196: Why do we need the type value in the 8 highest bits? Does it improve value distribution?

c/json.c
```diff
@@ -2061,13 +2556,259 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+    /* printf("merge scalar case parentKey=%s\n",(parentKey ? parentKey : "<noKey>"));
+       fflush(stdout);
+    */
+    copyJson(builder,parent,parentKey,overrides);
+    return JSON_MERGE_STATUS_SUCCESS;
+  }
+}
+
+Json *jsonMerge(ShortLivedHeap *slh, Json *overrides, Json *base, int flags, int *statusPtr){
+  JsonMerger merger;
```
ifakhrutdinov, line 2702: Just an FYI, you can do
```c
JsonMerger merger = {
  .flags = flags,
  .builder.parser.slh = slh,
};
```

The rest of the fields will be set to 0.

c/json.c
```diff
@@ -2061,13 +2556,259 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+  } else {
+    /* printf("merge scalar case parentKey=%s\n",(parentKey ? parentKey : "<noKey>"));
+       fflush(stdout);
+    */
+    copyJson(builder,parent,parentKey,overrides);
+    return JSON_MERGE_STATUS_SUCCESS;
+  }
+}
+
+Json *jsonMerge(ShortLivedHeap *slh, Json *overrides, Json *base, int flags, int *statusPtr){
```
ifakhrutdinov, line 2701: If `overrides` and/or `base` are never changed, making them `const` could be a good way to tell users they're not going to be modified.

c/json.c
```diff
@@ -2061,13 +2556,259 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+        for (i=0; i<sizeB; i++){
+          copyJson(builder,merged,NULL,jsonArrayGetItem(baseArray,i));
+        }
+        for (i=0; i<sizeO; i++){
+          copyJson(builder,merged,NULL,jsonArrayGetItem(overridesArray,i));
+        }
+        return JSON_MERGE_STATUS_SUCCESS;
+      case JSON_MERGE_FLAG_MERGE_IN_PLACE:     /* len(merge) = max(len(a),len(b)), a overrides corresponding elements of b */
+        {
+          int mergedSize = (sizeB > sizeO ? sizeB : sizeO);
```
ifakhrutdinov, line 2659: We have handy `min` and `max` macros in `h/utils.h` :-)

c/json.c
```diff
@@ -2061,13 +2556,259 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+            }
+          }
+        }
+        return JSON_MERGE_STATUS_SUCCESS;
+      case JSON_MERGE_FLAG_TAKE_BASE:          /* len(merge) = len(b) */
+        for (i=0; i<sizeB; i++){
+          copyJson(builder,merged,NULL,jsonArrayGetItem(baseArray,i));
+        }
+        return JSON_MERGE_STATUS_SUCCESS;
+      case JSON_MERGE_FLAG_TAKE_OVERRIDES:     /* len(merge) = len(a) */
```
ifakhrutdinov: Did you mean `= len(o)` in the comment? Asking because `O` is used in other places to represent the overrides.

c/json.c
JoeNemo: I hope not.  A key with size > 1 GB.  What is the world coming to?

c/json.c
JoeNemo: Yes.   If you are using V2 types, and you call jsonAsNumber before checking jsonIsFloat, or jsonIsInt64 and using their respective cases, you are "doing it wrong".   jsonAsNumber is really a legacy / backward-compatibility.   All current code aside from configmgr does NOT use V2 semantics.

c/json.c
JoeNemo: Yes.  Also, this function is more experimental than sane.  I don't like it and hope to remove it.  Eventually we need a tree-walking jsonEquals method.

c/json.c
JoeNemo: I need a "general improvements" some day.

JoeNemo:
I just commit another group of fixes and now have zero (0) compile warnings.

c/json.c
```diff
@@ -2061,13 +2545,268 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+  JsonPointerElement *element = (JsonPointerElement*)safeMalloc(sizeof(JsonPointerElement),"JsonPointerElement");
+  element->string = token;
+  element->type = type;
+  return element;
+}
+
+/* See RFC 6901 - absolute pointes only, so far */
+JsonPointer *parseJsonPointer(char *s){
+  int len = strlen(s);
+  /* char *reduced = safeMalloc(len+1); */
```
ifakhrutdinov: Dead code.

c/json.c
```diff
@@ -2061,13 +2545,268 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+  JsonPointer *jp = (JsonPointer*)safeMalloc(sizeof(JsonPointer),"JSONPointer");
+  ArrayList *list = &(jp->elements);
+  initEmbeddedArrayList(list,NULL);
+  if (len == 0){
+    return jp;
+  }
+  int pos = 1;
```
ifakhrutdinov, line 2745: Do we want to check the character at 0x00? In case `s[0] != '/'`, we'd return an error/NULL.

c/json.c
```diff
@@ -2061,13 +2545,268 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+      char c = s[i];
+      if (c < '0' || c > '9'){
+        allDigits = false;
+      }
+      if (pendingTilde){
+        if (c == '0'){
+          token[tPos++] = '~';
+        } else if (c == '1'){
+          token[tPos++] = '/';
+        } else {
```
ifakhrutdinov, line 2770: Wouldn't this be a syntax error? That is, a `~` not followed by `0` or `1`?

c/json.c
```diff
@@ -2061,13 +2545,268 @@ JsonObject *jsonObjectProperty(JsonObject *object, char *propertyName, int *stat
...
+        }
+      } else {
+        if (i == '~'){
+          pendingTilde = true;
+        } else {
+          token[tPos++] = c;
+        }
+      }
+    }
+    if (pendingTilde){
```
ifakhrutdinov, line 2783: Syntax error?

c/timeutls.c
```diff
@@ -393,7 +465,12 @@ int timeZoneDifferenceFor(int64 theTime)
   /* There has to be a better way... */
   time_t base = (time_t) theTime;
   struct tm local;
+#ifdef __ZOWE_OS_WINDOWS
+  struct tm *tempTM = localtime(&base);
```
ifakhrutdinov, line 469: Looks like `tm` can be NULL if bad data is passed in `theTime`, e.g. a date before 1970.

c/timeutls.c
```diff
@@ -353,13 +359,78 @@ uint64 sleazyToSeconds(const struct tm* time)
...
+                           struct tm *tm){
+#ifdef __ZOWE_OS_WINDOWS
+  memcpy(tm,gmtime(time),sizeof(struct tm));
+#else
+  gmtime_r(time,tm);
+#endif
+}
+
+/* A missing function in Windows -
+ * This implementation came from A MS VisStudio Support pag
```
ifakhrutdinov, line 373: Do we need any attribution? Or is it public domain (or similar)?

c/utils.c
```diff
@@ -1078,8 +1081,8 @@ void encodeBase64NoAlloc(const char buf[], int size, char result[], int *resultS
       int byte1 = data[inCursor++] & 0xff;
       int byte2 = data[inCursor++] & 0xff;
       *resPtr++ = (char)(translation[byte0 >> 2]); 
-      *resPtr++ = (char)(translation[(byte0 << 4) & 0x3f | (byte1 >> 4)]); 
-      *resPtr++ = (char)(translation[(byte1 << 2) & 0x3f | (byte2 >> 6)]); 
+      *resPtr++ = (char)(translation[((byte0 << 4) & 0x3f) | (byte1 >> 4)]);
```
ifakhrutdinov, line 1084: Was there a bug or is this just to make the precedence clear?

h/xlate.h
```diff
@@ -70,6 +70,21 @@ extern "C"
...
+  int charBuckets[0x100];
+  int rangeBuckets[0x10];
+} CharsetOracle;
+
+#define CHARSET_ORACLE_EBCDIC_FAMILY  0x10000
+#define CHARSET_ORACLE_UTF_FAMILY     0x20000
+
+CharsetOracle *makeCharsetOracle(void);
+
+void freeCharsetOracle(CharsetOracle *oracle);
```
ifakhrutdinov, line 89: Can you provide some doc? It's not clear what these functions do based on the names alone.

c/xlate.c
```diff
@@ -77,6 +80,48 @@ char* a2e(char *buffer, int len)
...
+  CharsetOracle *oracle = (CharsetOracle*)safeMalloc(sizeof(CharsetOracle),"CharsetOracle");
+  memset(oracle,0,sizeof(CharsetOracle));
+  return oracle;
+}
+
+void freeCharsetOracle(CharsetOracle *oracle){
+  safeFree((char*)oracle,sizeof(CharsetOracle));
+}
+
+void charsetOracleDigest(CharsetOracle *oracle, char *s, int len){
```
ifakhrutdinov, line 97: * `const char*` would be nice since `s` is not modified, and we could use this function with const strings.
* Are we okay with limiting the length to `INT_MAX`? We could use `size_t` as all the standard functions do.

c/zos.c
```diff
@@ -108,6 +108,7 @@ int supervisorMode(int enable){
 }
 
 int setKey(int key){
+#ifndef __ZOWE_COMP_XLCLANG  /* temp hack until Joe figures out what "&r" means */
```
ifakhrutdinov: > /* temp hack until Joe figures out what "&r" means */

From GCC:
> Means (in a particular alternative) that this operand is an earlyclobber operand, which is modified before the instruction is finished using the input operands. Therefore, this operand may not lie in a register that is used as an input operand or as part of any memory address. 

I think by default the compiler assumes that the input register aren't modified before the corresponding instruction have read the input values, which is not true in case of `SLL` which changes `%1` followed by `MODESET` reading `%1`. Because of that assumption the compiler may make some optimizations which will break the code.

We should probably change `return 8;` to actually set the key, or maybe abort, so it works in a predictable way.

c/zosfile.c
```diff
@@ -763,13 +785,16 @@ int fileInfo(const char *filename, BPXYSTAT *stats, int *returnCode, int *reason
   reasonCodePtr = reasonCode;
 #endif
 
+  printf("BPXSTA on filename\n");
```
ifakhrutdinov: Naked printfs. Can we use `zowelog` and `zowedump` since they are already used in this function?

c/bpxskt.c
```diff
@@ -107,6 +107,28 @@
 
 #endif
 
+/* xlclang and clang what prototypes, these are incomplete, but quiet the compiler */
+int BPXSOC();
+int BPXCON();
+int BPXGHN();
+int BPXSLP();
+int BPXCHR();
```
ifakhrutdinov, line 115: `BPXCHR` doesn't seem to be referenced anywhere.

h/zowetypes.h
```diff
@@ -249,6 +280,9 @@ typedef uint64_t uint64;
 #define INT64_INT(x) ((int)x)
 #define INT_INT64(x) ((int64)x)
 #define CHARPTR_INT64(x) ((int64)x)
+#define INT2PTR(x) ((void*)((int64_t)(x)))
```
ifakhrutdinov, line 283: Do we care about sign extension leading to an incorrect address? I can only think of some 16-bit address like 0xF004 which is cast to 0xFFFFFFFFFFFFF004. 31-bit addresses should not be in danger I guess.

c/collections.c
ifakhrutdinov: I believe this one would get a warning message:
```c
const char* str1 = "hello";
arrayListAdd(myList, str1);
```

h/collections.h
ifakhrutdinov: Then we could hide the struct in the c file in that case. We can then change the implementation. Up to you.

c/json.c
JoeNemo: Not a blocking issue.

c/collections.c
ifakhrutdinov: Ok, we may not be able to use this in xmem because we try not to ABEND there, and logging may also be an issue. Whereas a bad RC would let the caller know about OOM.

c/json.c
JoeNemo: Fixing...

c/json.c
ifakhrutdinov: Yes, I've used it in XLC.

c/json.c
JoeNemo: What's dead about it??

c/json.c
JoeNemo: fixing

c/json.c
ifakhrutdinov: It's commented out and never executed.

c/json.c
JoeNemo: I thought it was interpreted as "normal" text.  See RFC:


      json-pointer    = *( "/" reference-token )
      reference-token = *( unescaped / escaped )
      unescaped       = %x00-2E / %x30-7D / %x7F-10FFFF
         ; %x2F ('/') and %x7E ('~') are excluded from 'unescaped'
      escaped         = "~" ( "0" / "1" )
        ; representing '~' and '/', respectively

c/json.c
JoeNemo: See previous comment.

c/xlate.c
JoeNemo: I am really ok with 2GB at a time for now!  I would like to do a separate const-ification commit.   Zowe-common-c commits this sin in 500 places, so this is feature creep.

c/zos.c
JoeNemo: Is the use of &r really needed for the non-xlclang version.  I wrote this originally and didn't use "&r" and it worked.

c/zosfile.c
JoeNemo: fixing.

c/bpxskt.c
JoeNemo: You must be a very careful reader!!

h/zowetypes.h
JoeNemo: A short (16 bit) might have this issue, not an int.

c/json.c
JoeNemo: On second reading, you are right, and I am fixing the code and testing it.

c/timeutls.c
ifakhrutdinov: @1000TurquoisePogs what do you think? Not sure about policies in Zowe.

c/xlate.c
ifakhrutdinov: I think it'd be hard to change `int` to `size_t` in the future. So that change might be worth making right away. Not necessarily for this function, but just in general.

c/zos.c
ifakhrutdinov: It's hard to tell, I can ask the author of `&` in `&r`.

h/zowetypes.h
ifakhrutdinov: Ok, so as long as we use this for `int32_t` only it will work correctly.

c/embeddedjs.c
```diff
@@ -0,0 +1,710 @@
...
+    return JS_UNDEFINED;
+}
+
+static const JSCFunctionListEntry js_joe_funcs[] = {
+    JS_CFUNC_DEF("boop", 0, js_joe_boop ),
+};
+
+
+
+static int js_joe_init(JSContext *ctx, JSModuleDef *m)
```
ifakhrutdinov: Not sure if we should mention names in code. At least there was a rule to remove that when we were open-sourcing.

ifakhrutdinov:
@JoeNemo , @1000TurquoisePogs , just letting you know that I cannot review the new files like the parser, embedded js, config manager, tests, etc. within the given time frame. They're simply too big. If they need to be reviewed, I need much more time. Not sure how you want to proceed with those changes.

It'd be more productive to have multiple small pull-requests as things were being developed for this final PR.

c/embeddedjs.c
```diff
@@ -0,0 +1,710 @@
+
+
+/*
```
ifakhrutdinov, line 1: Does this require QuickJS? If so, should we include it as a dependency in zowe-common-c? Attribution might also be required because of the license.

h/microjq.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __ZOWE_MICROJQ__
```
ifakhrutdinov, line 1: Copyright needs to be present I believe. Not just this file.

c/json.c
ifakhrutdinov: Ok, I do think we should be just using `size_t` for sizes like the C library, maybe someone will change it in the future.

c/zos.c
ifakhrutdinov: Can we just abort until this gets implemented then?

c/json.c
```diff
@@ -372,14 +392,7 @@ void jsonConvertAndWriteBuffer(jsonPrinter *p, char *text, size_t len,
...
     if (escape) {
-      size_t bytesWritten;
-
-      bytesWritten = writeBufferWithEscaping(p, len, text);
-      if (bytesWritten < 0) {
-        ERROR("jsonConvertAndWriteBuffer() error: bytesWritten = %d\n",
-                bytesWritten);
-        return;
-      }
+      writeBufferWithEscaping(p, len, text);
```
ifakhrutdinov: Should we keep the check? Looks like the call can fail. The only fix needed would be changing `size_t` to `ssize_t`.

c/zos.c
```diff
@@ -275,7 +278,7 @@ OTCB *getOTCB(void) {
 
 ASCB *getASCB(void) {
   int *mem = (int*)0;
-  return (ASCB*)(mem[CURRENT_ASCB/sizeof(int)]&0x7FFFFFFF);
+  return (ASCB*)INT2PTR(mem[CURRENT_ASCB/4]);
```
ifakhrutdinov: Can the ASCB address ever have the highest bit set to 1? Asking because the old code used to and with `0x7fffffff`. Perhaps, that was there for a reason.

c/zos.c
```diff
@@ -1029,12 +1058,14 @@ static int safVerifyInternal(int options,
...
     if (safTrace){
-      fprintf(safTraceFile,"countedUserid %0.8X' password %0.8X' passphrase %0.8X'\n",
+      fprintf(safTraceFile,"countedUserid 0x%p' password 0x%p' passphrase 0x%p'\n",
               countedUserid, countedPassword, countedPassphrase);
-      fprintf(safTraceFile,"about to go call saf wrapper at %0.8X' verify block at %0.8X' acee %0.8X'\n",safWrapper,verifyRequest,
+      fprintf(safTraceFile,"about to go call saf wrapper at 0x%p' verify block at 0x%p' acee 0x%p'\n",safWrapper,verifyRequest,
               (aceeHandle != NULL ? *aceeHandle : NULL));
       dumpbuffer((char*)safWrapper,sizeof(safp));
-      fprintf(safTraceFile,"verify:\n",0);
+#ifndef METTLE
```
ifakhrutdinov: Why do we need to shield this calls against Metal and not the `fprintf` calls above?

c/zosfile.c
```diff
@@ -763,6 +785,7 @@ int fileInfo(const char *filename, BPXYSTAT *stats, int *returnCode, int *reason
   reasonCodePtr = reasonCode;
 #endif
 
+  dumpbuffer(filename,strlen(filename));
```
ifakhrutdinov: Please remove this naked `dumpbuffer` call.

c/zos.c
JoeNemo: I will test a fix that gets rid of the &.    I don't know who made this change but will find out.

c/embeddedjs.c
JoeNemo: Sean has a list of zowe dependencies that shows up somewhere for license compliance.

c/xlate.c
ifakhrutdinov: Having discussed this with Joe, we'll do a general clean up later.

c/json.c
JoeNemo: Fixed in source.

c/zosfile.c
JoeNemo: Fixed

c/zos.c
JoeNemo: fixed in code

c/zosfile.c
JoeNemo: done!

c/zos.c
ifakhrutdinov: Should not matter for ASCB according to experts.

c/json.c
ifakhrutdinov: Discussed this with Joe. General clean up will be done later.

c/collections.c
ifakhrutdinov: General hardening of the common data structures will be done separately.

c/zos.c
ifakhrutdinov: Joe fixed this by moving the left shift to the C code and removing `&`.

c/embeddedjs.c
ifakhrutdinov: Fixed.

h/microjq.h
ifakhrutdinov: Fixed.

c/configmgr.c
```diff
@@ -0,0 +1,1131 @@
...
+      }
+    } else if (jsonIsObject(value)){
+      JsonObject *object = jsonAsObject(value);
+      value = jsonObjectGetPropertyValue(object,element->string);
+      if (value == NULL){
+        *errorReason = JSON_POINTER_TOO_DEEP;
+        return NULL;
+      }
+    } else {
+      printf("SSS\n");fflush(stdout);
```
lchudinov: Bare `printf`. please hide it under a trace level

lchudinov:
The new changes look good to me, as far as I can tell by browsing them. I didn't test the latest changes.

@JoeNemo Please don't forget to create a follow-up issue saying that `import` statement has to be banned in embedded JS.

1000TurquoisePogs (approved):
Merging after all review complete.

c/timeutls.c
1000TurquoisePogs: From what I know, these things are public domain.

PR #270, opened 2022-02-09
Support APPL keyword  for IDT token call

vit-tomica:
Signed-off-by: Vit Tomica <vit.tomica@broadcom.com>

### DIFF
--- a/c/zos.c
+++ b/c/zos.c
@@ -903,6 +903,23 @@ int safVerify6(int options, char *userid, char *password,
                             idta));
 }
 
+int safVerify7(int options, char *userid, char *password,
+              ACEE **aceeHandle, char *appl,
+              int *racfStatus, int *racfReason, IDTA *idta) {
+  return (safVerifyInternal(options,
+                            userid,
+                            password,
+                            NULL,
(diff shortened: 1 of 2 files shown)


vit-tomica:
Hello @ifakhrutdinov @lchudinov, can you take a look at this PR, please?

lchudinov (approved):
Looks good to me

ifakhrutdinov:
@vit-tomica, @lchudinov  shouldn't we aim the `v2.x/staging` branch?

ifakhrutdinov:
Additionally, at some point in the future we may want to stop creating new versions of this function like this (it's already v7!) and look at some other solutions. Maybe something like the builder pattern?

vit-tomica:
@ifakhrutdinov I don't know if we should aim for v2.x/staging branch. Who could know that?

ifakhrutdinov:
@1000TurquoisePogs, do we merge such features to v2 only?

anton-brezina:
Concerning the conversation above, let me emphasize, that if this does go into v2 only we'll have an ongoing defect in v1 in SAF IDT.

PR #272, opened 2022-02-16
Convert config to env

lchudinov:
This PR implements `env` command for `configmgr` that converts the config into the list of environment variables.


JoeNemo (commented):
Looks good to me

PR #285, opened 2022-04-01
Feature/configmgr integration

JoeNemo:
This branch supports any changes needed to get the configuration manager integrated to zss-install-packaging and zss


c/bpxskt.c
```diff
@@ -381,6 +394,16 @@ Socket *tcpClient(SocketAddress *socketAddress,
   return tcpClient2(socketAddress,-1,returnCode,reasonCode);
 }
 
+Socket *makePipeBasedSyntheticSocket(int protocol, int inputFD, int outputFD){
+  Socket *socket = (Socket*)safeMalloc(sizeof(Socket),"Socket");
+  socket->sd = inputFD;
+  socket->pipeOutputSD = outputFD;
+  sprintf(socket->debugName,"PIPE(%d,%d)",inputFD,outputFD);
```
ifakhrutdinov: I'd use `snprintf` in case someone else comes and adds additional debug info overflowing this buffer.

c/charsets.c
```diff
@@ -60,37 +101,67 @@
...
+  } else {
+    wideTemp = safeMalloc(2*inputLength,"WideTemp");
+    needToFreeTemp = true;
+  }
+  int wideCharCount = 0;
+  int required = MultiByteToWideChar(inputCCSID,0,input,inputLength,NULL,0);
+  int status = MultiByteToWideChar(inputCCSID,0,input,inputLength,(LPWSTR)wideTemp,inputLength); /* 2nd length is count of wchar */
+  if (status == 0){
+    *reasonCode = GetLastError();
+    printf("win multiToWide reason = %d, inputCCSID=%d wideTemp=0x%p, inLen=%d req=%d\n",
```
ifakhrutdinov: Naked printf or are we not worried about that on Win?

c/charsets.c
```diff
@@ -60,37 +101,67 @@
...
+  }
+  
+  status = WideCharToMultiByte(outputCCSID,0,(LPWSTR)wideTemp,wideCharCount,outputBuffer,outputLength,NULL,NULL);
+  if (needToFreeTemp){
+    safeFree(wideTemp,2*inputLength);
   }
-  int status = MultiByteToWideChar(inputCCSID,0,input,inputLength,outputBuffer,outputLength);
   if (status == 0){
     *reasonCode = GetLastError();
+    printf("win wideToMulti reason = %d\n",*reasonCode);
```
ifakhrutdinov: Naked printf.

c/charsets.c
```diff
@@ -60,37 +101,67 @@
...
+  }
+  int wideCharCount = 0;
+  int required = MultiByteToWideChar(inputCCSID,0,input,inputLength,NULL,0);
+  int status = MultiByteToWideChar(inputCCSID,0,input,inputLength,(LPWSTR)wideTemp,inputLength); /* 2nd length is count of wchar */
+  if (status == 0){
+    *reasonCode = GetLastError();
+    printf("win multiToWide reason = %d, inputCCSID=%d wideTemp=0x%p, inLen=%d req=%d\n",
+           *reasonCode,inputCCSID,wideTemp,inputLength,required);
+    if (needToFreeTemp){
+      safeFree(wideTemp,2*inputLength);
```
ifakhrutdinov: I'd define a variable instead of having `2*inputLength` in several places. Something like `wideTempLength`.

c/crypto.c
```diff
@@ -111,7 +111,7 @@ int digestContextInit(DigestContext *context, CryptoDigestType digestType){
     wprintf(L"**** Error 0x%x returned by BCryptGetProperty\n", status);
     return CRYPTO_GENERAL_FAILURE;
   }
-  printf("cbHashObject=%d cbData=%d\n",cbHashObject,cbData);
+  printf("cbHashObject=%d cbData=%d\n",(int)cbHashObject,(int)cbData);
```
ifakhrutdinov, line 114: ~~I believe using `%ld` would be the correct way, otherwise the values are just truncated.~~ Doesn't work on other platforms, Joe's working on this.

c/zosfile.c
```diff
@@ -614,8 +620,15 @@ int fileCopy(const char *existingFileName, const char *newFileName, int *retCode
     return -1;
   }
 
+
   if (ccsid != CCSID_UNTAGGED) {
-    status = fileChangeTagPure(newFileName, &returnCode, &reasonCode, ccsid, isPureText);
+    if (shouldConvert && (ccsid != existingCCSID)){
+      *retCode = ERRNO_CCSID_MISMATCH;
+      *resCode = 0;
+      return -1;
```
ifakhrutdinov, line 629: I think we should also close the open files. Same when any other function fails.

h/unixfile.h
```diff
@@ -481,12 +485,20 @@ int setFileInfoCCSID(int ccsid);
 #endif
 int fileInfoCCSID(const FileInfo *info);
 int fileInfoUnixCreationTime(const FileInfo *info);
+int fileInfoUnixModificationTime(const FileInfo *info);
+/**
```
ifakhrutdinov: Do ween need this empty comment block?

h/le.h
```diff
@@ -74,7 +75,13 @@ typedef struct CAA_tag{
   Addr31 ceecaaecov;     /* 0x234 routine vector, important */
   int    ceecaactofsv;   /* 0x238 what is this? */
   Addr31 ceecaatrtspcae; /* 0x23C Another routine vector ?? */
-  char   reserved240[24];
+  char   reserved240[16];
```
ifakhrutdinov, line 79: I believe this will break existing ZIS plugins. Many of them link `zowelog` and related functionality statically and thus rely on the same offsets for these fields.

jwt/rscrypto/rs_icsfp11.h
```diff
@@ -103,10 +108,13 @@ int rs_icsfp11_getRandomBytes(const ICSFP11_HANDLE_T *token_handle,
                               unsigned char *randbuf, int *randbuf_len,
                               int *out_rc, int *out_reason);
 
+/* JOE - this is a duplicate of a later header */
+/*
```
ifakhrutdinov, line 112: Should we just remove this?

c/httpserver.c
```diff
@@ -703,9 +710,6 @@ static int readMachineAdvance(WSReadMachine *m, char *data, WSSession *wsSession
   int shouldClose = FALSE;
   m->trace = TRUE;
   while (available > 0){
-    if (loopCount++ > loopCount){
```
ifakhrutdinov, line 706: It was an interesting piece of code.

c/httpserver.c
```diff
@@ -1388,22 +1392,25 @@ static int initSessionTokenKey(SessionTokenKey *key) {
 
 #else
 
-#error Session token key generation has been implemented for z/OS only
+  printf("*** WARNING *** low quality random number in use for SessionToken \n");
```
ifakhrutdinov, line 1396: `zowelog` should be used.

c/httpserver.c
```diff
@@ -1431,6 +1442,13 @@ static int encodeSessionToken(ShortLivedHeap *slh,
   *result = encodedTokenText;
   return 0;
 
+#elif defined(__ZOWE_OS_WINDOWS)
+
+  printf("*WARNING* - Trivial Cipher in use\n");
```
ifakhrutdinov, line 1448: Same, `zowelog`. And there are more below.

c/httpserver.c
```diff
@@ -2476,8 +2524,102 @@ int isLowerCasePasswordAllowed(){
...
+
+/* because of the evil of the backwards mainframe default for hashing in ICSF
+   we need an open-source replacement of ICSF for single-user mode.
+
+   Poughkeepsie, this is why we can't have nice things on ZOS.
+   */
+static char *md5It(char *s, int len, char *hash){
+  ICSFDigest digest;
+  memset(hash,0,16);
+  icsfDigestInit(&digest,ICSF_DIGEST_MD5);
```
ifakhrutdinov, line 2545: Will anyone complain about MD5? I think security scanners pick it up now. It seems that we only store this hash in memory of ZSS where the clear text password is too, but if we could use some other hash we'd avoid people complaining about MD5.

c/httpserver.c
```diff
@@ -2476,8 +2524,102 @@ int isLowerCasePasswordAllowed(){
...
 #ifdef __ZOWE_OS_ZOS
+
+/* because of the evil of the backwards mainframe default for hashing in ICSF
+   we need an open-source replacement of ICSF for single-user mode.
+
+   Poughkeepsie, this is why we can't have nice things on ZOS.
+   */
+static char *md5It(char *s, int len, char *hash){
+  ICSFDigest digest;
+  memset(hash,0,16);
```
ifakhrutdinov: Can we use `#define` instead of magic numbers? This also applies to number 6 as the offset after `Basic ` below

c/httpserver.c
```diff
@@ -2476,8 +2524,102 @@ int isLowerCasePasswordAllowed(){
...
+  char asciiHeaderBuffer[100];
+  char *asciiHeader = NULL;
+  if (authenticationHeader){
+    asciiHeader = authenticationHeader->value;
+  } else{
+    if (strlen(request->username) > 32 || strlen(request->password) > 64){
+      printf("username or password too long\n");
+      return 0;
+    }
+    sprintf(cleartextBuffer,"Basic %s:%s",request->username,request->password);
```
ifakhrutdinov: This will cause a buffer overflow if `username` is 32 characters and `password` is 64. With `#define`d user name and password this could be avoided.

c/httpserver.c
```diff
@@ -1539,6 +1567,26 @@ HttpServer *makeHttpServerInner(STCBase *base,
...
+  /* pipe tracking and the arguments to makeSExt() */
+  SocketExtension *pipeExtension = makeSocketExtension(pipeSocket,
+                                                       server->slh,
+                                                       FALSE, /* allocateInSLH */
+                                                       server,
+                                                       READ_BUFFER_SIZE);
+  pipeSocket->userData = pipeExtension;
+  stcRegisterSocketExtension(server->base, pipeExtension, STC_MODULE_JEDHTTP);
+  
+  server->singleUserMode = true; /* this should be somewhere else some day */
```
ifakhrutdinov, line 1585: Could you add a "TODO" to this comment so it can be easily seen.

c/httpserver.c
```diff
@@ -2476,8 +2524,102 @@ int isLowerCasePasswordAllowed(){
...
+    asciiHeader[encodedSize] = 0;
+    if (singleUserTrace >= 1){
+      printf("JOE: Built ascii header\n");
+      dumpbuffer(asciiHeader,encodedSize+1);
+    }
+  }
+  char md5Hash[16];
+  md5It(asciiHeader,strlen(asciiHeader),md5Hash);
+  strupcase(request->username); /* upfold username */
+  char *safUser = acee->aceeuser;
```
ifakhrutdinov, line 2615: Should this be also compared with the user from the request? The request would be reject if the users don't match. Otherwise we'd be able to login under some other user and then use impersonation to, for example, access their files.

c/httpserver.c
```diff
@@ -5473,7 +5702,9 @@ static int httpHandleTCP(STCBase *base,
                                peerSocket->sd,
                                true);
         if ((0 != rc) || (NULL == peerSocket->tlsSocket)) {
-          zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG, "httpserver failed to negotiate TLS with peer; closing socket\n");
+          zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG, 
+		  "httpserver failed to negotiate TLS with peer; closing socket, tlsInit.rc=%d skt.dbg.name=%s\n",
```
ifakhrutdinov, line 5743: Just checking.. security wise could we expose anything by printing the RC here since it comes from gskit?

c/httpserver.c
```diff
@@ -5503,61 +5734,208 @@ static int httpHandleTCP(STCBase *base,
...
+						 Socket *demuxSocket, 
+						 int uniqueID){
+  printf("getOrMake=%d\n",uniqueID);
+  PipeSocketEntry *entry = htIntGet(server->syntheticPipeSockets,uniqueID);
+  printf("ht result = 0x%p\n",entry);
+
+  if (entry == NULL){
+    int fdsToHttp[2] ={ -1, -1};  /* readEnd, writeEnd */
+    int fdsFromHttp[2] ={ -1, -1};
+    pipe (fdsToHttp);
```
ifakhrutdinov: It seems a `pipe` call can fail. Should we check and log that?

ifakhrutdinov:
FYI, I haven't reviewed configmgr, jsonschema and embeddedjs yet.

c/httpserver.c
ifakhrutdinov: This is done intentionally according to the author.

c/httpserver.c
```diff
@@ -5503,61 +5746,218 @@ static int httpHandleTCP(STCBase *base,
...
+      int pipeErrno = 0;
+      PipeSocketEntry *entry = getOrMakePipeSocketEntry(server,module,socket,uniqueID,&pipeErrno);
+      if (entry){
+	Socket *forwardingSocket = entry->muxSocket;
+	
+	zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG2, "entry=0x%p forwardingSocket=0x%p %\n",
+		entry,forwardingSocket,(forwardingSocket ? forwardingSocket->debugName : ""));
+	socketWrite(forwardingSocket,readBuffer+headerLength,fragment->payloadLength,&returnCode,&reasonCode);
+	if (returnCode){
+	  printf("Pipe-forwarding tunneled Packet failed ret=%d, reason=0x%x\n",returnCode,reasonCode);
```
ifakhrutdinov: Could you use `zowelog` for all the printfs? This could be `ZOWE_LOG_SEVERE`.

Otherwise, it'll be hard to change formatting of all the messages, e.g. add a timestamp.

c/configmgr.c
```diff
@@ -942,12 +1207,250 @@ static void convertJsonArrayToEnv(FILE *out, const char *path, JsonArray *array)
...
+                                                          (EJSForeignFunction*)loadConfigurationsWrapper);
+  ejsAddMethodArg(ejs,loadConfiguration,"configName",EJS_NATIVE_TYPE_CONST_STRING);
+
+  EJSNativeMethod *getConfigData = ejsMakeNativeMethod(ejs,configmgr,"getConfigData",
+                                                       EJS_NATIVE_TYPE_JSON,
+                                                       (EJSForeignFunction*)getConfigDataWrapper);
+  ejsAddMethodArg(ejs,getConfigData,"configName",EJS_NATIVE_TYPE_CONST_STRING);
+
+  EJSNativeMethod *writeYAML = ejsMakeNativeMethod(ejs,configmgr,"writeYAML",
+						   EJS_NATIVE_TYPE_JSON,
```
ifakhrutdinov: Tabs

c/httpserver.c
```diff
@@ -5503,61 +5747,218 @@ static int httpHandleTCP(STCBase *base,
...
+  SocketExtension *extension = (SocketExtension*)socket->userData;
+  switch (socket->protocol){
+  case IPPROTO_SYNTHETIC_PIPE_DEMULTIPLEXER:  
+    {
+      /* every pipe chunk for read ready must be > 4 in size */
+      char *readBuffer = extension->readBuffer;
+      
+      int bytesRead = socketRead(socket,readBuffer,8,&returnCode,&reasonCode);
+      if (returnCode){
+	printf("pipe IO error 1 ret=%d reason=0x%x\n",returnCode,reasonCode);
```
ifakhrutdinov: Tabs.

c/stcbase.c
```diff
@@ -403,13 +413,14 @@ void stcBaseTerm(STCBase *stcBase) {
...
-                             int  (*tcpHandler)(STCBase *base, STCModule *module, Socket *socket),
-                             int  (*udpHandler)(STCBase *base, STCModule *module, Socket *socket),
-                             int  (*workElementHandler)(STCBase *base, STCModule *stcModule, WorkElementPrefix *prefix),
-                             int  (*backgroundHandler)(struct STCBase_tag *base, struct STCModule_tag *stcModule, int selectStatus))
+STCModule* stcRegisterModule2(STCBase *base,
+                              int moduleID,
+                              void *moduleData,
+                              int  (*tcpHandler)(STCBase *base, STCModule *module, Socket *socket),
+                              int  (*udpHandler)(STCBase *base, STCModule *module, Socket *socket),
+                              int  (*pipeHandler)(STCBase *base, STCModule *module, Socket *socket),
```
ifakhrutdinov, line 421: `pipHandler` doesn't seem to be used anywhere

c/httpserver.c
```diff
@@ -5503,61 +5747,218 @@ static int httpHandleTCP(STCBase *base,
...
 
-      zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "handle read JEDHTTP convo\n");
-      /* if successful, these methods enqueue work */
-      if (conversation->wsSession){
-        doWSReadWork(conversation,READ_BUFFER_SIZE);
-      } else {
-        doHttpReadWork(conversation,READ_BUFFER_SIZE);
+/* this method *only* runs in the STCBase main thread, putting and effective mutex 
+   on the pipes that talk up/down to process that owns this httpserver */
+static int httpHandlePipe(STCBase *base,
```
ifakhrutdinov, line 5861: Below is an example with 2 connections. Did I understand the logic correctly?
```
Connection 1 pipes:
* toHttp:   pipe(5,6)
* fromHttp: pipe(7,8)

Connection 2 pipes:
* toHttp:   pipe(9,10)
* fromHttp: pipe(11,12)

                              +
          tcp(http)           |              http
                              |
                              |
                              |
                              |
                  pipe_dmux   |        pipe_mux             pipe_hndl    http_handler
stdin  ========>   sd(0) ---+------> poSd(06) >==========>   sd(05)  ---> httpReq
stdout ========< poSd(1) <-+|-------   sd(07) <==========< poSd(08)  <--- httpRes
                           || |  
                           || |  
                           |+------> poSd(10) >==========>   sd(09)  ---> httpReq
                           +--------   sd(11) <==========< poSd(12)  <--- httpRes
                              |
                              |
                              |
                              |
                              +

```

I have a few questions:
* When are the use cases for this feature?
* Why does the incoming data have the TCP header? Can you point me to some context?
* How is a connection terminated and when are its resources freed?
* Related to the previous item, is it possible that the ID from the TCP header isn't actually identifying the same client because the original one terminated and some new one connected. And what if that happens in the middle of request/response?

c/httpserver.c
```diff
@@ -5503,61 +5747,218 @@ static int httpHandleTCP(STCBase *base,
...
+	  TCPFragment fragmentStorage;
+	  TCPFragment *fragment = &fragmentStorage;
+	  memset(fragment,0,sizeof(TCPFragment));
+	  fragment->magic = TCP_FRAGMENT_MAGIC;
+	  fragment->headerLength = sizeof(TCPFragment);
+	  fragment->type = TCP_FRAGMENT_TO_REMOTE;
+	  fragment->remoteClientID = entry->uniqueID;
+	  fragment->payloadLength = bytesRead;
+	  socketWrite(demuxSocket,(char*)fragment,sizeof(TCPFragment),&returnCode,&reasonCode);
+	  FILE *tempOut = fdopen(demuxSocket->pipeOutputSD,"w");
```
ifakhrutdinov, line 5963: Can you leave a comment about the purpose of this temp file? Also, should we close it?

c/httpserver.c
```diff
@@ -5503,61 +5747,218 @@ static int httpHandleTCP(STCBase *base,
...
+      if (entry == NULL){
+	printf("no PipeSocketEntry found for socket %s\n",socket->debugName);
+      } else{
+	char *readBuffer = extension->readBuffer;
+	Socket *demuxSocket = entry->demuxSocket;
+	Socket *muxSocket = socket;  /* the one that is readable */
+	int bytesRead = socketRead(muxSocket,readBuffer,READ_BUFFER_SIZE,&returnCode,&reasonCode);
+	zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG2, "MUX case: 0x%x bytes back from handler, ret=%d reason=0x%x\n",bytesRead,returnCode,reasonCode);
+	if (bytesRead > 0){
+	  int dumpLen = bytesRead > 0x400 ? 400 : bytesRead;
```
ifakhrutdinov: Did you mean to use 0x400 in both places?

c/httpserver.c
```diff
@@ -5503,61 +5747,218 @@ static int httpHandleTCP(STCBase *base,
...
+
+  if (entry == NULL){
+    int fdsToHttp[2] ={ -1, -1};  /* readEnd, writeEnd */
+    int fdsFromHttp[2] ={ -1, -1};
+    if (pipe(fdsToHttp)){
+      *pipeError = errno;
+      return NULL;
+    }
+    if (pipe(fdsFromHttp)){
+      *pipeError = errno;
```
ifakhrutdinov, line 5810: Should `fdsToHttp` be closed?

c/bpxskt.c
```diff
@@ -281,7 +281,8 @@ Socket *tcpClient3(SocketAddress *socketAddress,
       /* EINPROGRESS is the expected return code here but we are just being careful by checking
          for EWOULDBLOCK as well
       */
-      if (returnValue != 0){
+      if (returnValue){
```
ifakhrutdinov, line 284: This change doesn't do anything.

c/configmgr.c
```diff
@@ -103,14 +103,26 @@ typedef int64_t ssize_t;
 
     clang++ -c ../platform/windows/cppregex.cpp ../platform/windows/winregex.cpp
 
+<<<<<<< HEAD
```
ifakhrutdinov: Unresolved conflict?

c/configmgr.c
```diff
@@ -633,7 +649,9 @@ static Json *readJson(ConfigManager *mgr, ConfigPathElement *pathElement){
 #endif
       trace(mgr,INFO,"WARNING, yaml read failed, errorBuffer='%s'\n",errorBuffer);
       return NULL;
-    }  
+    }
+  } else if (pathElement->flags & CONFIG_PATH_OMVS_FILE){
```
ifakhrutdinov: This condition is the same as in the first `if`, so it will never be reached. Did you mean some other flag?

c/httpserver.c
JoeNemo: Maybe, but in single user mode it will end naturally when zss ends.   It's not a system-wide resource, it's  process-level resource that auto-cleans.

c/bpxskt.c
JoeNemo: sorry!  I had some horrible merge/rebase issues and it came as part of that.

c/httpserver.c
JoeNemo: This is the basis of single user Zowe/ZSS.   It does not have a TCP Header.    At no place can I pick this up, ie, no code is written below the level of normal socket or pipe code.   The data structure in question is below and associates the tunnelled conversation to the "remotClientID" which is as unique as the tcp connections made to the far-side (non-ZSS) side of the tunnel.    I have further work to go, there is no "production" code that uses this.   I am trying to get to single user mode being robust enough for demos, and make more fixes.    

typedef struct TCPFragment{
  uint32_t magic;        /* D0BED0BE */
  uint16_t headerLength;
  uint8_t  flags;
  uint8_t  reserved;
  uint32_t type;
  uint32_t remoteClientID; /* proxy for client socket at far end of tunnel */
  /* offset 10 */
  uint32_t payloadLength;
  char     optionalHeaderData[];
} TCPFragment;

c/httpserver.c
JoeNemo: oh my!! wow!! that's a bug!!

JoeNemo: But of course, only in the tracing!

c/configmgr.c
JoeNemo: This is the partially done PARMLIB code.   I stopped committing so we could finish this PR. Sorry

c/httpserver.c
JoeNemo: in next commit

c/httpserver.c
ifakhrutdinov: But who generates those headers? At this point I don't understand the full lifecycle of this, i.e. who (what kind of application) connects to ZSS in this mode, how the data is formed on the client end.

And I guess I got confused because it's called "TCP fragment", so I thought it was using the TCP protocol in some form.

c/stcbase.c
ifakhrutdinov: @JoeNemo Was that intended?

c/httpserver.c
ifakhrutdinov: Thanks, though there are more in the new code.

c/bpxskt.c
ifakhrutdinov: Will you fix/revert that?

jordanfilteau1995:
I think it's ok. I didn't see any glaring issues when going over the files with Joe. This looks like changes that will really help the Zowe installation experience. Joe/Irek, let me know if you need anything else from me. Thanks for your hard work on this.

c/httpserver.c
ifakhrutdinov: Talked to Joe, my questions have been answered.

c/httpserver.c
ifakhrutdinov: According to Joe this is WIP and will change in a next PR.

ifakhrutdinov:
LGTM, @struga0258 please test the security well because there have been changes in the user password handling code. Let me know if you need details.

ifakhrutdinov (approved):

PR #289, opened 2022-05-11
Support SAF IDT with APPL

achmelo:
This is a cherrypick from https://github.com/zowe/zowe-common-c/pull/270


JoeNemo (approved):
This API changes seems fine to me.

PR #244, opened 2021-09-24
API for Dataset/Dataset Member Allocation

timgerstel:
This pull request implements the following feature:
- API which enables the allocation of datasets and dataset members.
- Add 'PUT' method to /datasetContents, accepts JSON body to specify dataset parameters
- Sample request body:
 ```
{
    "ndisp": "CATALOG",
    "status": "NEW",
    "dsorg": "PS",
    "blksz": "27998",
    "lrecl": "84",
    "recfm": "V",
    "close": "true"
}
```

PR #292, opened 2022-05-16
API for Dataset/Dataset Member Allocation

aditya-ranshinge:
API for Dataset/Dataset Member Allocation was implemented in https://github.com/zowe/zowe-common-c/pull/244  this PR , which was reviewed by Leonty. 
This PR is continuation of above mentioned PR where Allocation Unit, Average Record Unit, Primary Space, Secondary Space, Directory Blocks, Dataset Name Type attributes added for dataset creation.

This pull request implements the following feature:
- API which enables the allocation of datasets and dataset members.
- Add 'PUT' method to /datasetContents, accepts JSON body to specify dataset parameters


h/dynalloc.h
```diff
@@ -681,9 +708,37 @@ int DeallocDDName(char *ddname);
...
+#define DISP_CATLG 0x02
+#define DISP_DELETE 0x04
+#define DISP_KEEP 0x08
+
+#define DALSYSOU_DEFAULT 0x08
+
+#define SPIN_UNALLOC 0x80
+#define SPIN_ENDJOB 0x40
+
+#define BYTE_LENGTH 8
```
ifakhrutdinov: These don't belong in this header. You can keep them private in the source file. In fact, remove anything new from the header which isn't going to be part of the API.

c/dynalloc.c
```diff
@@ -834,6 +854,65 @@ int dynallocDataset(DynallocInputParms *inputParms, int *reasonCode) {
 
 }
 
+int dynallocNewDataset(int *reasonCode, DynallocNewTextUnit *setTextUnits, int TextUnitsSize) {
```
ifakhrutdinov: Arguments (in your case `TextUnitsSize`) should start with a lower case letter.

c/dynalloc.c
```diff
@@ -834,6 +854,65 @@ int dynallocDataset(DynallocInputParms *inputParms, int *reasonCode) {
...
+  ALLOC_STRUCT31(
+    STRUCT31_NAME(below2G),
+    STRUCT31_FIELDS(
+      DynallocParms parms;
+      TextUnit ** __ptr32 textUnits;
+    )
+  );
+
+  below2G->textUnits = (TextUnit **)safeMalloc31(sizeof(TextUnit*) * TextUnitsSize, "Text units array");
```
ifakhrutdinov: Check for `NULL` and return -1 if it is.

c/dynalloc.c
```diff
@@ -834,6 +854,65 @@ int dynallocDataset(DynallocInputParms *inputParms, int *reasonCode) {
...
+
+  dynallocParmsSetTextUnits(parms, (TextUnit * __ptr32 *)below2G->textUnits, TextUnitsSize);
+
+  int rc;
+
+  do {
+    rc = -1;
+    for (int i = 0; i < TextUnitsSize; i++) { 
+      if (setTextUnits[i].type == TEXT_UNIT_STRING) {
+        below2G->textUnits[i] = createSimpleTextUnit2(setTextUnits[i].key, setTextUnits[i].data.string, setTextUnits[i].size);
```
ifakhrutdinov: Check these "create" functions for `NULL` and break out of this loop. This is how it's been done in `dynallocDataset`.

h/dynalloc.h
```diff
@@ -696,11 +751,24 @@ typedef struct DynallocInputParms_tag {
   char reserved[3];
 } DynallocInputParms;
 
+
+typedef struct DynallocNewTextUnit_tag {
```
ifakhrutdinov: I don't know how I feel about this. It seems like we're adding another layer on top of DYNALLOC, which doesn't really simplify the use of DYNALLOC. How does this help use DYNALLOC? Would asking people to pass the original text units be basically the same?

Also, why is this "New" text units? Can they only be used to allocate data sets?

h/dynalloc.h
```diff
@@ -681,9 +708,37 @@ int DeallocDDName(char *ddname);
...
+#define CLASS_WRITER_SIZE 8
+#define TOTAL_TEXT_UNITS 23
+#define LEN_THREE_BYTES 3
+#define LEN_ONE_BYTE    1
+
+#define TEXT_UNIT_STRING 1
+#define TEXT_UNIT_BOOLEAN 2
+#define TEXT_UNIT_CHARINT 3
+#define TEXT_UNIT_NULL 4
+#define JSON_TYPE_ERROR 666
```
ifakhrutdinov: What do you need `JSON_TYPE_ERROR` for?

ifakhrutdinov (commented):
I'm not sure this is the right approach. The solution is another layer on top of DYNALLOC, with almost the same level of complexity. How is this better than, for example, using DYNALLOC's text units directly or using https://www.ibm.com/docs/en/zos/2.4.0?topic=functions-dynalloc-allocate-data-set ?

@JoeNemo what do you think?

h/dynalloc.h
aditya-ranshinge: Resolved

c/dynalloc.c
aditya-ranshinge: Resolved

c/dynalloc.c
aditya-ranshinge: Resolved

c/dynalloc.c
aditya-ranshinge: Resolved

h/dynalloc.h
aditya-ranshinge: Existing function arguments in dynalloc.c are accepting parms in separate structure `DynallocInputParms` so to keep consistency in source code I have just renamed structure with `DynallocNewInputParms` instead of passing text units.  
Let me know how you feel about it.

h/dynalloc.h
aditya-ranshinge: Resolved

h/dynalloc.h
```diff
@@ -696,11 +741,23 @@ typedef struct DynallocInputParms_tag {
...
+    char *stringValue;
+  } data;
+} DynallocNewInputParms;
+
 #pragma map(dynallocDataset, "DYNAUALC")
 #pragma map(dynallocDatasetMember, "DYNAUALM")
 #pragma map(unallocDataset, "DYNADALC")
 
 int dynallocDataset(DynallocInputParms *inputParms, int *reasonCode);
+int dynallocNewDataset(DynallocNewInputParms *setInputParms, int TextUnitsSize, int *reasonCode);
```
ifakhrutdinov: On the one hand you've tried abstracting away text units by introducing `DynallocNewInputParms`, but on the other hand there is `TextUnitsSize`. 
* What does one need to pass in that argument?
* The name shouldn't start with a capital letter

In my opinion, we should either pass raw text units or pass the concrete data set attributes, not something in between.

aditya-ranshinge: I have renamed variable name to `inputParmsCount` as it is `DynallocNewInputParms` count passed from caller. 
In `dynallocDataset` there are fixed 3 number of `inputParms` but in `dynallocNewDataset`  `inputParms` count is not fixed so this argument is required.

h/dynalloc.h
```diff
@@ -696,11 +741,23 @@ typedef struct DynallocInputParms_tag {
   char reserved[3];
 } DynallocInputParms;
 
+
+typedef struct DynallocNewInputParms_tag {
+  int type;
```
ifakhrutdinov: * What do I use in `type`?
* Are the value in `key` the same as the ones used for real text units?

It looks like a leaky abstraction:
* In order to use this function I have to study the source code
* I end up using text units anyways, because this is what this `DynallocNewInputParms` is, sort of.

aditya-ranshinge: Resolved.  TextUnit structure is Used instead of DynallocNewInputParms.

JoeNemo (approved):
Approved

PR #295, opened 2022-05-26
downgrade zos max version check to warning

jackjia-ibm:
Signed-off-by: Jack (T.) Jia <jack-tiefeng.jia@ibm.com>

Similar change to https://github.com/zowe/zowe-common-c/pull/265. We tested this change and it works ok on new zOS.

### DIFF
--- a/c/le.c
+++ b/c/le.c
@@ -116,9 +116,10 @@ void abortIfUnsupportedCAA() {
   unsigned int zosVersion = ecvt->ecvtpseq;
 #ifndef METTLE
   if (zosVersion > LE_MAX_SUPPORTED_ZOS) {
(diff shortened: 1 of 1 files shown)


JoeNemo:
What version are you testing?  It looks like the code is working up to ZOS 2.5?

jackjia-ibm:
v2.5 is GA-ed. This is more about future internal z/OS versions not disclosed yet. The team wants to know if Zowe is working ok on new releases.

JoeNemo:
Until Zowe has Early Access to ZOS, we can't support doing things that only IBM members of the community sees.   You can keep this change in your own sandbox.   There is no QA possible for unreleased versions.

JoeNemo:
This will not happen as such.

PR #304, opened 2022-06-25
Support for REXX using ConfigMgr, and some clang-driven compile warni…

JoeNemo:
…ng cleanup


jordanfilteau1995 (approved):
Check for any not wanted printf's.

PR #310, opened 2022-07-06
fix missing XPLINK 31 cases in code

JoeNemo:

### DIFF
--- a/c/recovery.c
+++ b/c/recovery.c
@@ -17,6 +17,7 @@
 #include <metal/stdio.h>
 #include <metal/stdlib.h>
 #include <metal/string.h>
+#include <metal/stdint.h>
 #include <metal/inttypes.h>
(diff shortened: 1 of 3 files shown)


c/recovery.c
```diff
@@ -17,12 +17,14 @@
 #include <metal/stdio.h>
 #include <metal/stdlib.h>
 #include <metal/string.h>
```
ifakhrutdinov, line 19: @JoeNemo can you include `#include <metal/stdint.h>`? Otherwise Metal won't compile because `inttypes.h` uses `#include <stdint.h>` which results in including the LE version because of how our build scripts are written...

ifakhrutdinov (approved):
Please see my comment about `stdint.h`, otherwise this looks good to me, my tests pass, including LE 31-bit xplink.

PR #311, opened 2022-07-06
Make yaml strings quoted

1000TurquoisePogs:
Joe and I found that when converting json to yaml, it would unquote strings, which could then be misinterpreted as integers etc.
We found a way to tell the yaml parser to keep the quotes.


JoeNemo (approved):
Looks good.

PR #314, opened 2022-07-14
change branchname

1000TurquoisePogs:
Changing branch naming convention to align with others used in zowe


1000TurquoisePogs:
Merging this after consulting with james that this is an OK branch name scheme.

PR #316, opened 2022-07-21
object case was shadowing array in jsToJSON

JoeNemo:


1000TurquoisePogs (approved):
Tested as working

PR #317, opened 2022-07-22
Fix for large allocation on stack during unix file sending causing 0c4

1000TurquoisePogs:
It was noticed during testing of zss 31 bit that we were allocating too much memory on the stack during unix file serving. This has been switched to safeMalloc and appears to work well on 31 and 64.
Additionally, the error messages from recovery have been switched from debug3 to severe, and several unix file serving messages from debug3 to debug.


JoeNemo (approved):
This looks correct.

PR #321, opened 2022-08-04
fixing build for rexxcmgr.sh

jordanfilteau1995:
build_rexxcmgr.sh was failing due to missing files for new semantic validation.


JoeNemo (approved):
Looks good.

PR #325, opened 2022-08-30
rexx cfgmgr bugfixes and cleanup

jordanfilteau1995:
hiding printf behind traceLevel; fixing build for cfgmgr-rexx; optioally print validity exceptions


JoeNemo (approved):
Looks good!

PR #336, opened 2022-09-29
Improvements to ConfigMgr validation errors.

jordanfilteau1995:
better error messages for const, enum, type; only string and number supported for const and enum, should support int64.

### DIFF
--- a/c/jsonschema.c
+++ b/c/jsonschema.c
@@ -1290,10 +1290,10 @@ static VResult validateJSON(JsonValidator *validator,
     if (!eq){
       if (jsonIsString(valueSpec->constValue)) {
         return simpleFailure(validator,"unequal constant value at %s expecting value '%s' of type '%s'",
-                             validatorAccessPath(validator), jsonAsString(valueSpec->constValue), valueSpec->type);
+                             validatorAccessPath(validator), jsonAsString(valueSpec->constValue), getJSTypeName(valueSpec->type));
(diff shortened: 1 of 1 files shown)


jordanfilteau1995:
Example...

Validity Exceptions(s) with object at
  Validity Exceptions(s) with object at /zowe
    Validity Exceptions(s) with object at /zowe/setup
      Validity Exceptions(s) with object at /zowe/setup/certificate
        no matching enum value at /zowe/setup/certificate/type; expecting one of values [PKCS12, JCERACFKS] of type 'string'

  Validity Exceptions(s) with object at /zOSMF
    type 'integer' not permitted at /zOSMF/host; expecting type 'string'

unequal constant value at /components/izp/security/surrogateUser/super; expecting value 'IZPSRGSP' of type 'string'

jordanfilteau1995:
@1000TurquoisePogs Ready for review.

c/jsonschema.c
```diff
@@ -123,6 +123,7 @@ typedef struct PatternProperty_tag {
 
 typedef struct JSValueSpec_tag {
   int      typeMask;
+  int      type;
```
1000TurquoisePogs, line 126: this needs to be at the end of the struct to not wreck any plugin that depends upon the old struct.

1000TurquoisePogs (approved):
merging this and then fixing the struct in a second pr.

PR #343, opened 2022-11-07
Add an RLE field for application data

ifakhrutdinov:
#### Overview
This pull-request adds a new field to the RLE anchor which can be used to anchor any user data. In case of ZIS, that will be the ZIS context data structure.


JoeNemo (approved):
Approved.

PR #344, opened 2022-11-11
Add xplatform.appendFileUTF8

1000TurquoisePogs:
This adds xplatform.appendFileUTF8 as it was needed by zwe. zwe wants to write log messages to a file and the existing storefile was not an efficient way to accomplish this.


JoeNemo (approved):
This is a good and simple addition to the code.

PR #345, opened 2022-11-15
Add error codes to configmgr

1000TurquoisePogs:
configmgr had error codes that were not exposed to users calling the binary directly. so, to make success/fail decision making easier for users, i've changed the program to give return codes.


JoeNemo (approved):
This is simple and straightforward.

PR #364, opened 2023-03-10
Fix that file copy writing random bytes to destination

1000TurquoisePogs:
The file copy operation, when not doing convert codepage, would initialize the write buffer to itself, so that when we read X bytes, we would write X bytes of random junk to the destination.
I instead initialize the write buffer = read buffer when conversion is not needed, and this solves the issue.


JoeNemo (approved):
Approved.

PR #367, opened 2023-03-22
Fix no uid/gid shown in respondWithUnixFileMetadata 

1000TurquoisePogs:
respondWithUnixFileMetadata returns the owner / group strings of files when its known.
But, if a user account is removed, there is no mapping between a UID/GID number and a name.
If you `ls` such files, you'll see ls just returns the ID instead of the name when the name no longer exists.
So, we'll do the same in this API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,9 @@
 # Zowe Common C Changelog
 
-## `2.9.0`
-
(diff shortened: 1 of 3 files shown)


JoeNemo (approved):
Approved, and need issue for potential code code duplication.

PR #371, opened 2023-04-05
Client Certificate Authentication with R_usermap

jordanfilteau1995:
Client certificate is grabbed from gsk after connection is established.


c/httpserver.c
```diff
@@ -3161,7 +3172,30 @@ static int serviceAuthNativeWithSessionToken(HttpService *service, HttpRequest *
...
         authDataFound = TRUE;
+      } 
+    } 
+  }
+
+  char userid[9] = {0};
+  int mapReturnCode = 0, mapReasonCode = 0;
+  int rc = 0;
+
+  if (authDataFound == FALSE) {
```
jordanfilteau1995, line 3183: Might need to check for session token existing here.

c/tls.c
```diff
@@ -109,8 +164,7 @@ int tlsSocketInit(TlsEnvironment *env, TlsSocket **outSocket, int fd, bool isSer
   if (label) {
     rc = rc || gsk_attribute_set_buffer(socket->socketHandle, GSK_KEYRING_LABEL, label, 0);
   }
-  rc = rc || gsk_attribute_set_enum(socket->socketHandle, GSK_SESSION_TYPE,
-                                    isServer ? GSK_SERVER_SESSION : GSK_CLIENT_SESSION);
+  rc = rc || gsk_attribute_set_enum(socket->socketHandle, GSK_SESSION_TYPE, isServer ? GSK_SERVER_SESSION_WITH_CL_AUTH : GSK_CLIENT_SESSION);
```
ifakhrutdinov, line 155: Will we never be able to use this **without** a client certificate on the server side? Asking because it's either `GSK_SERVER_SESSION_WITH_CL_AUTH` or `GSK_CLIENT_SESSION`.

I can see cases where one would want to use this code just for a secure connection but without client authentication required.

c/tls.c
```diff
@@ -123,6 +177,7 @@ int tlsSocketInit(TlsEnvironment *env, TlsSocket **outSocket, int fd, bool isSer
     safeFree((char*)socket, sizeof(*socket));
     *outSocket = NULL;
   }
+  //printf("tlsSocketInit - rc=%d\n", rc);
```
ifakhrutdinov: No need for this.

c/tls.c
```diff
@@ -15,6 +15,40 @@
...
+
+  int rc = gsk_attribute_get_cert_info(soc_handle, GSK_PARTNER_CERT_INFO, &gskCertificateArray, &gskCertificateArrayElementCount);
+
+  if (rc != 0) {
+    return rc;
+  }
+
+  gsk_cert_data_elem *tmp = gskCertificateArray;
+  
+  for (int i = 0; i++ < gskCertificateArrayElementCount; tmp++) {
```
ifakhrutdinov: I would not increment in a condition; this seems error prone.

What about this?
```c
for (int i = 0; i < gskCertificateArrayElementCount; i++) {
  const gsk_cert_data_elem *elem = &gskCertificateArray[i];
  . . .
}
```

c/tls.c
```diff
@@ -15,6 +15,40 @@
 #include "fdpoll.h"
 #include "tls.h"
 
+int getClientCertificate(gsk_handle soc_handle, char *clientCertificate, unsigned int clientCertificateBufferSize, unsigned int *clientCertificateLength) {
+
+  if (clientCertificate == NULL || clientCertificateBufferSize <= 0) {
```
ifakhrutdinov, line 22: Nit: `clientCertificateBufferSize` can't be less than 0.

c/tls.c
```diff
@@ -15,6 +15,40 @@
...
+    return rc;
+  }
+
+  gsk_cert_data_elem *tmp = gskCertificateArray;
+  
+  for (int i = 0; i++ < gskCertificateArrayElementCount; tmp++) {
+    if (tmp->cert_data_id == CERT_BODY_DER) {
+      if (clientCertificateBufferSize >= tmp->cert_data_l) {
+        memcpy(clientCertificate, tmp->cert_data_p, tmp->cert_data_l);
+        *clientCertificateLength = tmp->cert_data_l;
```
ifakhrutdinov, line 43: At this point we've copied the data and continue. Is it okay to overwrite it if one of the next array entries matches the condition or do we need to break here?

c/tls.c
```diff
@@ -43,13 +94,15 @@ int tlsInit(TlsEnvironment **outEnv, TlsSettings *settings) {
     safeFree((char*)env, sizeof(*env));
     *outEnv = NULL;
   }
+  //printf("tlsInit - rc=%d\n", rc);
```
ifakhrutdinov: Please clean up dead code.

c/tls.c
```diff
@@ -29,6 +63,23 @@ int tlsInit(TlsEnvironment **outEnv, TlsSettings *settings) {
...
+   *
+   * rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_CLIENT_AUTH_TYPE, GSK_CLIENT_AUTH_PASSTHRU_TYPE);
+   */
+
+#ifdef DEV_DO_NOT_VALIDATE_CLIENT_CERTIFICATES
+  rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_CLIENT_AUTH_TYPE, GSK_CLIENT_AUTH_PASSTHRU_TYPE);
+#endif
+
+  /*
+   * Only allow requests with client certificates, maybe put behind a different ifdef.
```
ifakhrutdinov: Dead code?

c/tls.c
```diff
@@ -29,6 +63,23 @@ int tlsInit(TlsEnvironment **outEnv, TlsSettings *settings) {
   rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_PROTOCOL_TLSV1_1, GSK_PROTOCOL_TLSV1_1_OFF);
   rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_PROTOCOL_TLSV1_2, GSK_PROTOCOL_TLSV1_2_ON);
   rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_SERVER_EPHEMERAL_DH_GROUP_SIZE, GSK_SERVER_EPHEMERAL_DH_GROUP_SIZE_2048);
+
+  /*
+   * Don't validate certificates, maybe put behind a dangerous ifdef.
+   *
+   * rc = rc || gsk_attribute_set_enum(env->envHandle, GSK_CLIENT_AUTH_TYPE, GSK_CLIENT_AUTH_PASSTHRU_TYPE);
```
ifakhrutdinov: Dead code?

c/httpserver.c
```diff
@@ -2653,19 +2653,23 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
...
-    } else {
-      zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "non-mixed-case system, not a pass phrase, upfolding password\n");
-      strupcase(request->password); /* upfold password */
+      if (isLowerCasePasswordAllowed() || isPassPhrase(request->password)) {
+        zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "mixed-case system or a pass phrase, not upfolding password\n");
+        /* don't upfold password */
+      } else {
+        zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "non-mixed-case system, not a pass phrase, upfolding password\n");
+        strupcase(request->password); /* upfold password */
+      }
```
ifakhrutdinov, line 2672: Do we need these white space changes?

ifakhrutdinov:
@jordanfilteau1995 , please add a detailed description of these changes. I can see at least one change which isn't related to certificates. Hoe does the security logic changes?

c/httpserver.c
```diff
@@ -2675,10 +2679,17 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
...
+      pwdCheckRC = zisCheckUsernameAndPassword(privilegedServerName,
+          request->username, request->password, &status);
+      authResponse->type = AUTH_TYPE_RACF;
+      authResponse->responseDetails.safStatus = status.safStatus;
+    } else {
+      pwdCheckRC = zisCheckUsername(privilegedServerName,
+          request->username, &status);
+      authResponse->type = AUTH_TYPE_RACF;
+      authResponse->responseDetails.safStatus = status.safStatus;
+    }
```
ifakhrutdinov, line 2692: I need more info about this and why we need this.

c/httpserver.c
```diff
@@ -3162,9 +3173,51 @@ static int serviceAuthNativeWithSessionToken(HttpService *service, HttpRequest *
       if (service->authExtractionFunction(service, request) == 0){
         authDataFound = TRUE;
       }
+    } 
+  }
+
+  /* Doubtful that it would be greater than 8k... */
```
ifakhrutdinov: Can we know for sure? I.e. references to standards/docs would help here.

c/httpserver.c
```diff
@@ -2653,19 +2653,23 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
   } else if (authDataFound){
     ACEE *acee = NULL;
     strupcase(request->username); /* upfold username */
+    if (request->password == NULL) {
```
ifakhrutdinov: Using the password field to indicate whether this is a passwordless auth seems like a bad idea. If for some unexpected reason that field is `NULL` (is it `NULL` if a user just doesn't provide a password in the basic auth?), you'll assume this is just a passwordless auth and will use `zisCheckUsername`, which might say everything is okay.

This seems like an error prone solution with implicit logic.

Regardless of whether this approach with client certificates is correct in the first place, using a separate variable to explicitly indicate what needs to be checked to authenticate would probably be better.

ifakhrutdinov (commented):
I'll let Jordan add a description before continuing with the review. For now, see my current comments.

JoeNemo:
@jordanfilteau1995 , can we discuss a little more.  We probaby need to refactor for clarity because the different authentication methods are a little to close in the code, and could share variables and state in unexpected ways.   We also may need early help in testing multiple auth methods on a branch before merging.

c/tls.c
jordanfilteau1995: From my understanding and testing, it just allows client authentication to be an option if a client certificate is attached. It behaves the same way as usual if you don't provide a client certificate.

c/httpserver.c
JoeNemo: Irek, we need either a new API, or a new arg in the existing API to say if password is required.

c/httpserver.c
JoeNemo: Check the R_usermap doc for limits.  Some 'net searching yielded 8K, blame me for this.

JoeNemo (approved):
Approved

PR #378, opened 2023-05-24
symbol resolution library for C and QJS

1000TurquoisePogs:
This feature can be tested via configmgr and a builtin test.

```
> ./configmgr -script ../tests/js/symbol.js
Resolved static symbol &SYSNAME to "RS28"
Resolved dynamic symbol &DAY to "24"
done
```

It adds a function, `resolveSymbol`, available both as a C function and as a QJS zos module function.
It takes in a string starting with `&` and not ending with `.`, and will lookup the string to find the dynamic or static symbol value that corresponds to it.

This is done with a combination of techniques.
At first, it will read through the symbol table off the ecvt to see if it's a static symbol.
If it isn't there, it may be a dynamic symbol, and that can be checked through a bit of assembly code that runs next. If still not found, null is returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Zowe Common C Changelog
 
+## `2.9.0`
+
+- Feature: configmgr's zos module now has a "resolveSymbol" function which takes a string starting with & which can be used to resolve static and dynamic zos symbols
+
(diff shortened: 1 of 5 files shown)


JoeNemo (approved):
Approved

PR #390, opened 2023-07-27
Allow bypassing zos version check with env var ZWE_zowe_launcher_unsafeDisableZosVersionCheck

1000TurquoisePogs:
This PR changes the zOS version check from an error to a warning in the presence of the env var ZWE_zowe_launcher_unsafeDisableZosVersionCheck
It must be an env var because of the early execution of this code. Putting it into the yaml will not work.

### DIFF
--- a/c/le.c
+++ b/c/le.c
@@ -120,9 +120,15 @@ void abortIfUnsupportedCAA() {
   unsigned int zosVersion = ecvt->ecvtpseq;
 #ifndef METTLE
   if (zosVersion > LE_MAX_SUPPORTED_ZOS) {
-    printf("error: z/OS version = 0x%08X, max supported version = 0x%08X - "
-           "CAA fields require verification\n", zosVersion, LE_MAX_SUPPORTED_ZOS);
-    abort();
+    const char *continueWithWarning = getenv("ZWE_zowe_launcher_unsafeDisableZosVersionCheck");
+    if (!strcmp(continueWithWarning, "true")) {
+      printf("warning: z/OS version = 0x%08X, max supported version = 0x%08X - "
+             "CAA fields require verification\n", zosVersion, LE_MAX_SUPPORTED_ZOS);
+    } else {
+      printf("error: z/OS version = 0x%08X, max supported version = 0x%08X - "
+             "CAA fields require verification\n", zosVersion, LE_MAX_SUPPORTED_ZOS);
+      abort();
+    }
   }
 #else
   /* Metal uses its own copy of CAA, reserved fields will always be available */


JoeNemo (approved):
Approved.

bobbydixon:
Hi y'all, I'm working an IBM case TS019702504 where the client is bumping into the error when trying to install Zowe V2 and V3 on z/OS V3.2.  Bruce Armstrong pointed me to a thread in the #zowe-ibm-users channel, where another client reported the same issue, and Bruce, JoeW and Vijay worked with the client who tried adding variables to the yaml file, the startup JCL, but eventually changed the build flag to allow 3.2 and recompiled:

`-DLE_MAX_SUPPORTED_ZOS=0x01030200u`

It sounds like JoeW discussed this with Sean Grady, and it looks like some doc updates are necessary if we're going to keep this check in the code.

Joe-Winchester:
Chatted to the z/OS Squad who suggested trying
```
zowe:
  launcher:
    unsafeDisableZosVersionCheck: true
```
as described in the schema at https://github.com/zowe/zowe-install-packaging/blob/ac8197443e137a3fdbc2db01c697cd24d93a0615/schemas/zowe-yaml-schema.json#L560.
Updated Torin from IBM who has a 3.2 environment, and we'll try and carve out some time to give that a shot.

bobbydixon:
Hi Joe, thanks for the update.  

Would this change be made to the zowe.yaml file?

Joe-Winchester:
Hi Bobby,
Yes that's a `zowe.yaml` change.  I was told that it did work for the developer who had tried it (they didn't have 3.2 access, but they were on 3.1 and set the option back to 3.0 on the -R to prove it worked across that boundary.

bobbydixon:
My client added the suggested zowe.yaml statements, but reported that after some experimentation done by both him and another person on the z/OS 3.2 RBP team, that it looks like the check for the z/OS version happens before the yaml file is located and read.  The client deduced this by renaming their zowe.yaml file to something else, then tried again, but still received the same failure:

`error: z/OS version = 0x01030200, max supported version = 0x01030100 - CAA fields require verification`

bobbydixon:
This is all we see at the bottom of the ZWESLSTC joblog:

```
 2025-07-07 22:52:45.234 <ZWELNCH:16777548> ZWESVUSR INFO ZWEL0021I Zowe Launcher starting
 2025-07-07 22:52:45.235 <ZWELNCH:16777548> ZWESVUSR INFO -- If you cant see '500' at the end of the line, your log is too short to read!80--------90------ 100----------------------125----------------------150----------------------175-----------------
 error: z/OS version = 0x01030200, max supported version = 0x01030100 - CAA fields require verification
 CEE5207E The signal SIGABRT was received.

```

bobbydixon:
Hi @Joe-Winchester ,  please take a peek at the last couple of update and let me know who to reach out to to confirm the clients suspicions, and where to place those statements?

PR #391, opened 2023-08-03
Move WTO code from metal.c to zos.c + add shim

DivergentEuropeans:
## Proposed changes
The WTO code should be moved from metal.c to zos.c because it works in both LE C & Metal C, and can therefore be used by components like the Launcher to print to the syslog critical Zowe messages

This PR depends upon the following PRs: https://github.com/zowe/launcher/pull/93
https://github.com/zowe/zowe-install-packaging/pull/3553

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zss/blob/v2.x/master/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes
- [ ] Relevant update to CHANGELOG.md
- [ ] My changes generate no new warnings


DivergentEuropeans:
Have not tried removing it from metal.c completely yet, pending discussion

1000TurquoisePogs (dismissed):
You should define these functions in the zos.h file
But also, they need different names. zis uses both zos.c and metalio.c so same name will give conflict.

For now, wtoPrintf2 and message2 are fine, and we can adjust it after discussion later.

DivergentEuropeans:
wtoPrintf2 is already defined so using wtoPrintfMetal as a placeholder for now

h/zos.h
```diff
@@ -1529,6 +1529,10 @@ DSAB *getDSAB(char *ddname);
 
 int dsabIsOMVS(DSAB *dsab);
 
+void message2(char *message);
+
+void wtoPrintfMetal(char *formatString, ...);
```
DivergentEuropeans: rename to wtoPrintf3

h/zos.h
```diff
@@ -1529,6 +1529,10 @@ DSAB *getDSAB(char *ddname);
 
 int dsabIsOMVS(DSAB *dsab);
 
+void message2(char *message);
```
ifakhrutdinov: Why do you need this to be global? If it's not used outside zos.c, make it static and remove the declaration.

DivergentEuropeans: Without doing this, we get
` IEW2278I B352 INVOCATION PARAMETERS -
          TERM=YES,PRINT=NO,MSGLEVEL=4,STORENX=NEVER,LIST=NOIMP,XREF=YES,MAP=YE
          S,PRINT=YES,MSGLEVEL=0,ac=1,rent,case=mixed,map,xref,reus
 IEW2456E 9207 SYMBOL message2 UNRESOLVED.  MEMBER COULD NOT BE INCLUDED FROM
          THE DESIGNATED CALL LIBRARY.
 IEW2665S 40FF MODULE ZWESIS01  IS NON-EXECUTABLE AND WAS NOT SAVED BECAUSE
          STORENX=NEVER.
 IEW2008I 0F03 PROCESSING COMPLETED.  RETURN CODE =  12.
IEW5033 The binder ended with return code 12.`

1000TurquoisePogs: looks like because its used in metalio.c because metalio has a different set of functions that do wto for a different purpose than wtoPrintf was, in particular vfprintf.
I dont want to move that without knowing its purpose.

ifakhrutdinov: I don't think `message` was supposed to be global in the first place, and it's not used anywhere outside `metalio.c` Perhaps it's safe to move it to `zos.c` entirely and not have `message2`?

h/zos.h
```diff
@@ -1529,6 +1529,10 @@ DSAB *getDSAB(char *ddname);
...
+
+void wtoPrintf3(char *formatString, ...);
```
ifakhrutdinov: Why 3 and not 2?

c/zos.c
```diff
@@ -1469,6 +1469,87 @@ int safStat(int options, char *safClass, char *copy, int copyLength, int *racfSt
    get current TCB AC
    */
 
+/* begin WTO SECTION 
+
+TODO(?): This is code duplication from metalio.c but the following works for LE & metal */
+
+typedef struct WTO2Common31_tag{
```
ifakhrutdinov: Didn't we want to to duplicate this?

h/zos.h
DivergentEuropeans: wtoPrintf2 already exists @ifakhrutdinov

c/zos.c
DivergentEuropeans: Yes

DivergentEuropeans: think you meant to not duplicate

h/zos.h
DivergentEuropeans: It's not safe @ifakhrutdinov because message( ) is also used in vfprintf( ) of metalio.c , a method that isn't being moved to zos.c 

so either message2 stays or I can move message entirely to zos.c but we make it not static , and declare it in the header

DivergentEuropeans: For now, I'll keep message2 unless you think it's better otherwise

1000TurquoisePogs (approved):
now lgtm

h/zos.h
ifakhrutdinov: * `message()` wasn't supposed to be global
* `message2()` just re-enforces that flaw by also being global
* you could move `message()` and keep it global, that `vfprintf` would probably still be able use it, wouldn't it?

I suggest:
1. If you want to keep `message()` in metalio, at least call `message2()` something more descriptive, like `wtoMessage()`. `message2()` is just a bad name for a global library function.
2. Also, make the arguments in `message2()` (or `wtoMessage`) and `wtoPrintf2()` `const`, it's good practice because people will be able to pass constant literals without producing a compiler warning
```c
void message2(const char *message);
void wtoPrintf3(const char *formatString, ...);
```
Use `const` in the definitions too.

DivergentEuropeans: Makes sense 👍

ifakhrutdinov (approved):

PR #393, opened 2023-08-13
Fixing config manager error messages

jordanfilteau1995:
I've assigned message identifier to some common config manager errors involving yaml formatting (bad spaces, characters, etc). I've made it so we stop printing a double ebcdic buffer because we were wrongly converting. We also weren't converting part of the explanation given by the parser, so we missed some useful information.

For example...

Yaml File

```
firstName: George
test:
  abc: [
```

Before:

```
WARNING, yaml read failed, errorBuffer='YAML parser error: ▒▒▒%▒▒/▒▒▒>ŀ/▒▒%?π>?▒▒ at line 4, column 1
▒▒▒>?Ȁ▒▒>▒▒▒▒▒▒▒▒▒>?▒▒▒?>▒▒>▒ at line 4, column 1'
WARNING, yaml read failed, errorBuffer='▒▒▒▒@▒▒▒▒
▒@
▒▒▒▒z@while parsing a flow node@▒▒@▒▒▒
@▒k@▒▒▒▒▒▒@▒did not find expected node content@▒▒@▒▒▒
@▒k@▒▒▒▒▒▒@▒'
top json is not an object
Failed to load configuration, element may be bad, or less likey a bad merge
```

After:

```
ZWEL0318E - Couldn't parse file '/u/ts5873/git/sand/zss/deps/zowe-common-c/bin/test.yaml': while parsing a flow node at line 4, column 1, did not find expected node content at line 4, column 1.
ZWEL0319E - Failed to load configuration, element may be bad, or less likely a bad merge.
```

`


ifakhrutdinov:
Ok, I have run into this!! It's an honour to review this PR :-)

c/configmgr.c
```diff
@@ -668,11 +668,38 @@ int cfgSetConfigPath(ConfigManager *mgr, const char *configName, char *configPat
 #define YAML_ERROR_MAX 1024
 #define MAX_PDS_NAME 1000 /* very generous */
 
+static void replaceFileNameInString(char *string, char *replaceWith, int all) {
```
ifakhrutdinov: Use `bool` for `all` as in the function below.

c/configmgr.c
```diff
@@ -668,11 +668,38 @@ int cfgSetConfigPath(ConfigManager *mgr, const char *configName, char *configPat
...
+
+  index = strstr(string, "&name");
+
+  while (index != NULL) {
+    strncat(buffer, start, index - start);
+    strcat(buffer, replaceWith);
+    start = index + strlen("&name");
+    if (!all) {
+      break;
+		}
```
ifakhrutdinov: I think you're mixing tabs and space so the formatting is broken. Can you fix that?

c/configmgr.c
```diff
@@ -668,11 +668,38 @@ int cfgSetConfigPath(ConfigManager *mgr, const char *configName, char *configPat
...
+
+  char  buffer[YAML_ERROR_MAX+USS_MAX_PATH_LENGTH+1] = {0};
+  char *index;
+  char *start = string;
+
+  index = strstr(string, "&name");
+
+  while (index != NULL) {
+    strncat(buffer, start, index - start);
+    strcat(buffer, replaceWith);
```
ifakhrutdinov: You never check if the destinations (`buffer`, `string`) have enough room. Seems like this is a candidate for buffer overflow.

c/configmgr.c
```diff
@@ -682,11 +709,28 @@ static Json *readYamlIntoJson(ConfigManager *mgr, char *filename, bool allowMiss
...
     return NULL;
   } else{
-    trace(mgr,INFO,"WARNING, yaml read failed, errorBuffer='%s'\n",errorBuffer);
-#ifdef __ZOWE_OS_ZOS
-    a2e(errorBuffer,YAML_ERROR_MAX);
+#ifndef __ZOWE_OS_ZOS
+    /*
+     * Let's convert to ascii if we aren't on z/OS?
+     */
+    e2a(errorBuffer,YAML_ERROR_MAX);
```
ifakhrutdinov: Use `sizeof(errorBuffer)` since in case someone changes `char errorBuffer[YAML_ERROR_MAX];` to, for example, `char errorBuffer[YAML_ERROR_MAX + 1];`, this would cause issues.

c/configmgr.c
```diff
@@ -682,11 +709,28 @@ static Json *readYamlIntoJson(ConfigManager *mgr, char *filename, bool allowMiss
...
+     * variable is a relative path. It will make debugging more difficult.
+     *
+     * msg is a buffer that can take the max path length and error length. We'll replace &name
+     * here for better user experience.
+     */
+    char msg[YAML_ERROR_MAX + USS_MAX_PATH_LENGTH+1] = {0};
+    strcpy(msg,errorBuffer);
+    char fileOrPath[USS_MAX_PATH_LENGTH+1] = {0};
+    if (!realpath(filename,fileOrPath)) {
+      trace(mgr,DEBUG,"realpath(): ERRNO=%d\n",errno); /* i hope not */
```
ifakhrutdinov: What do you hope not?

c/yaml2json.c
```diff
@@ -123,50 +123,72 @@ static int yamlReadHandler(void *data, unsigned char *buffer, size_t size, size_
...
 static void decodeParserError(yaml_parser_t *parser, char *errorBuf, size_t errorBufSize) {
+  size_t problemLen = strlen(parser->problem);
+  char problemNative[problemLen + 1];
+  snprintf (problemNative, problemLen + 1, "%s", parser->problem);
+  convertToNative(problemNative, problemLen);
+  char *nativeContext = NULL;
+  if (parser->context) {
+    size_t contextLen = strlen(parser->context);
+    char contextNative[contextLen + 1];
+    snprintf(contextNative, contextLen + 1, "%s", parser->context);
```
ifakhrutdinov: Any reason not to just `strcpy`?

c/yaml2json.c
```diff
@@ -123,50 +123,72 @@ static int yamlReadHandler(void *data, unsigned char *buffer, size_t size, size_
...
+    convertToNative(contextNative, contextLen);
+    nativeContext = contextNative;
```
ifakhrutdinov: This is a bug, `contextNative` goes out of scope here; this is similar to returning a pointer to a local variable from a function.

c/yaml2json.c
```diff
@@ -123,50 +123,72 @@ static int yamlReadHandler(void *data, unsigned char *buffer, size_t size, size_
...
-      size_t problemLen = strlen(parser->problem);
-      char problemNative[problemLen + 1];
-      snprintf (problemNative, problemLen + 1, "%s", parser->problem);
-      convertToNative(problemNative, problemLen);
       if (parser->problem_value != -1) {
-        snprintf(errorBuf, errorBufSize, "YAML reader error: %s: #%X at %ld", problemNative, parser->problem_value, (long)parser->problem_offset);
+        snprintf(errorBuf,
+                 errorBufSize,
+                 "Couldn't read file '&name': %s, #%x at %ld.",
+                 problemNative, parser->problem_value, (long)parser->problem_offset);
```
ifakhrutdinov: Is this `size_t`? Then you don't need to cast, just use `%zu`.

c/configmgr.c
```diff
@@ -682,11 +709,28 @@ static Json *readYamlIntoJson(ConfigManager *mgr, char *filename, bool allowMiss
...
+      strcpy(fileOrPath,filename);
+    }
+		replaceFileNameInString(msg,fileOrPath,1);
```
ifakhrutdinov: Not sure about introducing such complexity for a file name. Could you just print another message here saying something "Error processing file %s" and then print the details in the next message.

ifakhrutdinov (dismissed):
There is one obvious bug where you use a dangling pointer. I'm also not sure with this name-substitution solution, see my comment about that.

c/configmgr.c
jordanfilteau1995: Good idea! Will change.

jordanfilteau1995:
@ifakhrutdinov Thank you for your review! I will get to this later today.

c/configmgr.c
jordanfilteau1995: (No longer using function.)

c/configmgr.c
jordanfilteau1995: (Fixed).

c/configmgr.c
jordanfilteau1995: (Fixed).

c/yaml2json.c
jordanfilteau1995: No. Fixed.

c/yaml2json.c
jordanfilteau1995: I don't believe it is size_t.

c/configmgr.c
jordanfilteau1995: Ended up passing the file name one function down.

jordanfilteau1995:
@ifakhrutdinov @JoeNemo I've made changes.

1. Fixed dangling pointer issue.
2. Removed overly complex string replacement logic.

c/configmgr.c
jordanfilteau1995: I think the chance of that failing is low, so it would be unfortunate if it failed. I recently discovered that when I tested with PARMLIBS it was failing, so I've now changed my code to only attempt to get the absolute path when it is an OMVS file type. both pds member names and file names are passed into this function.

c/yaml2json.c
jordanfilteau1995: I've changed to just simply allocating both and freeing at the end. The function decodeParser error just has a bunch of breaks so we can't return early without freeing.

ifakhrutdinov: You don't need to allocate anything. Simply move that local buffer out of that original `if` scope.

For example, this would cause issues:
```c
char *string = "original text";
if (someCondition) {
  char stringData[9];
  . . . // doing some work
  string = stringData;
} // stringData is not valid after this point and it is a bug to reference it
printf("%s", string); // potential segfault
```
This should be fine:
```c
char *string = "original text";
char stringData[9];
if (someCondition) {
  . . . // doing some work
  string = stringData;
} // stringData is still in the scope where we use it
printf("%s", string);
```

ifakhrutdinov: Real example:
```c
#include <stdio.h>
#include <string.h>

int main(int argc, char **argv) {

    char *string = "foo";
    {
        char stringData[9];
        printf("stringData address = %p\n", stringData);
        strcpy(stringData, "bar");
        string = stringData;
    }

    char someData[32] = {"you shouldn't see this printrf"};
    printf("someData address   = %p\n", someData);

    printf("result: %s\n", string);

    return 0;
}

> clang -O1 -o main main.c
> ./main

stringData address = 0x16ae1f3f0
someData address   = 0x16ae1f3f0
result: you shouldn't see this printrf

```
And this one is fine:
```c
#include <stdio.h>
#include <string.h>

int main(int argc, char **argv) {

    char *string = "foo";
    char stringData[9]; // <- the only change: stringData has been moved
    {
        printf("stringData address = %p\n", stringData);
        strcpy(stringData, "bar");
        string = stringData;
    }
    
    char someData[32] = {"you shouldn't see this printrf"};
    printf("someData address   = %p\n", someData);

    printf("result: %s\n", string);

    return 0;
}

> clang -O1 -o main main.c
> ./main

stringData address = 0x16d6cb40c
someData address   = 0x16d6cb3e0
result: bar

```

c/configmgr.c
ifakhrutdinov: Then I'd comment that this can potentially fail for a data set, or not comment at all.

c/yaml2json.c
ifakhrutdinov: Sorry, I wasn't clear enough, I was talking about `problem_offset`. Is that this field: https://github.com/yaml/libyaml/blob/0.2.5/include/yaml.h#L1099 ?

c/yaml2json.c
jordanfilteau1995: Hi Irek. I understand. Thank you for providing an example. :).

I don't know the length of the context until I do a `strlen()`. I was doing the `strlen()` in the null check (condition) that was causing the scoping problem.

Would you prefer I allocate a buffer outside on the stack that should be big enough to contain the context rather than allocate the exact amount on the heap?

`
char contextNative[1024] = {0};
if (parser->context) {
  if (strlen(parser->context < contextNative)) {
    strcpy(contextNative, parser->context);
  }
}
`

I could also do a `strncpy` and avoid the `strlen()` check.

c/yaml2json.c
jordanfilteau1995: Yes. I see what you mean now. Thanks!

JoeNemo (approved):
Approved.

PR #394, opened 2023-08-13
Fix introduced compiler warning from recent merge.

jordanfilteau1995:
There is a compiler warning for build_cmgr_xlclang.sh.


JoeNemo (approved):
Good. Thanks.

PR #395, opened 2023-08-21
Trim env var for startup check

1000TurquoisePogs:
Like in the launcher, any code that's expected to be run with env vars set from JCL should also expect that env vars can come in with trailing spaces.
This is a little ugly but the version-checking code in this PR is expected to be entirely removed soon, so this is another stop-gap to get people able to use zowe on untested versions of zos.

### DIFF
--- a/c/le.c
+++ b/c/le.c
@@ -123,7 +123,7 @@ void abortIfUnsupportedCAA() {
     const char *continueWithWarning = getenv("ZWE_zowe_launcher_unsafeDisableZosVersionCheck");
     /* using STDENV in JCL can result in env values that need trimming... */
     char *trimmed[strlen(continueWithWarning)+1];
-    strncopy(trimmed, continueWithWarning, strlen(continueWithWarning));
+    strncpy(trimmed, continueWithWarning, strlen(continueWithWarning));
     trimRight(trimmed, strlen(trimmed));
     
     if (!strcmp(trimmed, "true")) {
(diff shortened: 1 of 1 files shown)


c/le.c
```diff
@@ -121,7 +121,12 @@ void abortIfUnsupportedCAA() {
 #ifndef METTLE
   if (zosVersion > LE_MAX_SUPPORTED_ZOS) {
     const char *continueWithWarning = getenv("ZWE_zowe_launcher_unsafeDisableZosVersionCheck");
-    if (!strcmp(continueWithWarning, "true")) {
+    /* using STDENV in JCL can result in env values that need trimming... */
+    char *trimmed[strlen(continueWithWarning)+1];
+    strncpy(trimmed, continueWithWarning, strlen(continueWithWarning));
```
ifakhrutdinov, line 126: This will leave the destination unterminated because `trimmed` isn't initialized to zeroes; `trimRight` may 0C4.

c/le.c
```diff
@@ -121,7 +121,12 @@ void abortIfUnsupportedCAA() {
 #ifndef METTLE
   if (zosVersion > LE_MAX_SUPPORTED_ZOS) {
     const char *continueWithWarning = getenv("ZWE_zowe_launcher_unsafeDisableZosVersionCheck");
-    if (!strcmp(continueWithWarning, "true")) {
+    /* using STDENV in JCL can result in env values that need trimming... */
+    char *trimmed[strlen(continueWithWarning)+1];
```
ifakhrutdinov, line 125: @JoeNemo just pointed to a bug here. This shouldn't have `*`.

ifakhrutdinov, line 125: Can you just save the length once and reuse it?
```c
size_t valueLen = strlen(continueWithWarning);
```

PR #400, opened 2023-09-01
fixing bug with dangling pointer

jordanfilteau1995:
When you use client certificate to authenticate directly to an api rather than call `/login`, the `request->username` goes out of scope because it's allocated on the stack inside `if (authDataFound)` and results in:

`BPXTLS failed: rc=-1, return code=121, reason code=0x0be800ca`

(Which is a parameter error when calling BPXTLS.)

I'm allocating it on the SLH so we have it through the lifecycle of the request/response.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -2653,7 +2653,7 @@ static int safAuthenticate(HttpService *service, HttpRequest *request, AuthRespo
   } else if (authDataFound){
     ACEE *acee = NULL;
     strupcase(request->username); /* upfold username */
-    if (!(request->flags & HTTP_REQUEST_NO_PASSWORD)) {
+    if (request->flags & HTTP_REQUEST_NO_PASSWORD) {
       zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "Password is null. Calling safAuthenticate without a password.\n");
     } else {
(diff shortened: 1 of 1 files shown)


c/httpserver.c
```diff
@@ -3197,10 +3198,10 @@ static int serviceAuthNativeWithSessionToken(HttpService *service, HttpRequest *
      * We don't want to do this if we already found authentication data.
      */
     if (authDataFound == FALSE) {
-#define TLS_USERID_LENGTH 9
-      char userid[TLS_USERID_LENGTH] = {0};
       int racfReturnCode = 0, racfReasonCode = 0;
       zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG, "There was no token or credentials found in the request. Server is attempting to map the client certificate.\n");
+      userid = SLHAlloc(response->request->slh, 9);
+      memset(userid, 0, 9);
```
JoeNemo, line 3204: Ooof.  That pointer from heap (regular) memory to stack memory,  `char userid[]`.   It's amazing it ever worked.  Really, how did it ever work?

jordanfilteau1995: I think I was just lucky. I have the same thought...

JoeNemo (approved):
Approved.

PR #403, opened 2023-09-18
support for tls1.3

jordanfilteau1995:
Propagating ciphers and key shares to tls environment. Turning on tls 1.3 support.


1000TurquoisePogs (approved):
I tested this directly (GET /plugins) and indirectly (login to desktop with sso and access the editor) and saw both that tls 1.3 was being used, and that i saw no regressions.

Issue #404, opened 2023-10-02
Avoid using the linkage query service in recovery


ifakhrutdinov:
#### Overview
Due to performance concerns, we should avoid using IEALSQRY and replace it with something more lightweight (walking the linkage stack stack?).

PR #412, opened 2023-10-27
Feature/linkage stack query optimization

ifakhrutdinov:
### Linkage-stack query behaviour 
* Add a recovery router flag to disable linkage-stack queries globally
* Add a recovery state flag to force a linkage-stack query for that recovery state
* Disable linkage-stack queries in the PC-ss and PC-cp routines globally

The reason I have not made all the queries an opt-in feature globally is that there is actually code which may break (not in ZIS, but in AUX), so this seemed like a good compromise to me.

I'm also thinking of making the router level switch an external parameter in case it breaks someone else's plug-in.

### Performance optimizations
* Use `EPSW` instead of `ESTA` to extract the key

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/404

## Type of change
- [x] [**not clear yet**] Breaking change (fix or feature that would cause existing functionality to not work as expected) 

## PR Checklist
- [ ] Relevant update to CHANGELOG.md

## Testing

TBD


JoeNemo (approved):
Hope it works well in testing.

PR #414, opened 2023-11-20
Add cipher customization support

1000TurquoisePogs:
adds reading array of iana cipher suite strings to customize ciphers


jordanfilteau1995 (approved):
No issues found. Static array is a good strategy to me.

PR #417, opened 2023-11-28
move getesm out of "utils" artifactory location

1000TurquoisePogs:
I thought /org/zowe/utils/getesm.pax would be a fine location, but then realized this has limitations in zowe-install-packaging which uses the structure "org.zowe.utils" to refer to only one object, so i cannot make a nice utils folder like that. instead, i will just call it "org.zowe.getesm"

### DIFF
--- a/.github/workflows/build-getesm.yml
+++ b/.github/workflows/build-getesm.yml
@@ -71,7 +71,7 @@ jobs:
         with:
           artifacts: |
             .pax/getesm/getesm.pax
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs:
success: build went to 
      "target": "libs-snapshot-local/org/zowe/getesm/2.13.0-BUGFIX-GETESM-MOVE/getesm-2.13.0-2023112829.pax"

Martin-Zeithaml (approved):

PR #418, opened 2023-11-28
Add logging for json

1000TurquoisePogs:
JSON logging was hidden behind a DEBUG flag. Instead it is easier to use zowelog in the form of "_zss.json"


Martin-Zeithaml:
I have added the new logger to the doc update [PR3067](https://github.com/zowe/docs-site/pull/3067/files).

c/json.c
```diff
@@ -210,14 +200,14 @@ void jsonWriteBufferInternal(jsonPrinter *p, char *text, int len) {
       loopCount++;
       if (newWriteReturn < 0) {
         /* TODO: Replace by zowelog(...) */
```
Martin-Zeithaml: Seems this can be deleted.

c/json.c
```diff
@@ -210,14 +200,14 @@ void jsonWriteBufferInternal(jsonPrinter *p, char *text, int len) {
...
-        JSONERROR("JSON: write error, rc %d, return code %d, reason code %08X\n",
+        zowelog(NULL, LOG_COMP_JSON, ZOWE_LOG_SEVERE, "JSON: write error, rc %d, return code %d, reason code %08X\n",
               newWriteReturn, returnCode, reasonCode);
         jsonSetIOErrorFlag(p);
         break;
       }
       if (loopCount > 10) {
         /* TODO: Replace by zowelog(...) */
```
Martin-Zeithaml: Seems this can be deleted.

c/json.c
```diff
@@ -185,7 +175,7 @@ void jsonWriteBufferInternal(jsonPrinter *p, char *text, int len) {
   if (jsonShouldStopWriting(p)) {
     return;
   }
-  JSON_DEBUG("write buffer internal: text at %p, len %d\n", text, len);
+  zowelog(NULL, LOG_COMP_JSON, ZOWE_LOG_DEBUG, "write buffer internal: text at %p, len %d\n", text, len);
   DUMPBUF(text, len);
```
ifakhrutdinov, line 179: Perhaps, it's worth replacing `DUMPBUF` with `zowedump`.

c/json.c
```diff
@@ -322,11 +310,11 @@ writeBufferWithEscaping(jsonPrinter *p, size_t len, char text[len]) {
 
     UTF8_GET_NEXT_CHAR(len, text, &i, &utf8Char, &getCharErr);
     if (getCharErr != 0) {
-      JSONERROR("JSON: invalid UTF-8, rc %d\n", getCharErr);
+      zowelog(NULL, LOG_COMP_JSON, ZOWE_LOG_SEVERE, "JSON: invalid UTF-8, rc %d\n", getCharErr);
```
ifakhrutdinov, line 313: I'm not sure if library code should start spitting out messages like this. I think, that in general, it's best to pass  info to the caller and the caller would decide if/what to print. Also, we don't know if after these changes, an error in JSON can cause log flooding.

One other thing is, we don't have message IDs for non-debug messages here. It's not a request to change this PR, but we should consider improving the messages here in the future.

JoeNemo:
This needs log limiting before it can be committed

PR #437, opened 2024-03-12
Using a temporary buffer pointer to avoid pointer corruption.

Gautham-coder:
## Proposed changes
The original pointer used for malloc() was getting corrupted while data write was happening, now we use a copy/temporary pointer. Thereby safeguarding the original pointer and avoiding the wrong free(), which was causing the abend.

This PR addresses Issue: https://github.com/zowe/zss/issues/693

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas

## Testing
You can create a script to upload chunks of data continuously using the PUT method.

### DIFF
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -935,12 +935,12 @@ int writeAsciiDataFromBase64(UnixFile *file, char *fileContents, int contentLeng
                              &reasonCode);
      if (status == 0) {
        int writtenLength = 0;
-       char *dataToWriteToFile = dataToWrite;
+       char *currentBufferStart = dataToWrite;
        while (dataSize != 0) {
-         writtenLength = fileWrite(file, dataToWriteToFile, dataSize, &returnCode, &reasonCode);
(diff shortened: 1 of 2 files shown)


c/httpfileservice.c
JoeNemo, line 1: Is dataToWrite being used for any other purpose.   It looks like the entire problem is that the author said "-=" when they should have said "+=" in the first place.

Gautham-coder: Hi Joe, it is not getting modified anywhere else. But we cannot do the correct operation '+=' on dataToWrite, because it is the pointer we get from malloc. Hence, I have used a temporary pointer.

JoeNemo: Ok.   Please call it "currentBufferStart" or something like that rather than "dateToWriteToFile" because that expresses it's intent better.

JoeNemo (approved):
Approved.

PR #438, opened 2024-03-12
changing certificate logging from info to debug

jordanfilteau1995:
This is being printed over and over again. It should be changed to a debug.


1000TurquoisePogs (approved):
Makes sense. I see the same under specific circumstances (mainly, APIML not being used.)

PR #444, opened 2024-04-29
ZSS Logging Improvements

### LINKED ISSUE #167: Extend the logging facility to include additional information in messages
#### Overview
At the moment there is a community requirement to prefix all the log messages with additional information like timestamp, line number, severity level, etc.

Here's an example
```
2019-03-19 11:23:57.776 <ZWEAGW1:threadInformation> userID INFO (locationInformation) ZWEA001I Message text
```

The existing logging facility has the ability to use a custom formatter but the caller cannot pass the context information to it (e.g. the line number). The purpose of this ticket is to discuss how we could re-factor/extend the code in `logging.c` to meet the new requirements while keeping backward compatibility.

jordanfilteau1995:
## Proposed changes

Changing the logging format to have useful information. I've added the following:

    - Time Stamp
    - TCB (Task Control Block) Address
    - Thread ID
    - Process ID
    - File Name
    - Function Name
    - Line Number
    - Severity Level

It looks like this:

2024-04-30 18:53:40.345 <ZWESZ1:0x8FB2F8:0x37455E58:67241934> TS5873 CRITICAL (loggingtest.c:main:18) oh no!

## Type of change
- [X] New feature (non-breaking change which adds functionality)

## Testing
It should be non-breaking. If the new code is in, then the log messages done by ZSS will have the additional metadata by default.

Output from `loggingtest.c`.

```
2024-04-30 18:53:40.345 <ZWESZ1:0x8FB2F8:0x37455E58:67241934> TS5873 CRITICAL (loggingtest.c:main:18) Hello from zowelog2(). This log should have every field.
2024-04-30 18:53:40.345 <ZWESZ1:0x8FB2F8:0x37455E58:67241934> TS5873 INFO Hello from zowelog(). This log should lack fileLocation information.
2024-04-30 18:53:40.345 <ZWESZ1:0x8FB2F8:0x37455F20:67241934> TS5873 NA Hello from zowelog(). This log should lack level and fileLocation information.
```

This demonstrates support for backwards compatibility and native improvements to all current log statements, without changing code.

## Further comments
Note: I think that we should move toward a logger that takes a message key and has messages stored in JSON or YAML with various metadata. This is something that I believe would be useful.

`
    {
      "key": "someZssError",
      "prefix": "ZSS",
      "number": 1,
      "severity": "I",
      "format": "This is a test message: %s."
    }
`


1000TurquoisePogs (dismissed):
I love the idea of improving the logger but please do not invent a different syntax.

This is the format all of the servers except zis/zss are following:
https://github.com/zowe/zac/issues/90

Please revise the PR to make zss match that.
( note, launcher already does as https://github.com/zowe/launcher/blob/v2.x/staging/src/main.c#L288 )

jordanfilteau1995:
@1000TurquoisePogs Gotcha! I've made changes to reflect the documentation you sent me.

h/logging.h
```diff
@@ -135,7 +160,11 @@ typedef void (*LogHandler)(struct LoggingContext_tag *context,
                            LoggingComponent *component,
```
jordanfilteau1995, line 160: Look into making `LogHandler2` as Irek pointed out.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+
+  struct timeval now;
+  gettimeofday(&now, NULL);
+  int milli = now.tv_usec / 1000;
+  snprintf(result.value+strlen(result.value), 5, ".%03d", milli);
+
+  return result;
+}
+
+static char *getServiceName() {
```
ifakhrutdinov: Try to use `const` whenever possible; e.g., here you could return `const char *` because you wouldn't want anyone to assume this string could be mutated.

h/logging.h
```diff
@@ -147,6 +183,7 @@ typedef struct LoggingDestination_tag{
   char  *name;
   void  *data;          /* used by destination to hold internal state */
   LogHandler handler;
+  LogHandler2 handler2;
```
ifakhrutdinov: I'm wondering if adding this in the middle can cause issues. Also, do we have code that creates this structure manually (or accesses the fields) and then passes to the functions in `logging.c` or do we always create it via `logConfigureDestinationX`?

jordanfilteau1995: I've moved it to the end regardless for safety.

c/logging.c
jordanfilteau1995: Gotcha. I made this change.

h/logging.h
```diff
@@ -255,9 +292,11 @@ extern LoggingContext *theLoggingContext;
 #define getLoggingContext GTLOGCTX
 #define setLoggingContext STLOGCTX
 #define zowelog ZOWELOG
+#define zowelog2 ZOWELOG2
 #define zowedump ZOWEDUMP
 #define logConfigureDestination LGCFGDST
 #define logConfigureDestination2 LGCFGDS2
+#define logConfigureDestination3 LGCFGDS3
```
ifakhrutdinov, line 301: There is no declaration for `logConfigureDestination3`. Please add it.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+    case ZOWE_LOG_DEBUG: return "DEBUG";
+    case ZOWE_LOG_DEBUG2: return "FINER";
+    case ZOWE_LOG_DEBUG3: return "TRACE";
+    default: return "NA";
   }
-  return destination;
 }
 
-void printStdout(LoggingContext *context, LoggingComponent *component, void *data, char *formatString, va_list argList){
+typedef struct zl_time_t {
```
ifakhrutdinov: Even though I prefer this naming style over camel-case, we shouldn't mix the two.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+
+static zl_time_t gettime(void) {
+
+  time_t t = time(NULL);
+  const char *format = "%Y-%m-%d %H:%M:%S";
+
+  struct tm lt;
+  zl_time_t result;
+
+  gmtime_r(&t, &lt);
```
ifakhrutdinov, line 494: This call can fail and return a `NULL`; in that case, I'm not sure if `lt` is safe to use on the subsequent calls (e.g., it's left unterminated).

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+
+  strftime(result.value, sizeof(result.value), format, &lt);
```
ifakhrutdinov: The result needs to be checked because, based on the doc:
> If the length of the result string (including the terminating null byte) would exceed max bytes, then strftime() returns 0, and the contents of the array are undefined.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+
+  struct timeval now;
+  gettimeofday(&now, NULL);
```
ifakhrutdinov: This can fail.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+  int milli = now.tv_usec / 1000;
+  snprintf(result.value+strlen(result.value), 5, ".%03d", milli);
```
ifakhrutdinov: Why 5? I also suggest you check the return value for `-1` and at least make sure the buffer is terminated if the call fails.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+
+  return result;
+}
+
+static const char *getServiceName() {
+  return "ZWESZ1";
+}
+
+static void getTaskInformation(char *taskInformation, unsigned int taskInformationSize) {
```
ifakhrutdinov, line 515: It's also fine to use `size_t` as the type for sizes because it's what the C library functions do (I know, you're probably not going to have a buffer larger than 4GB but that's not the only reason).

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
   }
-  return destination;
 }
 
-void printStdout(LoggingContext *context, LoggingComponent *component, void *data, char *formatString, va_list argList){
+typedef struct zl_time_t {
+  char value[32];
+} zl_time_t;
+
+static zl_time_t gettime(void) {
```
ifakhrutdinov: With all the non-Metal stuff, the ZIS build fails now:
```
ERROR CCN3277 ../../deps/zowe-common-c/c/logging.c:490   Syntax error: possible missing ';' or ','?
ERROR CCN3045 ../../deps/zowe-common-c/c/logging.c:490   Undeclared identifier time_t.
ERROR CCN3007 ../../deps/zowe-common-c/c/logging.c:493   "struct tm" is undefined.
ERROR CCN3045 ../../deps/zowe-common-c/c/logging.c:496   Undeclared identifier t.
ERROR CCN3007 ../../deps/zowe-common-c/c/logging.c:500   "struct timeval" is undefined.
ERROR CCN3277 ../../deps/zowe-common-c/c/logging.c:513   Syntax error: possible missing ';' or ','?
ERROR CCN3045 ../../deps/zowe-common-c/c/logging.c:513   Undeclared identifier pthread_t.
ERROR CCN3045 ../../deps/zowe-common-c/c/logging.c:514   Undeclared identifier pThread.
ERROR CCN3277 ../../deps/zowe-common-c/c/logging.c:605   Syntax error: possible missing ')' or ','?
CCN0793(I) Compilation failed for file ../../deps/zowe-common-c/c/logging.c.  Object file not created.
```

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+static const char *getServiceName() {
+  return "ZWESZ1";
+}
+
+static void getTaskInformation(char *taskInformation, unsigned int taskInformationSize) {
+  pthread_t pThread = pthread_self();
+  snprintf(taskInformation, taskInformationSize, "0x%p:0x%p:%d", getTCB(), &pThread, getpid());
+}
+
+static void getMessage(char *message, unsigned int messageSize, const char *formatString, va_list argList) {
```
ifakhrutdinov, line 520: Why do you need this one liner with such a generic name? I think if you called `vsnprintf` in place, it'd be clearer what your intention was.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+}
+
+/*
+ * 1. Try to get the name from the TCB.
+ * 2. Try to get the name from the login data base.
+ */
+static void getUserID(char *user, unsigned int userSize) {
+  TCB *tcb = getTCB();
+  ACEE *acee = (ACEE *) tcb->tcbsenv;
+  if (acee) {
```
ifakhrutdinov, line 531: But no ACEE can also mean no impersonation and there should probably be a job-level ACEE and user in that case. Why have you chosen `getlogin_r` over that? And will `getlogin_r` return the proper value if zss is running as an STC?

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+static void prependMetadata(int logLevel,
+                            char *fullMessage,
+                            unsigned int fullMessageSize,
+                            char *formatString,
+                            va_list argList,
+                            char *fileName,
+                            char *functionName,
+                            unsigned int lineNumber) {
+  char user[LOG_USER_ID_MAX_LENGTH + 1] = {0};
+  getUserID((char *) &user, sizeof(user) - 1);
```
ifakhrutdinov, line 556: The function already null-terminates, so this may confuse people.

c/logging.c
ifakhrutdinov: Can you name this function according to the style of this file? E.g., `getCurrentTime`.

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+             user,
+             logLevelAsString,
+             basename(fileName),
+             functionName,
+             lineNumber,
+             message);
+  }
+}
+
+void printToLog(FILE *destination,
```
ifakhrutdinov: Why is this global? Why does this have such a generic name while also not working in Metal C?

c/logging.c
```diff
@@ -439,47 +444,219 @@ LoggingDestination *logConfigureDestination(LoggingContext *context,
...
+   * logging format.
+   *
+   * If one of these (1,2,3,4) isn't present, then they all aren't based on this implementation; therefore, we'll just format with what's
+   * available to us without relying on new arguments. The previous LogHandler will return null or empty values for (1,2), a 0 for (3), and
+   * ZOWE_LOG_NA for (4), which when going thorugh logLevelToString() will be NA.
+   *
+   * The existing logs in the server should contain the level, as long as logConfigureDestination3() is used. If not, it will have the placeholder.
+   *
+   */
+  if (!fileName || strlen(fileName) == 0 || !functionName || strlen(functionName) == 0 || lineNumber == 0) {
```
ifakhrutdinov: Could you factor this check out and call it something meaningful?

For example (you may want to change the name to something more fitting):
```c
bool newFormat = fileName && strlen(fileName) != 0 && functionName &&
                 strlen(functionName) != 0 && lineNumber != 0;
```

c/logging.c
```diff
@@ -649,7 +826,14 @@ int logGetLevel(LoggingContext *context, uint64 compID){
   return component ? component->currentDetailLevel : ZOWE_LOG_NA;
 }
 
-void zowelog(LoggingContext *context, uint64 compID, int level, char *formatString, ...){
+void zowelogInner(LoggingContext *context,
```
ifakhrutdinov: Should this be static?

c/logging.c
```diff
@@ -748,15 +953,22 @@ void zowedump(LoggingContext *context, uint64 compID, int level, void *data, int
...
       }
       else {
         break;
       }
     }
 
   }
+}
 
+void zowedump2(LoggingContext *context, uint64 compID, int level, void *data, int dataSize, char *fileName, char *functionName, unsigned int lineNumber) {
```
ifakhrutdinov: No declaration.

c/logging.c
```diff
@@ -680,34 +861,58 @@ void zowelog(LoggingContext *context, uint64 compID, int level, char *formatStri
...
+  if (destination->handler2) {
+    destination->handler2(context,component,destination->data,formatString,argPointer,level,fileName,functionName,lineNumber);
+  } else {
+    destination->handler(context,component,destination->data,formatString,argPointer);
+  }
   va_end(argPointer);
 }
 
-void zowedump(LoggingContext *context, uint64 compID, int level, void *data, int dataSize){
+void zowedumpInner(LoggingContext *context,
```
ifakhrutdinov: This should be static.

ifakhrutdinov (changes requested):
Please see my comments.

And another note: I don't think you have addressed my concerns (in the issue) about the `handler2` being potentially missing in a mixed level environment. Do you think we need to worry about that?

jordanfilteau1995:
@ifakhrutdinov thanks so much for reviewing! I'll get to these :).

c/logging.c
jordanfilteau1995: I think the definition of `tm` is a series of `int`. Do we need to worry about null termination in that case?

Issue #446, opened 2024-05-02
Incorrect check in the recovery facility


ifakhrutdinov:
**Describe the bug**

The recovery facility might free the state cell-pool by accident if an ABEND occurs right after the ESTAEX has been set but before any recovery states with the retry option have been pushed. If this happens, any subsequent PC-ss call will ABEND.

The bug: [this BZ](https://github.com/zowe/zowe-common-c/blob/zss-v2.14.0/c/recovery.c#L500) should be BNZ.

**Steps to Reproduce**

There is no simple way to reproduce this for now.

**Expected behavior**

The recovery code must not free the shared cell-pool.

**Describe your environment**

* Zowe >= v1.9.0
* Zowe >= v2.0.0

PR #448, opened 2024-05-07
Fix an incorrect check in the recovery router (v2)

ifakhrutdinov:
The recovery facility might erroneously free the state cell-pool if an ABEND occurs right after the ESTAEX has been set but before any recovery states with the retry option have been pushed; this issued is caused by an incorrectly coded check in the recovery router code (the check is inverted).

## Proposed changes

The PR fixes the incorrect instruction and adds some code to leave "bread crumbs" when the state cell-pool is deleted.

This PR addresses Issue: #446

## Type of change

## PR Checklist
- [x My changes generate no new warnings

## Testing

There is no reliable way to reproduce this, but in the context of the cross-memory server, one could write a service which disables all the recovery states and then ABENDs on purpose; all the subsequent calls to this server should ABEND without the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## `2.17.0`
 - Fixed `xplatform.loadFileUTF8` when trying to open nonexistent file (#454)
+- Bugfix: fix an incorrect check in the recovery router code which might lead to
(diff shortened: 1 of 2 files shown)


JoeNemo (approved):
Approved.

PR #450, opened 2024-05-09
Allow strings to be null for backward compat on recent bugfix

1000TurquoisePogs:
Prior to v2.16.0 rc1, yaml entries such as `foo: ` were wrongly identified as an empty string when yaml spec says these should be null.
However, some programs were relying upon this in their schema, and upon us fixing the bug, their schemas became invalid when they strictly checked for type=string, when in reality their config was set up so that type could be either string OR null.

This is a temporary fix which allows the same behavior.
It does so by, upon seeing a null, checking if the schema says it should instead be a string.
If so, it will craft an empty string and do a string validation upon it, because it is still possible the user's schema required a string with size greater than 0.

This behavior is toggled by validator->allowStringToBeNull
I suggest removing it in v3, but keeping it for now to not cause disruption.


JoeNemo (approved):
Approved

Issue #463, opened 2024-07-25
Debug option excess spam


nosrednayduj:
**Describe the bug**
If you specify components.zss.logLevels._zss.httpserver: 5  then you get dozens of lines,
tokenEnd=8
tokenEnd=9
....
tokenEnd=956
tokenEnd=957
(or however long your token is)

Looks to me like this line should be moved OUTSIDE the loop!

https://github.com/zowe/zowe-common-c/blob/194623a65f095a2ae57c67c11d79992a0ec0e4a5/c/httpserver.c#L2916

**Steps to Reproduce**

1. enable debug components.zss.logLevels._zss.httpserver: 5
2. hit ZSS with some web requests using token authentication

**Expected behavior**

Only one line of tokenEnd, however long the token is.

**Logs**
see snippet above

**Describe your environment**

- Zowe version number (Check the Desktop login screen, or manifest.json in the Zowe install folder): 2.15.0
- Install method (pax, smpe, kubernetes, github clone): either pax or smpe
- Operating system (z/OS, kubernetes, etc) and OS version: z/OS 3.1
- Node.js version number (Shown in logs, or via `node --version`): N/A
- Java version number (Shown in logs, or via `java -version`): N/A
- z/OSMF version: N/A
- What is the output of log message ZWES1014I:
ZWES1014I ZIS status - 'Ok' (name='IZPS055         ', cmsRC='0', description='Ok', clientVersion='2')
- Environment variables in use: N/A

**Additional context**
Had occasion to enable this debug.  If your token is short it's not too bad (e.g. token 40 chars) but some of the tokens are really long if they come from APIML.

nosrednayduj:
Another place, same problem, end note is inside loop.
https://github.com/zowe/zowe-common-c/blob/194623a65f095a2ae57c67c11d79992a0ec0e4a5/c/httpserver.c#L2849

PR #464, opened 2024-07-25
Minor debug message change - Mix/Max TLS requested

Martin-Zeithaml:
Debug option `components.zss.logLevels._zss.httpserver=5` will produce messages
```
Max TLS requested=3
Min TLS requested=3 
```
The number is index of `TLS_NAMES` array, which is not helpful much.

This PR prints the actual item:

```
Max TLS requested=TLSv1.2     
Min TLS requested=TLSv1.2     
```


Martin-Zeithaml:
Once discussed, I can fix https://github.com/zowe/zowe-common-c/issues/463 here.

1000TurquoisePogs (approved):
Looks like a good change and makes sense to me.

c/tls.c
```diff
@@ -76,21 +76,20 @@ static int isTLSV13Available(TlsSettings *settings) {
...
   "TLSv1.2",
   "TLSv1.3"
 };
 
-#define TLS_NAMES_LENGTH 5
-
 static int getTlsMax(TlsSettings *settings) {
   if (settings->maxTls != NULL) {
-    for (int i = 0; i < TLS_NAMES_LENGTH; i++) {
+    for (int i = 0; i < TLS_NAMES_COUNT; i++) {
```
ifakhrutdinov: Nit: you could do `sizeof(TLS_NAMES) / sizeof(TLS_NAMES[0])` (we should use a macro for this in fact because this is ugly), so that you have a dependency on the thing you want the size of instead of a dependency on a dependency which can change (e.g., someone starts using something else for the size of `TLS_NAMES` and doesn't check if `TLS_NAMES_COUNT` is used somewhere else). Of course, my proposal can also break if someone starts `malloc`ing `TLS_NAMES`, but there seem to be solutions for that too :-)

ifakhrutdinov (approved):

PR #476, opened 2024-08-16
SLHAlloc must return NULL when internal allocations fail

ifakhrutdinov:
## Proposed changes

The SLHAlloc function keeps using the results of the internal allocations even when they're NULL which leads to S0C4 ABENDs. This commit makes SLHAlloc return NULL when the internal allocations fail.

This PR depends upon the following PRs: https://github.com/zowe/zowe-common-c/pull/475

## Type of change
Please delete options that are not relevant.
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Change in a documentation
- [ ] Refactor the code 
- [ ] Chore, repository cleanup, updates the dependencies.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist
Please delete options that are not relevant.
- [x] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [x] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zss/blob/v2.x/master/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made
- [x] Relevant update to CHANGELOG.md
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

## Testing

Write an application which uses SLHAlloc; exhaust storage by reaching MEMLIMIT; SLHAlloc calls must return NULL and not ABEND.

PR #479, opened 2024-08-30
SLHAlloc must return NULL when internal allocations fail (Zowe v3)

ifakhrutdinov:
This is the v3 counterpart of https://github.com/zowe/zowe-common-c/pull/476.


ifakhrutdinov:
Already merged from v2.x by @1000TurquoisePogs, closing.

PR #469, opened 2024-08-05
Returning internal return code through function argument for 'httpClientSessionInit'

Gautham-coder:
## Proposed changes
Function 'httpClientSessionInit' calls other functions inside it and the return codes for those calls are not availabe outside 'httpClientSessionInit'. Adding a return code as argument will solve this requirement.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/468

This PR depends upon the following PRs:

## Type of change
Please delete options that are not relevant.
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Change in a documentation
- [ ] Refactor the code 
- [ ] Chore, repository cleanup, updates the dependencies.
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## PR Checklist
Please delete options that are not relevant.
- [x] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [x] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zss/blob/v2.x/master/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] New and existing unit tests pass locally with my changes
- [ ] video or image is included if visual changes are made
- [x] Relevant update to CHANGELOG.md
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

PR #481, opened 2024-09-04
Revert "Returning internal return code through function argument for 'httpClientSessionInit'"

JoeNemo:
Reverts zowe/zowe-common-c#469.  The original API should have been retained and it can be implemented by calling the new API with a dummy argument for the return code.


ifakhrutdinov (approved):
lgtm

Issue #405, opened 2023-10-02
Module reuse in cross-memory server


ifakhrutdinov:
#### Overview
Implement a COW-like feature for LPA modules (e.g. ZWESIS01, ZWESISDL, plugins). This will reduce the amount of common storage multiple ZIS instances could consume.

PR #492, opened 2024-09-26
Module registry

ifakhrutdinov:
## Proposed changes

To address potential LPA shortages when multiple ZIS instances are running on the same LPAR, a new feature--the module registry--has been introduced.

The module registry is a set of data structures in common storage in which eligible modules are traced. An eligible module is a module marked with a special macro; the mark and the module names are then used to uniquely identify the module. Once ZIS and its plug-in modules are marked, they can be added to the registry and shared: if another ZIS instance needs to load an identical module to the LPA, it will reuse a previously registered (and loaded to the LPA) module instead. Sharing identical modules should dramatically decrease the LPA footprint.

This feature must not be used when an application wants to remove its modules from the LPA (e.g., ZIS does that in dev mode) because it will affect any other applications using the same modules.

This PR addresses Issue: #405

## Type of change

## PR Checklist

## Testing
* Clone this repo and check out the source branch for this change;
* Go to `./tests` and run `build_modreg.sh`;
* Make sure that the load library with the test module is APF-authorized;
* Make sure a ZIS instance has been run since the latest IPL;
* Use the JCL from `./tests/modregtest.c` to run the test program;
* The job must end with RC 0.


JoeNemo (approved):
Approved.

Issue #468, opened 2024-08-05
httpClientSessionInit does not return rc/rsn to callers


Gautham-coder:
httpClientSessionInit has its own set of status values it returns, but the actual rc/rsn of commands it does within are hidden behind compile-time printf statements.

This is unhelpful because this function does fail sometimes, and on zOS the reason is opaque and requires a GSK trace to then discover the cause. It just costs extra debug time then.

Is there any opposition to just add more arguments to httpClientSessionInit to allow passing back rc/rsn?

PR #499, opened 2024-11-08
Adding httpClientSessionInitv2 which through an argument will give the internal return code

Gautham-coder:
## Problem
httpClientSessionInit has its own set of status values it returns, but the actual rc/rsn of commands it does within are hidden behind compile-time printf statements.
This is unhelpful because this function does fail sometimes, and on zOS the reason is opaque and requires a GSK trace to then discover the cause. It just costs extra debug time then.

## Proposed changes
Here we have httpClientSessionInit to  httpClientSessionInitv2 to allow passing back internal rc. The original api still exists and calls the new version by passing a dummy argument.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/468

## Type of change

## PR Checklist


JoeNemo (approved):
Approved.

PR #516, opened 2025-03-03
Bugfix/v2/socket timeout

1000TurquoisePogs:
draft, description soon


c/bpxskt.c
```diff
@@ -370,17 +370,37 @@ Socket *tcpClient3(SocketAddress *socketAddress,
...
 }
 
+Socket *tcpClient3(SocketAddress *socketAddress,
+       int connectTimeoutInMillis,
+       int tlsFlags,
+       int *returnCode, /* errnum */
+       int *reasonCode) { /* errnum - JR's */
+  return tcpClient4(socketAddress,
+                    connectTimeoutInMillis,
+                    0,
```
ifakhrutdinov: Should this be `-1`? The value 0 will cause blocking mode which isn't the current behaviour for `tcpClient3`.

c/bpxskt.c
```diff
@@ -370,17 +370,37 @@ Socket *tcpClient3(SocketAddress *socketAddress,
       *returnCode = 0;
       *reasonCode = 0;
       socket->protocol = IPPROTO_TCP;
+
+      if (readTimeoutInMillis >= 0) {
+        setSocketBlockingMode(socket, TRUE, returnCode, reasonCode);
```
1000TurquoisePogs, line 374: it looks like i also need to do a
`setSocketOption(socket, SO_RCVTIMEO, length, readTimeoutInMillis, returnCode, reasonCode);`

SO_RCVTIMEO implied from EWOULDBLOCK note in https://www.ibm.com/docs/en/zos/2.1.0?topic=descriptions-read-bpx1red-bpx4red-read-from-file-socket
and identified as a valid option within the BPXYSOCK https://www.ibm.com/docs/en/zos/2.1.0?topic=31-bpxysock-map-sockaddr-structure-constants

PR #520, opened 2025-04-07
Add blocking option for httpclient

1000TurquoisePogs:
~~this PR recreates PR https://github.com/zowe/zowe-common-c/pull/516 for v3.~~
It attempts to address:
~~1) The HTTP client code does not block when it should~~
~~2) The HTTP client code does not propagate the code, EWOULDBLOCK, when it's in non-blocking mode, so callers can do little about it.~~
1) The HTTP client code only has a nonblocking mode, there is no blocking mode.

### DIFF
--- a/c/bpxskt.c
+++ b/c/bpxskt.c
@@ -397,7 +397,19 @@ Socket *tcpClient4(SocketAddress *socketAddress,
         writeTimeout.tv_usec = (long)((writeTimeoutInMillis % 1000UL)*1000UL);
 
         setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(struct timeval), (char*)&readTimeout, returnCode, reasonCode);
+        if (returnCode) {
+          if (socketTrace) {
+            printf("Failed to set read timeout for socket. rc=0x%x, rsn=0x%x\n", returnCode, reasonCode);
+          }
+          return NULL;
+        }
         setSocketOption(socket, SOL_SOCKET, SOCK_SO_SNDTIMEO, sizeof(struct timeval), (char*)&writeTimeout, returnCode, reasonCode);
+        if (returnCode) {
+          if (socketTrace) {
+            printf("Failed to set write timeout for socket. rc=0x%x, rsn=0x%x\n", returnCode, reasonCode);
+          }
+          return NULL;
+        }
         setSocketBlockingMode(socket, TRUE, returnCode, reasonCode);
       } else {
         setSocketBlockingMode(socket, FALSE, returnCode, reasonCode);
--- a/c/bpxskt.c
+++ b/c/bpxskt.c
@@ -14,10 +14,12 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <sys/time.h>
 #include "zowetypes.h"
 #include "alloc.h"
 #include "utils.h"
 #include "bpxnet.h"
+
 
 #ifdef USE_RS_SSL
 #include "rs_ssl.h"
@@ -222,6 +224,7 @@ int getSocketDebugID(Socket *s){
 Socket *tcpClient4(SocketAddress *socketAddress,
          int connectTimeoutInMillis,
          int readTimeoutInMillis,
+         int writeTimeoutInMillis,
          int tlsFlags,
          int *returnCode,
          int *reasonCode) {
@@ -369,8 +372,32 @@ Socket *tcpClient4(SocketAddress *socketAddress,
       /* manual says returnCode and value only meaningful if return value = -1 */
       *returnCode = 0;
       *reasonCode = 0;
+
+      if (readTimeoutInMillis > 0 && writeTimeoutInMillis <= 0) {
+        writeTimeoutInMillis = readTimeoutInMillis;
+        if (socketTrace){
+          printf("Cannot set write timeout without read timeout. Read timeout value set to write timeout value\n");
+        }
+      }
+      if (writeTimeoutInMillis > 0 && readTimeoutInMillis <= 0) {
+        readTimeoutInMillis = writeTimeoutInMillis;
+        if (socketTrace){
+          printf("Cannot set read timeout without write timeout. Write timeout value set to read timeout value\n");
+        }
+      }
+
       if (readTimeoutInMillis >= 0) {
-        setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(int), (char*)&readTimeoutInMillis, returnCode, reasonCode);
+        struct timeval readTimeout;
+        struct timeval writeTimeout;
+
+        readTimeout.tv_sec = (long)(readTimeoutInMillis / 1000UL);
+        readTimeout.tv_usec = (long)((readTimeoutInMillis % 1000UL)*1000UL);
+
+        writeTimeout.tv_sec = (long)(writeTimeoutInMillis / 1000UL);
+        writeTimeout.tv_usec = (long)((writeTimeoutInMillis % 1000UL)*1000UL);
+
+        setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(struct timeval), (char*)&readTimeout, returnCode, reasonCode);
+        setSocketOption(socket, SOL_SOCKET, SOCK_SO_SNDTIMEO, sizeof(struct timeval), (char*)&writeTimeout, returnCode, reasonCode);
         setSocketBlockingMode(socket, TRUE, returnCode, reasonCode);
       } else {
         setSocketBlockingMode(socket, FALSE, returnCode, reasonCode);
@@ -390,6 +417,7 @@ Socket *tcpClient3(SocketAddress *socketAddress,
   return tcpClient4(socketAddress,
                     connectTimeoutInMillis,
                     -1,
+                    -1,
                     tlsFlags,
                     returnCode,
                     reasonCode);
@@ -401,6 +429,7 @@ Socket *tcpClient2(SocketAddress *socketAddress,
        int *reasonCode) { /* errnum - JR's */
   return tcpClient4(socketAddress,
                     connectTimeoutInMillis,
+                    -1,
                     -1,
                     0,
                     returnCode,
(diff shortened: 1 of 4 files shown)


c/bpxskt.c
```diff
@@ -369,18 +369,39 @@ Socket *tcpClient3(SocketAddress *socketAddress,
       /* manual says returnCode and value only meaningful if return value = -1 */
       *returnCode = 0;
       *reasonCode = 0;
+      if (readTimeoutInMillis >= 0) {
+        setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(int), (char*)&readTimeoutInMillis, returnCode, reasonCode);
```
ifakhrutdinov: `SOCK_SO_RCVTIMEO` accepts the `timeval` struct which is 8 bytes. Are we sure this actually works? Does the call return 0?

ifakhrutdinov: If this call fails and the next (set blocking mode) succeeds, zss might stuck waiting for data forever.

c/bpxskt.c
```diff
@@ -369,18 +369,39 @@ Socket *tcpClient3(SocketAddress *socketAddress,
...
+        setSocketBlockingMode(socket, TRUE, returnCode, reasonCode);
```
ifakhrutdinov, line 413: You may also want to set a non-zero send time-out because this blocking mode may lead to that zss thread hanging forever, waiting for the peer socket to become available for writes.

c/httpclient.c
```diff
@@ -1049,6 +1053,9 @@ int httpClientSessionReceiveNativeLoop(HttpClientContext *ctx, HttpClientSession
       bytesRead = socketRead(session->socket, buf, sizeof(buf), &bpxrc, &bpxrsn);
       if (bytesRead < 1) {
         HTTP_CLIENT_TRACE_VERBOSE("http client nativeLoop socket read error rc=%d, rsn=0x%x\n", bpxrc, bpxrsn);
+        if (bpxrc == 0x44E){
```
ifakhrutdinov, line 1056: I think we want to check for the time-out as well, otherwise it'll be unclear if this call has succeeded or timed out.

Something like this maybe:
```c
      if (rc == EWOULDBLOCK && rsn == JRTIMEOUT) {
        return RC_SEND_TIMEOUT;
      }
```

c/httpclient.c
```diff
@@ -1093,6 +1100,9 @@ int httpClientSessionReceiveNative(HttpClientContext *ctx, HttpClientSession *se
     buf = SLHAlloc(session->slh, maxlen);
     buflen = socketRead(session->socket, buf, maxlen, &bpxrc, &bpxrsn);
     if (buflen < 1) {
+      if (bpxrc == 0x44E){
```
ifakhrutdinov: I believe the recommendation is to also check `EAGAIN` (which is 0x70 if I'm not mistaken).

c/bpxskt.c
ifakhrutdinov: I've performed a stand-alone test where I passed 4 bytes to `BPX1OPT` (this is what `setSocketOption` uses) and got `rv=-1, rc=121, rsn=0x11107315` which is:
```
JRInvalidOptLength: The option length specified is not valid

Action: The value specified for the option length is not valid. Re-issue the
socket call with a valid option length.
```

ifakhrutdinov (changes requested):
@1000TurquoisePogs, please see my comments. We may need to address the findings.

c/httpclient.c
1000TurquoisePogs: yes. i noticed errno.h exists to make this easier, but no such header exists for metal.
i think we should make one but not here.

c/bpxskt.c
```diff
@@ -369,8 +372,32 @@ Socket *tcpClient4(SocketAddress *socketAddress,
       /* manual says returnCode and value only meaningful if return value = -1 */
       *returnCode = 0;
       *reasonCode = 0;
+
+      if (readTimeoutInMillis > 0 && writeTimeoutInMillis <= 0) {
+        writeTimeoutInMillis = readTimeoutInMillis;
+        if (socketTrace){
```
ifakhrutdinov, line 378: I think you can. I didn't suggest they were related, it's just one usually wants to adjust both.

I'd do something like this:
```c
readTimeoutInMillis = readTimeoutInMillis >= 0 ? readTimeoutInMillis : DEFAULT_RECV_TIMEOUT_MS;
writeTimeoutInMillis = writeTimeoutInMillis >= 0 ? writeTimeoutInMillis : DEFAULT_SEND_TIMEOUT_MS;
```

c/bpxskt.c
```diff
@@ -369,8 +372,32 @@ Socket *tcpClient4(SocketAddress *socketAddress,
...
+          printf("Cannot set read timeout without write timeout. Write timeout value set to read timeout value\n");
+        }
+      }
+
       if (readTimeoutInMillis >= 0) {
-        setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(int), (char*)&readTimeoutInMillis, returnCode, reasonCode);
+        struct timeval readTimeout;
+        struct timeval writeTimeout;
+
+        readTimeout.tv_sec = (long)(readTimeoutInMillis / 1000UL);
```
ifakhrutdinov: Not sure you need to case to `long` if you already have one `long` operand.

c/bpxskt.c
```diff
@@ -369,8 +372,32 @@ Socket *tcpClient4(SocketAddress *socketAddress,
...
+        readTimeout.tv_usec = (long)((readTimeoutInMillis % 1000UL)*1000UL);
+
+        writeTimeout.tv_sec = (long)(writeTimeoutInMillis / 1000UL);
+        writeTimeout.tv_usec = (long)((writeTimeoutInMillis % 1000UL)*1000UL);
+
+        setSocketOption(socket, SOL_SOCKET, SOCK_SO_RCVTIMEO, sizeof(struct timeval), (char*)&readTimeout, returnCode, reasonCode);
```
ifakhrutdinov, line 399: `setSocketOption` can fail.

c/httpclient.c
```diff
@@ -1053,10 +1054,13 @@ int httpClientSessionReceiveNativeLoop(HttpClientContext *ctx, HttpClientSession
       bytesRead = socketRead(session->socket, buf, sizeof(buf), &bpxrc, &bpxrsn);
       if (bytesRead < 1) {
         HTTP_CLIENT_TRACE_VERBOSE("http client nativeLoop socket read error rc=%d, rsn=0x%x\n", bpxrc, bpxrsn);
-        if (bpxrc == 0x44E){
+        if (bpxrc == 0x44E) {
           sts = HTTP_CLIENT_EWOULDBLOCK;
+        } else if (bpxrc == 0x70) {
+          sts = HTTP_CLIENT_EAGAIN;
```
ifakhrutdinov: As far as I know, these two errors mean virtually the same and they're usually checked together. At the HTTP client level, exposing these is too low-level; I'd just check both (EWOULDBLOCK or EAGAIN) and return a common HTTP_CLIENT RC.

Having said that, I wonder, how do we want to handle this condition? Retry or disconnect from the HTTP server?

In one of our products we check the combination `rc == EWOULDBLOCK && rsn == JRTIMEOUT` to tell the caller that's the timeout has been reached; any other combination of the return codes `EAGAIN`/`EWOULDBLOCK` and their reason codes results in a generic error. This is done with raw sockets, so I feel like for HTTP, you'd just want to reconnect because otherwise, for a retry, you have to track how much bytes have been sent (e.g., what if we've sent a half of an HTTP request and got a timeout?).

c/bpxskt.c
1000TurquoisePogs: i am putting this limitation here whether or not the code is capable, because it seems weird to not set both.

c/httpclient.c
1000TurquoisePogs: can EWOULDBLOCK even happen when blocking is enabled?
if blocking is not enabled, EWOULDBLOCK can indicate its just not ready yet.
this code is already handled - you just try again. zss is already doing that. it detects HTTP_CLIENT_EWOULDBLOCK and just tries again.

if blocking is enabled, then this is more interesting but again it seems up to the user of the httpclient to decide what to do. i dont see what additional work should be done in this PR for that error handling.

c/bpxskt.c
1000TurquoisePogs: is it bad? i saw this logic in timeutils.c

c/bpxskt.c
ifakhrutdinov: Not sure if this is weird. In one of our product, we have a non-zero default timeout for receive operations but a zero default timeout (i.e., the call block indefinitely) for send operations. The justification, I think, is that we can tolerate a "slow" peer when we send, but once sent, we expect a relatively quick response because a slow response would indicate issues/incompatibility in our case. Timeout which depend on each other look unintuitive.

c/bpxskt.c
ifakhrutdinov: It does nothing, but it also shows the author might not have known how implicit conversion works in C (the rules aren't simple I must admit).

* First, via a usual arithmetic conversion, `readTimeoutInMillis` is converted to `unsigned long` (BTW, I think `time_t` is `long`, why `UL` and not `L`?)
* Then, even without the cast, the result is converted to `long` since `tv_sec` is `long` via an assignment conversion

I think you can drop all the `UL` and casts. 

For example:
```c
readTimeout.tv_sec =  readTimeoutInMillis / 1000;
readTimeout.tv_usec = (readTimeoutInMillis % 1000) * 1000;
```

c/httpclient.c
ifakhrutdinov: > can EWOULDBLOCK even happen when blocking is enabled?

Yes, along with a special reason code (`JRTIMEOUT`), you'll get it when the timeout has been reached.

c/bpxskt.c
1000TurquoisePogs: Yes, time_t is long - looked it up.

c/httpclient.c
1000TurquoisePogs: So, added detect of JRTIMEOUT, and zss will see it as "HTTP_CLIENT_SOCKET_TIMEOUT" and decide to abort its persuit of data on that socket, and try again next time interval.

1000TurquoisePogs:
I have put this PR to draft status for whoever is looking to work on it.
I am no longer interested in it. I do not need blocking to achieve the use case it was originally intended for within ZSS, and discussions of that lead to delay of fixes for ZSS. I have separated out the content useful to my needs in a separate PR to move work forward.

PR #524, opened 2025-05-02
add configmgr api `copyConfigurationAndDeleteKey`

MarkAckert:
## Proposed changes
**Note**: Ready for review.

This PR adds a new `copyConfigurationAndDeleteKey` API to configmgr which can be used to delete keys from a YAML document. e.g., deleting `zowe.setup` or `zowe.setup.dataset` is now possible. The `copyConfigurationAndDeleteKey` API requires that a new configuration name be supplied, so the mutation happens in a **new copy** of the configuration (matching other update APIs). This is required as part of the Node.JS removal from zwe scripts, and is last step to fully replace the `config-converter` component we use from zowe-install-packaging-tools. 

The existing modification APIs were not used after experimentation, as the modification / overlay is a union style merge operation rather than a replacement. While that could've been enhanced to support deleting keys, it felt better to have a clean line of separation between the 2 behaviors.

Limitations in regex made parsing of keys containing multiple periods less readable than I wanted, so the code has additional comments in the parsing section.

This PR addresses Issue: Part of https://github.com/zowe/zowe-install-packaging/pull/4307

## Type of change
- [ ] Refactor the code 

## PR Checklist

## Testing
Added unit tests in `tests/js/config1.js`.

## Further comments
Supports:
* Deleting composite array and object keys. e.g., both `zowe.externalHosts` and `zowe.externalHosts[0]` are supported (and yield different outputs).
* Deleting composite keys where a key entry contains periods. Syntax: `zowe.[_zsf.logging.level]`. Works for arrays too: `zowe.[_demo.key[0]].[another.key]`
```yaml
zowe:
  _zsf.logging.level: TRACE
  _demo.key:
    - another.key: SOMETHING
```

* If the delete key supplied does not exist, this is a no-op, and returns the document as-is. This matches the config-converter behavior.

Notes:
* If you delete the only property of an object, the render outputs curly braces. e.g.:
```yaml
zowe:
  a:
```
 Running `delete "zowe.a"` gives:
 ```yaml
zowe: {}
```
This is legal YAML, but might be confusing if a user comes across it? I don't know how to change the render behavior.

* Keys with `[` in position one will not parse as keys, but instead as a key containing periods, which means they will not match and the resulting output will be a no-op. I think this is an acceptable limitation until we have a use case which needs `[` in position one. This might be worked around by: `[[key]`, but is not tested.

### DIFF
--- a/c/configmgr.c
+++ b/c/configmgr.c
@@ -858,6 +858,38 @@ static int overloadConfiguration(ConfigManager *mgr,
   }
 }
 
+int cfgDeleteFromConfiguration(ConfigManager* mgr, 
+         const char *configName,
+         const char *modifiedConfigName,
+         const char *keyToDelete) {
+
+  CFGConfig *config = getConfig(mgr, configName);
+  if (!config) {
+    return ZCFG_UNKNOWN_CONFIG_NAME;
+  }
+  int deleteStatus = 0;
+
+  Json *modifiedData = jsonDelete(mgr->slh, 
+            keyToDelete, 
+            config->configData, 
+            &deleteStatus);
+  
+  if (deleteStatus) {
+    return deleteStatus;
+  }
+
+  CFGConfig *modifiedConfig = cfgAddConfig(mgr,modifiedConfigName);
+  modifiedConfig->schemaPath = config->schemaPath;
+  /* is this path really true anymore?? */
+  modifiedConfig->configPath = config->configPath;
+  modifiedConfig->topSchema = config->topSchema;
+  modifiedConfig->otherSchemas = config->otherSchemas;
+  modifiedConfig->otherSchemasCount = config->otherSchemasCount;
+  modifiedConfig->parmlibMemberName = config->parmlibMemberName;
+  modifiedConfig->configData = modifiedData;
+  return ZCFG_SUCCESS;
+}
+
 int cfgMakeModifiedConfiguration(ConfigManager *mgr, 
 				 const char *configName, 
 				 const char *modifiedConfigName,
--- a/c/configmgr.c
+++ b/c/configmgr.c
@@ -1702,7 +1702,7 @@ int cfgDeleteFromConfiguration(ConfigManager* mgr,
          const char *modifiedConfigName,
          const char *keyToDelete) {
    */
-  EJSNativeMethod *deleteFromConfiguration = ejsMakeNativeMethod(ejs,configmgr,"deleteFromConfiguration",
+  EJSNativeMethod *deleteFromConfiguration = ejsMakeNativeMethod(ejs,configmgr,"copyConfigurationAndDeleteKey",
     EJS_NATIVE_TYPE_INT32,
     (EJSForeignFunction*)deleteFromConfigurationWrapper);
   ejsAddMethodArg(ejs,deleteFromConfiguration,"configName",EJS_NATIVE_TYPE_CONST_STRING);
--- a/c/configmgr.c
+++ b/c/configmgr.c
@@ -867,6 +867,7 @@ static int overloadConfiguration(ConfigManager *mgr,
   }
 }
 
+// This method always succeeds. If the keyToDelete is not found, configuration is returned with modification.
 int cfgDeleteFromConfiguration(ConfigManager* mgr, 
          const char *configName,
          const char *modifiedConfigName,
@@ -876,20 +877,13 @@ int cfgDeleteFromConfiguration(ConfigManager* mgr,
   if (!config) {
     return ZCFG_UNKNOWN_CONFIG_NAME;
   }
-  int deleteStatus = 0;
 
   Json *modifiedData = jsonDelete(mgr->slh, 
             keyToDelete, 
-            config->configData, 
-            &deleteStatus);
+            config->configData);
   
-  if (deleteStatus) {
-    return deleteStatus;
-  }
-
   CFGConfig *modifiedConfig = cfgAddConfig(mgr,modifiedConfigName);
   modifiedConfig->schemaPath = config->schemaPath;
-  /* is this path really true anymore?? */
   modifiedConfig->configPath = config->configPath;
   modifiedConfig->topSchema = config->topSchema;
   modifiedConfig->otherSchemas = config->otherSchemas;
--- a/c/configmgr.c
+++ b/c/configmgr.c
@@ -1516,6 +1516,19 @@ static int loadConfigurationWrapper(ConfigManager *mgr, EJSNativeInvocation *inv
   return EJS_OK;
 }
 
+static int deleteFromConfigurationWrapper(ConfigManager *mgr, EJSNativeInvocation *invocation){
+  const char *configName = NULL;
+  ejsStringArg(invocation,0,&configName);
+  const char *modifiedConfigName = NULL;
+  ejsStringArg(invocation,1,&modifiedConfigName);
+  const char *keyToDelete = NULL;
+  ejsStringArg(invocation,2,&keyToDelete);
+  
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by JoeNemo** [comment]
Is this meant to be a persistent change in yaml?   Or is this only an in-memory change for the program that has integrated configmgr?

**Comment by MarkAckert** [comment]
This is an in-memory change only; the caller is responsible for persisting changes to disk. I followed the structure of `updateZoweYaml` / `makeModifiedConfiguration`. 

These lines in the zowe-install-packaging PR shows the call pattern: https://github.com/zowe/zowe-install-packaging/blob/5726c3792404e72ead99468fe37621348d408e7d/bin/libs/configmgr.ts#L425-L429 , where `deleteConfig` is a wrapper around the configmgr API. The caller modifies the YAML in-memory, then issues a separate call to write out the changes.

**Comment by JoeNemo** [comment]
There really are no "destructive" operations in configmgr to prevent bugs where one consumer/user of the data is not in sync with another.   Can we re-conceive this as making a copy of the config under a new name with the following edits applied.

**Comment by 1000TurquoisePogs** [comment]
You may find https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/commands/internal/config/get/index.ts  `zwe internal config set` to be an alternative to achieving similar goals. What it does, instead of "deleting", is it reads the existing config, makes an edit, places the edit into a location, and then says to configmgr "now load name_of_config_revision_1" and points our TS code to that as the current config variable.

**Comment by MarkAckert** [comment]
@JoeNemo This does currently create a copy of the configuration before taking destructive actions; a user always needs to call the API with the next revision name. I'll join the squad call next week to answer other questions that may come up

File: c/json.c [hunk 1]
```diff
@@ -2632,6 +2696,97 @@ static void copyJson(JsonBuilder *builder, Json *parent, char *parentKey, Json *
   }
 }
 
+static void deleteJson(Json *base, const char *deleteKey) {
+  if (deleteKey) {
+    Json *parentNode = NULL;
+    Json *activeNode = base;
+    char* copiedKey = strdup(deleteKey);
+    char workStr[500]; // huge working buffer on the stack for potential keys, all should realistically be <50 chars
```
Martin-Zeithaml: According to [this](https://github.com/zowe/zowe-common-c/blob/5d4f38341119b5e1406fd4ac110c6c74b7472382/c/yaml2json.c#L357), 256 should be fine. The best approach is to use the same `#define` everywhere.

c/json.c
```diff
@@ -2632,6 +2696,97 @@ static void copyJson(JsonBuilder *builder, Json *parent, char *parentKey, Json *
...
+          if (arrayIdx < nodeArray->count) {
+            tokenMatchFound = true;
+            parentNode = activeNode;
+            activeNode = nodeArray->elements[arrayIdx];
+            arrayIdx=-1; // reset
+          }
+        }
+      } else {
+        // If we have format like [a.b.c], ensure the jsonTok is [a.b.c], and not [a.b.c
+        memset(workStr, 0x00, sizeof(char)*500);
```
Martin-Zeithaml: Same as previous [comment](https://github.com/zowe/zowe-common-c/pull/524/files#r2108432695).

c/configmgr.c
```diff
@@ -867,6 +867,38 @@ static int overloadConfiguration(ConfigManager *mgr,
...
+  CFGConfig *modifiedConfig = cfgAddConfig(mgr,modifiedConfigName);
+  modifiedConfig->schemaPath = config->schemaPath;
+  /* is this path really true anymore?? */
+  modifiedConfig->configPath = config->configPath;
+  modifiedConfig->topSchema = config->topSchema;
+  modifiedConfig->otherSchemas = config->otherSchemas;
+  modifiedConfig->otherSchemasCount = config->otherSchemasCount;
+  modifiedConfig->parmlibMemberName = config->parmlibMemberName;
+  modifiedConfig->configData = modifiedData;
+  return ZCFG_SUCCESS;
```
Martin-Zeithaml: `jsonDelete` seems **not** to update `deleteStatus` at all, this block always ends with success.

c/json.c
MarkAckert: Moved the `#define MAX_JSON_KEY` to `json.h` and will use it here.

c/configmgr.c
MarkAckert: Removed the status from the call chain, as the intended design is to always return success.

c/json.c
```diff
@@ -2632,6 +2696,97 @@ static void copyJson(JsonBuilder *builder, Json *parent, char *parentKey, Json *
...
+
+        // use regexp to parse jsonToks of formats: "a[0]", "a.b.c[213132]". Invalid arrays like "abc]", "abc[a2]" are left untouched.
+        if (jsonTok[strlen(jsonTok)-1] == ']') {
+          int rrc = 0;
+          regmatch_t matches[4];
+          regexComp(argPattern,pat,REG_EXTENDED);
+          regexExec(argPattern,jsonTok,4,matches,0);
+          // if we have a matched array index, set the arrayIdx, otherwise, leave jsonTok as-is
+          if (matches[3].rm_so!= -1) {
+            arrayIdx = parseInt(jsonTok + matches[3].rm_so, 0, matches[3].rm_eo - matches[3].rm_so);
```
Martin-Zeithaml, line 2758: Space will break it:
```javascript
const delCases = [
  "colors[ 0 ]",
]
```

```c
            printf(">> %s\n", jsonTok);
            printf("   %li\n", matches[3].rm_so);
            printf("   %li\n", matches[3].rm_eo);
            arrayIdx = parseInt(jsonTok + matches[3].rm_so, 0, matches[3].rm_eo - matches[3].rm_so);
```
Will produce:
```
>> colors[ 0 ]
   343734695488
   343738147632
CEE3204S The system detected a protection exception (System Completion Code=0C4).
         From compile unit T800:/git/repos/common.c.deleteJsonApi/build/../c/utils.c at entry point parseInt at statement 57 at compile unit offset +000000003B94ECA2 at entry offset +000000000000005A at address 000000003B94ECA2.
```

`parseInt` is not the best function, however the abend is caused those huge numbers are used as index.

c/json.c
```diff
@@ -2632,6 +2696,97 @@ static void copyJson(JsonBuilder *builder, Json *parent, char *parentKey, Json *
...
+            memmove(workStr, workStr + 1, strlen(workStr) - 1);
+            workStr[strlen(workStr) - 2] = '\0';
+          }
+          jsonTok = workStr;
+        }
+        //printf("jsonTokPostGroup: %s\n", jsonTok); fflush(stdout);
+
+        // use regexp to parse jsonToks of formats: "a[0]", "a.b.c[213132]". Invalid arrays like "abc]", "abc[a2]" are left untouched.
+        if (jsonTok[strlen(jsonTok)-1] == ']') {
+          int rrc = 0;
```
Martin-Zeithaml: Unused variable.

c/json.c
MarkAckert: Addressed. Properly checking the `rc` from `regexExec` with an additional defensive check that `rm_so` < `rm_eo`, which should always be true for the array regex when `rc == 0`. 

In the example case here, the solution treats `colors [ 0 ]`  as a key rather than an array + index. I think this is OK, as we can say valid array index access requires no spaces between the square brackets.

c/json.c
MarkAckert: Addressed in solution to `colors[ 0 ]`.

Martin-Zeithaml (approved):

1000TurquoisePogs:
As review and discussion has been complete and marked approved, I'm going to merge this.

PR #525, opened 2025-05-06
`configmgr extract`: no new line for the last array item

Martin-Zeithaml:
`configmgr extract` feature is printing new line character for each key found. But for the array, it prints new line char twice, which is might be confusing for arrays with empty strings.

```yaml
test:
  array1:
    - 'apple'
    - 'banana'
    - 'kiwi'
    - ''
```

## Without fix
There are 2 empty lines at the end, which might be considered as 2 empty items:
```
apple
banana
kiwi


```

## With fix
```
apple
banana
kiwi

```

## Empty item, not last and with fix
```yaml
test:
  array1:
    - 'apple'
    - 'banana'
    - 'kiwi'
    - ''
    - 'plum'
```

```
apple
banana
kiwi

plum
```


JoeNemo (approved):
Approved.

JoeNemo (approved):
Approved.

PR #533, opened 2025-06-17
Fix unhandled EWOULDBLOCK on httpclient

1000TurquoisePogs:
This PR adds status codes HTTP_CLIENT_UNBLOCKED_TRY_AGAIN (19) and HTTP_CLIENT_SOCKET_TIMEOUT (20) to httpclient.
httpclient is a non-blocking client, so it is up to users to implement wait-retry logic, yet codes such as EWOULDBLOCK were unhandled before, so there was no way to do this.

This appears to have been a bug that would prevent all use of httpclient, except that in the case of GSK, either it's blocking or introduces just enough delay that EWOULDBLOCK is uncommon.

This bug became apparent when using ATTLS, in which GSK is not present within Zowe itself. In this case, the socket read can happen faster than data is available, and therefore clients need to handle EWOULDBLOCK somehow.

This is accomplished by assigning such EWOULDBLOCK and errors like it (EAGAIN, JRTimeOut) to HTTP_CLIENT_UNBLOCKED_TRY_AGAIN and HTTP_CLIENT_SOCKET_TIMEOUT.

This is an fix to functions httpClientSessionReceiveNativeLoop and httpClientSessionReceiveNative

### DIFF
--- a/c/httpclient.c
+++ b/c/httpclient.c
@@ -1049,7 +1049,17 @@ int httpClientSessionReceiveNativeLoop(HttpClientContext *ctx, HttpClientSession
       bytesRead = socketRead(session->socket, buf, sizeof(buf), &bpxrc, &bpxrsn);
       if (bytesRead < 1) {
         HTTP_CLIENT_TRACE_VERBOSE("http client nativeLoop socket read error rc=%d, rsn=0x%x\n", bpxrc, bpxrsn);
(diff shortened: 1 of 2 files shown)


ifakhrutdinov (approved):
Same as for v2, I really don't like the RC names, but this is a good enough fix for a bad bug.

PR #534, opened 2025-06-17
Fix unhandled EWOULDBLOCK on httpclient (v2)

1000TurquoisePogs:
See https://github.com/zowe/zowe-common-c/pull/533 for description - this is the same but for v2.


ifakhrutdinov (approved):
I really don't like the RC names, but this is a good enough fix for a bad bug.

PR #537, opened 2025-07-17
Update le.c for next zos version

1000TurquoisePogs:
This PR updates the version number check for common such that "ZWE_zowe_launcher_unsafeDisableZosVersionCheck" is not needed to run on the upcoming zos.


JoeNemo (approved):
Approved.

PR #539, opened 2025-07-21
IPv6 enhancements for bpx/mtl socket

1000TurquoisePogs:
This PR adds new functions for handling ipv6 ips. It is backwards compatible in that functions like getAddressByName() are unchanged, calling getAddressByName2() which has a boolean for declaring whether to consider ipv6 or not.


JoeNemo (approved):
Approved.

PR #540, opened 2025-07-23
Fix memory leak in safeFree64Internal

Gautham-coder:
Related to issue: https://github.com/zowe/zss/issues/777

This bug was found when uploading large Unix file contents in chunks. With every chunk, the content memory was allocated SLH. But after processing it was not released and memory kept increasing with every chunk.

This free() stops the leak for every chunk.

Before fix:
======
Here the PRIVATE-64 will keep increasing with every chunk
Before_fix

After Fix:
======
Will increase only initially to create the working memory if you can call it like that. Because this memory will be reused for the upcoming chunks.
After_upload

Note:
====
The current behavior is such that we don't see increase the memory till the chunk size is equal to or less than the one used initially. That is, it is advisable to use the same chunk size for all the chunks.

### DIFF
--- a/c/alloc.c
+++ b/c/alloc.c
@@ -601,7 +601,7 @@ static void safeFree64Internal(char *data, int size, long long token){
 #if defined(METTLE) && defined(_LP64)
   freemain64(data,NULL,NULL);
 #else
-  /* do nothing - because 64 bit allocation in LE is not ready */
+  free(data);
 #endif
 }
 
--- a/c/alloc.c
+++ b/c/alloc.c
@@ -600,8 +600,12 @@ static void safeFree64Internal(char *data, int size, long long token){
 
 #if defined(METTLE) && defined(_LP64)
   freemain64(data,NULL,NULL);
-#else
+#elif defined(_LP64) /* LE case */
   free(data);
+#elif defined(_MSC_VER) && defined(_M_X64) /* Windows 64 case */
+  free(data);
+#else
+  /* Do nothing - because safeMalloc64Internal returns NULL for 31-bit */
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


c/alloc.c
```diff
@@ -601,7 +601,7 @@ static void safeFree64Internal(char *data, int size, long long token){
 #if defined(METTLE) && defined(_LP64)
   freemain64(data,NULL,NULL);
 #else
-  /* do nothing - because 64 bit allocation in LE is not ready */
+  free(data);
```
ifakhrutdinov, line 606: @Gautham-coder , please check `safeMalloc64Internal` (also, in the snippet below); it handles this a bit differently, so we should probably do something similar.
```c
#if defined(METTLE) && defined(_LP64)
  res = getmain64((long long)sizeInMegabytes,NULL,NULL); 
#elif defined(_LP64)   /* LE case */
  res = malloc(size);   /* According to Redpaper http://www.redbooks.ibm.com/redpapers/pdfs/redp9110.pdf allocated above bar */
#elif defined(_MSC_VER) && defined(_M_X64) /* Windows 64 case */
  res = malloc(size);
#else
  res = NULL;  /* Invalid if not compiled _LP64 */
#endif
```
Perhaps, add similar `#elif`s with `free()` calls. And for 31-bit, well, do nothing I guess because the `safeMalloc64Internal` returns `NULL` in that case.

Gautham-coder: Hi @ifakhrutdinov, I've added the `#elif`s similar `safeMalloc64Internal `.
Thanks.

ifakhrutdinov (approved):

PR #545, opened 2025-08-07
Protect against rexx variable changes in addConfig

1000TurquoisePogs:
A problem was identified when https://github.com/zowe/zowe-common-c/pull/540 was added that `zwegener` and `zwegen00` would fail when using `cfgrxcfg`.
I now think that PR 540 had nothing to do with it.
I traced with @JoeNemo and found that the REXX calls to `addConfig(name)` and `getConfig(name)` would have the pointer of name having contents change.
I did not investigate why this is, such as if there was a change in the REXX code of `zwegen00`, or if rexx does mysterious things, but to insulate configmgr from that, instead of holding onto the pointer, this creates a copy of the string so that configmgr is the owner of that string.

### DIFF
--- a/c/configmgr.c
+++ b/c/configmgr.c
@@ -763,7 +763,9 @@ CFGConfig *cfgAddConfig(ConfigManager *mgr, const char *configName){
   }
   CFGConfig *newConfig = (CFGConfig*)safeMalloc(sizeof(CFGConfig),"CFGConfig");
   memset(newConfig,0,sizeof(CFGConfig));
-  newConfig->name = configName;
+  char *copyName = safeMalloc(strlen(configName)+1, "configName");
+  strcpy(copyName, configName);
+  newConfig->name = copyName;
   if (mgr->firstConfig){
     mgr->lastConfig->next = newConfig;
     mgr->lastConfig = newConfig;


c/configmgr.c
```diff
@@ -763,7 +763,9 @@ CFGConfig *cfgAddConfig(ConfigManager *mgr, const char *configName){
   }
   CFGConfig *newConfig = (CFGConfig*)safeMalloc(sizeof(CFGConfig),"CFGConfig");
   memset(newConfig,0,sizeof(CFGConfig));
-  newConfig->name = configName;
+  char *copyName = safeMalloc(strlen(configName)+1, "configName");
```
Martin-Zeithaml, line 766: In such case, we should include the reverted #540 in here and update changelog like "Various memory leaks fixed".
Probably best if @Gautham-coder will do it to keep the track of authorship.

I would not blame rexx, it fails at the beginning of the processing, which was almost unchanged.
I have double checked that, it was working without #540 and failing with #540. Mystery.

And nice to have all in 3.3.0.

Martin-Zeithaml:
**This branch**
```
/zowe-common-c/build: git st
On branch bugfix/v3/add-config-rexx
Your branch is up-to-date with 'origin/bugfix/v3/add-config-rexx'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   ../c/alloc.c
```

**With #540 fix**
```diff
--- a/c/alloc.c
+++ b/c/alloc.c
@@ -600,8 +600,12 @@ static void safeFree64Internal(char *data, int size, long long token){

 #if defined(METTLE) && defined(_LP64)
   freemain64(data,NULL,NULL);
+#elif defined(_LP64) /* LE case */
+  free(data);
+#elif defined(_MSC_VER) && defined(_M_X64) /* Windows 64 case */
+  free(data);
 #else
-  /* do nothing - because 64 bit allocation in LE is not ready */
+  /* Do nothing - because safeMalloc64Internal returns NULL for 31-bit */
 #endif
 }
```

**Seems to be working fine**
`ZWEGENER SYSTSPRT`  looks goods, result RC=0 and templates processed.
```
ConfigMgr is about to add a configuration.                                        
ConfigMgr added a configuration.                                                  
ConfigMgr is about to set trace level to 5.                                       
ConfigMgr set trace level to 5.                                                   
ConfigMgr is about to load your schemas.                                          
ConfigMgr loaded your schemas.                                                    
ConfigMgr is about to process your configuration.                                 
ConfigMgr has processed your configuration.                                       
ConfigMgr is about to load your configuration.                                    
ConfigMgr has loaded your configuration.                                          
ConfigMgr is about to validate your configuration.                                
ConfigMgr has validated your configuration.                                       
ConfigMgr is about to get configuration data.                                     
ConfigMgr got configuration data.                                                 
Checking required configuration keys                                              
...
```

c/configmgr.c
1000TurquoisePogs: I made this PR in my sandbox that had #540 included
So, yes I don't know why it makes a difference but this does work around that.

JoeNemo (approved):
There's a tiny low-frequency leak here, but we will need to fix later.

c/configmgr.c
ifakhrutdinov: I think the code that causes that issue is this:
```c
    char *argCopy = SLHAlloc(argHeap,arg->length+1);
    copiedArgs[index] = argCopy;
    memcpy(argCopy,arg->address,arg->length);
    argCopy[arg->length] = 0;
    if (mgr->traceLevel >= 1){
      fprintf(out,"--- Arg %d of args=0x%p and arg=0x%p--- \n",index,args,arg);
      dumpBufferToStream((char*)arg,8,out);
      fprintf(out,"arg data\n");
      dumpBufferToStream((char*)arg->address,arg->length,out);
    }
    index++;
  } while (index < invocation->argCount && index < 30);
  char *arg0 = copiedArgs[0];
  int status = 0;
  if (!strcmp(arg0,"setTraceLevel")){
    status = rexxSetTraceLevel(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"addConfig")){
    status = rexxAddConfig(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"setConfigPath")){
    status = rexxSetConfigPath(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"loadSchemas")){
    status = rexxLoadSchemas(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"loadConfiguration")){
    status = rexxLoadConfiguration(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"setParmlibMemberName")){
    status = rexxSetParmlibMemberName(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"getConfigData")){
    status = rexxGetConfigData(mgr,invocation,invocation->argCount,copiedArgs);
  } else if (!strcmp(arg0,"validate")){
    status = rexxValidate(mgr,invocation,invocation->argCount,copiedArgs);
  } else{
    status = ZCFG_UNKNOWN_REXX_FUNCTION;
  }
  SLHFree(argHeap);
```
`rexxAddConfig` gets the ownership of that SLH storage by using it as a config name, but the SLH gets freed before that ownership is dropped.

The config manager functions should have documented how they treated the parameters, but the default behaviour should have been to make a copy of anything we didn't know/control the lifetime of, so this fix is fine. @1000TurquoisePogs , but where do you free that storage now? You seem to have introduced a leak.

PR #548, opened 2025-08-13
Handle returnValue and add rc/rsn to socket trace

1000TurquoisePogs:
To trace https://github.com/zowe/zss/issues/794, I need to see rc/rsn that aren't handled thus far.
This PR prints them in socketTrace
But also, I saw the TBD message about that none of them are even inspected. I added handling of returnValue.


JoeNemo (commented):
I don't think this code is called.  But your intent for this PR was valid and I can't remove a public API mid-release.  But I needed to do some fixes for things that were latent memory errors, not introduced by your change.

PR #560, opened 2025-10-13
IPv6 httpclient

1000TurquoisePogs:
This PR follows on from https://github.com/zowe/zowe-common-c/pull/539 where we added httpserver ipv6 handling, but forgot to add client code. This PR adds that missing stuff.


JoeNemo (approved):
Approved.

PR #569, opened 2026-01-20
Do not terminate with active callers in PC

ifakhrutdinov:
## Proposed changes

This commit pull-request an additional global counter which is incremented when the PC-routines are entered and is decremented when the PC-routines are exited.

Upon termination, the counter is checked; if the counter is non zero, the cross-memory server waits (and checks the counter again if needed) before freeing the server resources and terminating.

This enhancement should prevent situations where the resources are released before the caller is done using them, which often leads to ABENDs inside the PC-routines.

To make sure that concurrent callers do not cause race conditions, the counter is modified with compare-and-swap. Additionally, the server unsets the "ready" flag preventing any new callers from entering the PC routines and then it checks the counter; whereas, callers increment the counter first before checking the "ready" flag; this order should prevent race conditions between the PC prolog and `waitForCallerToExit` code.

Additional changes:
  * New messages for indicating the number of active callers during termination
  * An additional field in the display command's output which show the current number of active callers
  * The register use in the PC-ss and PC-cp prolog and epilog HLASM has been made more consistent between PC-ss and PC-cp
  * The cross-memory global area DSECT has been expanded to match the C structure
  * Mark some global fields volatile, so that, when optimizations are on, access statements don't get rearranged relative to other seemingly unrelated statements. Example are:
     * Reading `callerCount` in the for-loop in `waitFotCallersToExit` (the `callerCount` read can get moved out of the loop);
      * Setting `serverFlags` in `setServerReady` which can get inlined and rearranged relative to the code following the call.

## Type of change

## PR Checklist

## Testing
* Regression testing of ZIS
* Verifying the new feature:
  * From a test client application, invoke a ZIS service in a loop; use multiple threads/processes to hit the PC routine concurrently
  * Terminate ZIS using the `P` operator command
  * You should get the new message `0119I` and the callers should get RC 30 (not ready)

### DIFF
--- a/c/crossmemory.c
+++ b/c/crossmemory.c
@@ -1453,19 +1453,36 @@ LSSPRLG  LARL  10,LSSPRLG          establish addressability using r10   \n\
          USING LSSPRLG,10                                               \n\
          LGR   5,4                 save latent parm as CPOOL uses r4    \n\
          USING PCLPARM,5                                                \n\
-         LLGT  2,PCLPPRM1          load global area from latent parm    \n\
-         USING GLA,2                                                    \n\
+         LLGT  6,PCLPPRM1          load global area from latent parm    \n\
+         USING GLA,6                                                    \n\
+         LLGF  3,GLACCNT           current caller count                 \n\
+LSSINCR  LGR   4,3                 save caller counter for incrementing \n\
+         AGHI  4,1                 increment caller counter             \n\
+         CS    3,4,GLACCNT         compare and swap                     \n\
+         JNZ   LSSINCR             counter has changed already, retry   \n\
+         LLGF  3,GLASFLG           load server flags                    \n\
+         TMLL  3,GLASRDY           server ready?                        \n\
+         JNZ   LSSSTACK            yes, go allocate stack storage       \n\
+         LA    15,30               set the 'not ready' RC               \n\
+         J     LSSABORT            abort the call                       \n\
+LSSSTACK DS    0H                                                       \n\
          SAM31                                                          \n\
          SYSSTATE AMODE64=NO                                            \n\
          CPOOL GET,C,CPID=GLASTCP,REGS=USE   allocate a stack cell      \n\
          SAM64                                                          \n\
          SYSSTATE AMODE64=YES                                           \n\
-         DROP  2                   drop, r2 is changed by CPOOL         \n\
          LTGFR 1,1                 check if we have a cell              \n\
          BNZ   LSSINIT             go init, if we have it               \n\
          LA    15,52               if not, leave with a bad RC (52)     \n\
+LSSABORT DS    0H                                                       \n\
+         LLGF  3,GLACCNT           current caller count                 \n\
+LSSDECR1 LGR   4,3                 save caller counter for decrementing \n\
+         AGHI  4,-1                decrement caller counter             \n\
+         CS    3,4,GLACCNT         compare and swap                     \n\
+         JNZ   LSSDECR1            counter has changed already, retry   \n\
          EREGG 0,1                 restore r0 and r1 before leaving     \n\
          PR                        return to caller                     \n\
+         DROP  6                   drop, global area is not needed      \n\
 LSSINIT  DS    0H                                                       \n\
          LGR   13,1                use new storage as stack in r13      \n\
          USING PCHSTACK,13                                              \n\
@@ -1494,8 +1511,8 @@ LSSRET   DS    0H                                                       \n\
          USING PCHSTACK,13                                              \n\
          LG    5,PCHPLLPL          load latent parm from PC parm        \n\
          USING PCLPARM,5                                                \n\
-         LLGT  2,PCLPPRM1          load global area from latent parm    \n\
-         USING GLA,2                                                    \n\
+         LLGT  6,PCLPPRM1          load global area from latent parm    \n\
+         USING GLA,6                                                    \n\
          SAM31                                                          \n\
          SYSSTATE AMODE64=NO                                            \n\
          CPOOL FREE,CPID=GLASTCP,CELL=(3),REGS=USE  free stack cell     \n\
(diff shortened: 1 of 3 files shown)


c/crossmemory.c
```diff
@@ -5858,14 +5952,18 @@ void crossmemoryServerDESCTs(){
       "GLASBP   DS    X                                                        \n"
       "GLASIZE  DS    H                                                        \n"
       "GLAFLAG  DS    F                                                        \n"
-      "GLARSV1  DS    CL60                                                     \n"
+      "GLACCNT  DS    F                                                        \n"
+      "GLARSV1  DS    CL52                                                     \n"
+      "GLAUANC  DS    A                                                        \n"
```
Martin-Zeithaml, line 5957: Defined, but not directly referenced in asm.

c/crossmemory.c
```diff
@@ -5858,14 +5952,18 @@ void crossmemoryServerDESCTs(){
...
       "GLASNAM  DS    CL16                                                     \n"
       "GLALSAD  DS    A                                                        \n"
       "GLAASID  DS    H                                                        \n"
       "GLAPAD1  DS    2X                                                       \n"
       "GLASFLG  DS    F                                                        \n"
+      "GLASRDY  EQU   X'02'                                                    \n"
       "GLAECSN  DS    F                                                        \n"
-      "GLARSV2  DS    CL48                                                     \n"
+      "GLAMTSP  DS    CL32                                                     \n"
+      "GLARSV2  DS    CL16                                                     \n"
```
Martin-Zeithaml, line 5966: Also defined, but not (directly) used.

c/crossmemory.c
```diff
@@ -5877,7 +5975,8 @@ void crossmemoryServerDESCTs(){
       "GLAPCLOG DS    F                                                        \n"
       "GLARCVP  DS    AD                                                       \n"
       "GLASTCP  DS    F                                                        \n"
-      "GLARSV3  DS    CL484                                                    \n"
+      "GLADYNL  DS    A                                                        \n"
+      "GLARSV3  DS    CL480                                                    \n"
```
Martin-Zeithaml, line 5979: Defined, but unused?

c/crossmemory.c
```diff
@@ -5858,14 +5952,18 @@ void crossmemoryServerDESCTs(){
...
-      "GLARSV1  DS    CL60                                                     \n"
+      "GLACCNT  DS    F                                                        \n"
+      "GLARSV1  DS    CL52                                                     \n"
+      "GLAUANC  DS    A                                                        \n"
       "GLASNAM  DS    CL16                                                     \n"
       "GLALSAD  DS    A                                                        \n"
       "GLAASID  DS    H                                                        \n"
       "GLAPAD1  DS    2X                                                       \n"
       "GLASFLG  DS    F                                                        \n"
+      "GLASRDY  EQU   X'02'                                                    \n"
```
Martin-Zeithaml, line 5963: This will work as it is now. But it can cause a problem in the future. Why? The corresponding c definition is "fullword", because the entire flag is fullword.

Setting and testing of this is safe.
```
#define CROSS_MEMORY_SERVER_FLAG_READY        0x00000002
```

In case, we need to set this flag in asm, we have to reference the last byte, otherwise we can set the first byte:
```
* Let's say GLASFLG is X'00000000'
* OI instruction is usually used for flag setting
  OI  GLASFLG,GLASRDY  --> X'02000000'
* But we need to get X'00000002'
```

c/crossmemory.c
ifakhrutdinov: Yes, the intention is to match the layout of C struct.

c/crossmemory.c
ifakhrutdinov: Yes, the intention is to match the layout of C struct.

c/crossmemory.c
ifakhrutdinov: This is really just for checking that exact value in HLASM; not setting it from HLASM. If I were to replicate the C code, the HLASM would get complicated for no reason but consistency; I'm not sure if it's important here. What could you propose to improve this check in the HLASM code?

Also, since `TMLL` uses 16-bit immediate values, it would have probably been better if I had used `x'0002'`, but `x'02'` seems to get the job done too.

Martin-Zeithaml: As I wrote, the code as it is written now is fine, because there is just checking of the flag. 
Generally it is not a good idea to define fullword flag area and 1 byte mask.

Let's keep it as it is.

Martin-Zeithaml (dismissed):

ifakhrutdinov:
All, I've rebased the branch and made some minor changes documented at the end of "Additional changes:" in the description. The diff can be seen using: `git diff feature/active-caller-count-old feature/active-caller-count`. 
```diff
diff --git a/c/crossmemory.c b/c/crossmemory.c
index 14cb758..0e9afc3 100644
--- a/c/crossmemory.c
+++ b/c/crossmemory.c
@@ -1599,7 +1599,7 @@ LSSTERM  DS    0H                                                       \n\
 "        PUSH  USING                                                    \n\
 LCPPRLG  LARL  10,LCPPRLG          establish addressability using r10   \n\
          USING LCPPRLG,10                                               \n\
-         LGR   5,4                 latent parm to r4 for consistency    \n\
+         LGR   5,4                 latent parm to r5 for consistency    \n\
          USING PCLPARM,5                                                \n\
          LLGT  6,PCLPPRM1          load global area from latent parm    \n\
          USING GLA,6                                                    \n\
diff --git a/h/crossmemory.h b/h/crossmemory.h
index 21dfc94..8815af8 100644
--- a/h/crossmemory.h
+++ b/h/crossmemory.h
@@ -235,7 +235,7 @@ typedef struct CrossMemoryServerGlobalArea_tag {
   unsigned short size;
   unsigned int flags;
 #define CMS_GLOBAL_AREA_FLAG_PRIVATE_MODULE 0x00000001
-  int callerCount;
+  volatile int callerCount;
   char reserved1[52];
 
   void * __ptr32 userServerAnchor;
@@ -244,7 +244,7 @@ typedef struct CrossMemoryServerGlobalArea_tag {
   struct CrossMemoryServer_tag * __ptr32 localServerAddress;
   unsigned short serverASID;
   char padding1[2];
-  unsigned int serverFlags;
+  volatile unsigned int serverFlags;
   int ecsaBlockCount; /* must be on a 4-byte boundary for atomicIncrement */
 #define CMS_MAX_ECSA_BLOCK_NUMBER 32
 #define CMS_MAX_ECSA_BLOCK_SIZE   65536
```

Martin-Zeithaml (approved):

JoeNemo:
@ifakhrutdinov, I assume you were getting the compiler to not over-optimize variable access in recent change.  It looks good by my.  Are you going to merge?

PR #575, opened 2026-03-10
(v3) Rename F_LOCK struct to avoid standard library conflict

1000TurquoisePogs:
The F_LOCK struct we have collides with <unistd.h>, but I guess not many people were using it in the past so we did not notice.

After discussion with Joe, we agreed to rename it to FLock.


JoeNemo (approved):
Approved

PR #577, opened 2026-03-17
Improved peer hint when schema invalid

1000TurquoisePogs:
Alternative slighty more complex PR to https://github.com/zowe/zowe-common-c/pull/576

This additionally
1) has better output to give you an example of what you may do to correct the situation
2) changes all output slashes to dots and gets rid of the leading one. configmgr validation errors will now match how we reference YAML properties to users.

Output preview:

image


JoeNemo (dismissed):
approved.

JoeNemo (approved):
approved.

PR #578, opened 2026-03-17
Add semantic validation for user, group, file and directory

1000TurquoisePogs:
For discussion.

This PR allows one to add "zoweValidate" to a json schema with values like "directoryExists".
This makes configmgr perform existence checks on the YAML value backed by the schema.

This is implemented by having function pointers in jsonschema.c which may be implemented, in this case, by configmgr.c
If jsonschema.c encounters a value for zoweValidate which nobody has implemented, it ignores it.
That is, it checks the schema as it would have without this PR.
But for any validator it does have, it calls the function and uses the boolean to determine if it should give an error or not.

image

### DIFF
--- a/build/build_cmgr_xlclang.sh
+++ b/build/build_cmgr_xlclang.sh
@@ -163,6 +163,7 @@ xlclang \
   ${COMMON}/c/yaml2json.c \
   ${COMMON}/c/zos.c \
   ${COMMON}/c/zosfile.c \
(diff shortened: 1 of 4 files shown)


1000TurquoisePogs:
When I spoke with Joe, he said it would be good to have an option in configmgr to bypass these checks, because we never know if someone's going to use configmgr in ways we didnt anticipate in which we want to not do such checks.

PR #580, opened 2026-03-24
Read symlink targets and return them in file api

1000TurquoisePogs:
This changes the file API such that if we detect an object is a symlink, we dig deeper to return the path of the target in a new field "symlinkTarget".
Additionally, the "directory" boolean is now the value of the symlink target rather than the symlink. This is opinionated but my opinion is it allows existing code to 'do the right thing' without code changes on their end - if a symlink target is a directory, programs should act like the symlink itself is a directory.

the api, and results of fixing this directory behavior, are seen in this screenshot
symlink-target

The old behavior where the editor did not know a symlink was a directory was not so good...
symlink-broken


JoeNemo (approved):
approved.

PR #593, opened 2026-04-23
Update file api encoding behavior based on user preference

1000TurquoisePogs:
## Proposed changes

This PR addresses that the file content endpoint should accept encoding name strings and integer CCSIDs as query parameters and the related encoding-name coverage improvements from the IANA character sets registry.

This PR includes https://github.com/zowe/zowe-common-c/pull/592

---

### Summary

Three related improvements to the charset and HTTP file-serving layer:

**1. `parseEncodingValue()` — new function in `charsets.c` / `charsets.h`**

A single resolver that accepts either a charset name string or a decimal CCSID integer string and returns the CCSID integer:
- Named charsets (e.g. `"IBM-1047"`, `"UTF-8"`) are resolved via `getCharsetCode()`.
- Integer strings (e.g. `"1047"`, `"819"`) are parsed with `sscanf` in the range 1–65535.
- The sentinel string `"binary"` (case-insensitive) resolves to `CCSID_BINARY` (65535).
- Returns -1 for unrecognised or out-of-range values.

**2. Extended IBM codepage coverage in `getCharsetCode()`**

All IBM codepage names from the IANA character sets registry are now recognised, using the convention `IBM-NNNN` (dash-separated, no leading zeros). This covers:
- EBCDIC single-byte: IBM-37, IBM-273 through IBM-297, IBM-420 through IBM-424, IBM-500, IBM-870 through IBM-905, IBM-918, IBM-924, IBM-1026, IBM-1047, IBM-1140 through IBM-1149
- PC code pages: IBM-437, IBM-775, IBM-850 through IBM-869
- Aliases: `IBM1047` (legacy unformatted form), `IBM-819` (alias for ISO-8859-1)
- The function header now includes a reference to the IANA registry URL.

**3. Independent `sourceEncoding`/`targetEncoding` parameters in `respondWithUnixFile2()`**

The GET endpoint for USS file content (`/unixfile/contents/**`) previously required both `sourceEncoding` and `targetEncoding` to be supplied together, and rejected requests with only one. The logic has been changed so each parameter is fully independent:

| `sourceEncoding` | `targetEncoding` | Behaviour |
|---|---|---|
| absent / `"auto"` | absent / `"auto"` | file tag → auto web encoding (original default) |
| explicit (e.g. `"IBM-1047"`) | absent / `"auto"` | caller source → auto web encoding |
| absent / `"auto"` | explicit (e.g. `"UTF-8"`) | file tag → caller target |
| explicit | explicit | caller source → caller target |
| `"binary"` | *(any)* | raw bytes, no conversion |
| *(any)* | `"binary"` | raw bytes, no conversion |

`"auto"` is now a valid explicit value identical to omitting the parameter. The old broken `force=enable` block (which ran after `writeHeader()` and therefore could not return 400 errors) has been removed. Validation of encoding params is performed before `writeHeader()`.

## Type of change

## PR Checklist

## Testing

1. **Named charset strings accepted**: `GET /unixfile/contents/path/file.c?sourceEncoding=IBM-1047&targetEncoding=UTF-8` should return UTF-8 text converted from EBCDIC 1047.
2. **Integer CCSID strings accepted**: `GET /unixfile/contents/path/file.c?sourceEncoding=1047&targetEncoding=1208` should return the same result as above.
3. **Single param (source only)**: `GET /unixfile/contents/path/file.c?sourceEncoding=IBM-1047` should convert from 1047 to the auto-selected web encoding (ISO-8859-1 for single-byte CCSIDs on z/OS).
4. **Single param (target only)**: `GET /unixfile/contents/path/file.c?targetEncoding=UTF-8` should convert from the file's tagged CCSID to UTF-8.
5. **`auto` explicit**: `GET /unixfile/contents/path/file.c?sourceEncoding=auto&targetEncoding=UTF-8` should behave identically to test 4.
6. **Binary sentinel**: `GET /unixfile/contents/path/file.c?sourceEncoding=binary` should stream raw bytes without conversion.
7. **Invalid value**: `GET /unixfile/contents/path/file.c?sourceEncoding=bogus` should return HTTP 400.
8. **No params**: existing default behaviour preserved — file tag used as source, auto web encoding as target.

## Further comments

The `"auto"` sentinel was added to support callers that want to be explicit about which parameter is defaulting without inventing a side-channel convention. zlux plugins and URI broker can use `sourceEncoding=auto&targetEncoding=binary` to request raw byte streaming while making the intent clear in the URL.

The removal of the paired-parameter requirement is backward-compatible: any caller that already supplied both params continues to work identically.


ifakhrutdinov:
@1000TurquoisePogs , have you used AI for this? If so, have you verified every single line?

JoeNemo:
This work got merged into a project that was merged with PR 630.

PR #595, opened 2026-05-12
Use "%.*s" version of snprintf to stop overreading in 'zosResolveSymbol()'.

Gautham-coder:
## Description
Resolving system symbols causes abend sometimes.

Root cause
----------
The original resolveSymbol() code copied resolved subtext with snprintf(..., "%s", ...).
In offset mode, symbol values are length-delimited in a packed symbol area and are not guaranteed to be null-terminated.
Using "%s" could therefore read past the intended value into adjacent packed entries and eventually into invalid storage, causing an 0C4 abend.

A second issue existed in the QuickJS wrapper: if resolveSymbol() returned NULL, qjszos.c would still call strlen(result), which could also crash.

## Proposed changes
1. resolveSymbol() now copies using "%.*s" which limits the number of bytes read to subtextLength.
2. If a symbol is defined but has no value:
   - offset mode: subtextLength == 0 -> return an allocated empty string
   - pointer mode: subtextPtr == NULL or subtextLength == 0 -> return an allocated empty string
3. Allocation failures now set RESOLVESYMBOL_RETURN_ALLOC_FAILED.
4. zosResolveSymbol() in c/qjszos.c now checks for NULL returns:
   - if rc != 0, it throws a JS exception using JS_ThrowReferenceError
   - otherwise it returns JS_NULL
   - this avoids assuming JS_ThrowInternalError is available in the embedded QuickJS version
5. resolveSymbolBySyscall() was updated for consistency **BUT still needs changes as the current code cannot differentiate between a blank and symbol not found and returns NULL in these cases.**

### Description of %.*s in snprintf() 
Syntax
`snprintf(dest, destSize, "%.*s", precision, src);`
How it works
%.*s is a width-bounded string format specifier. The `.*` means the precision is supplied as a runtime argument (the int before src).
Argument: Role
dest: Destination buffer to write into
destSize: Maximum bytes to write including null terminator
precision: Maximum bytes to read from src
src: Source pointer

This PR addresses Issue: [*Link to Github issue within https://github.com/zowe/zowe-common-c/issues* if any]

This PR depends upon the following PRs:

## Type of change
- [x] Refactor the code 

## PR Checklist
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

## Testing
Using a lot of system symbol in your yaml and recycle your Zowe.
Before:
You will get a OC4 ABEND

After:
You will not see the abend.

### DIFF
--- a/c/zos.c
+++ b/c/zos.c
@@ -405,11 +405,11 @@ char *resolveSymbolBySyscall(const char *inputSymbol, int *rc, int *rsn) {
       result = safeMalloc(outputLen+1, "output");
       if (result == NULL) {
         *rc = RESOLVESYMBOL_RETURN_ALLOC_FAILED;
-    } else {
-      snprintf(result, outputLen+1, "%.*s", outputLen, below2G->output);
-    }
-    FREE_STRUCT31(STRUCT31_NAME(below2G));
-    return result;
+      } else {
+        snprintf(result, outputLen+1, "%.*s", outputLen, below2G->output);
+      }
+      FREE_STRUCT31(STRUCT31_NAME(below2G));
+      return result;
     }
 
 }
--- a/c/zos.c
+++ b/c/zos.c
@@ -396,23 +396,21 @@ char *resolveSymbolBySyscall(const char *inputSymbol, int *rc, int *rsn) {
     }
 
 //    printf("input '%s', output '%s'\n", below2G->input, below2G->output);
+    // We are not checking for return code here and therefore cannot differentiate between a blank and not found.
+    // For more info: https://www.ibm.com/docs/en/zos/2.2.0?topic=hsp-asasymbm-asasymbf-substitute-text-symbols
+    // printf("ASASYMBF return code %x\n", below2G->returnCode);
     int outputLen = strlen(below2G->output);
-    if (outputLen != 0){
-      char *result = safeMalloc(outputLen+1, "output");
+    char *result = NULL;
+    if (outputLen > 0) {
+      result = safeMalloc(outputLen+1, "output");
       if (result == NULL) {
         *rc = RESOLVESYMBOL_RETURN_ALLOC_FAILED;
-        FREE_STRUCT31(STRUCT31_NAME(below2G));
-        return NULL;
-      }
-      memcpy(result, below2G->output, outputLen);
-      result[outputLen] = '\0';
-      FREE_STRUCT31(STRUCT31_NAME(below2G));
-      return result;
-    } else{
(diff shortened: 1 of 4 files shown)


c/zos.c
```diff
@@ -396,18 +396,22 @@ char *resolveSymbolBySyscall(const char *inputSymbol, int *rc, int *rsn) {
...
 
 //    printf("input '%s', output '%s'\n", below2G->input, below2G->output);
+    // We are not checking for return code here and therefore cannot differentiate between a blank and not found.
+    // For more info: https://www.ibm.com/docs/en/zos/2.2.0?topic=hsp-asasymbm-asasymbf-substitute-text-symbols
+    // printf("ASASYMBF return code %x\n", below2G->returnCode);
     int outputLen = strlen(below2G->output);
-    if (outputLen != 0){
-      char *result = safeMalloc(sizeof(below2G->output), "output");
-      snprintf(result, outputLen+1, "%s", below2G->output);
+    char *result = NULL;
```
ifakhrutdinov, line 403: According to the description:
> The original resolveSymbol() code copied resolved subtext with snprintf(..., "%s", ...).
In offset mode, symbol values are length-delimited in a packed symbol area and are not guaranteed to be null-terminated.
> Using "%s" could therefore read past the intended value into adjacent packed entries and eventually into invalid storage, causing an 0C4 abend.

Wouldn't then `int outputLen = strlen(below2G->output);`, which we still use here, also ABEND?

c/zos.c
```diff
@@ -396,18 +396,22 @@ char *resolveSymbolBySyscall(const char *inputSymbol, int *rc, int *rsn) {
...
+    if (outputLen > 0) {
+      result = safeMalloc(outputLen+1, "output");
+      if (result == NULL) {
+        *rc = RESOLVESYMBOL_RETURN_ALLOC_FAILED;
+      } else {
+        snprintf(result, outputLen+1, "%.*s", outputLen, below2G->output);
+      }
```
ifakhrutdinov, line 410: Looks like we now never call `FREE_STRUCT31` when `outputLen <= 0`?

c/zos.c
ifakhrutdinov: Or was it only about the other `snprintf` call?

Gautham-coder: Yes Irek, that explanation was for the snprintf() in resolveSymbol().

c/zos.c
Gautham-coder: That's correct Irek, fixing it now. Thank you.

JoeNemo (approved):
Approved.

PR #600, opened 2026-05-27
JWT `extractParts` minor update

Martin-Zeithaml:
The variable `i` is defined as unsigned int, this condition
```c
for (i = nParts - 1; i > 0; i--) {
```
could be wrong: if `nParts` is `0`, then `i = UINT_MAX` and the loop is over the unsigned int range.


JoeNemo (dismissed):
Approved.

jwt/jwt/jwt.c
```diff
@@ -92,7 +92,8 @@ int setJwtTrace(int toWhat) {
 static int extractParts(char base64Buf[], int maxParts,
                         char *dparts[],  int pLen[], char *decodedText) {
   char *tokenizer;
-  unsigned int nParts, i = 0;
```
ifakhrutdinov, line 95: Looking at the code, does `nParts` even need to be unsigned? If not, a better fix is to drop `unsigned`.

ifakhrutdinov (approved):
lgtm, thank you for the fix.

JoeNemo (approved):
Approved

JoeNemo (approved):
Approved.

PR #603, opened 2026-05-27
Avoid ABEND when receiving request content and chunks

ChongZhou-Broadcom:
## Proposed changes

When `processHttpFragment` is receiving the request content, either the first block or the subsequent chunks, it calls `SLHAlloc` to allocate buffer. However, `SLHAlloc` will trigger ABEND when the preallocated short-lived buffer runs out of space. For example:

https://github.com/zowe/zowe-common-c/blob/9803cfcf0b5644b2da7dab841b0bbfde6dffa4d6/c/httpserver.c#L2390-L2395

https://github.com/zowe/zowe-common-c/blob/9803cfcf0b5644b2da7dab841b0bbfde6dffa4d6/c/utils.c#L1495-L1513

It makes DDoS possible. This PR changes the behavior to return http/500 instead of ABEND when it fails to allocate buffer for request content.

However, I did not implement a configurable limit for the max content size, I only avoided ABEND. I was hesitating because it sounds like a good idea but the HTTP spec does not force such limit.

This PR addresses Issue: N/A

This PR depends upon the following PRs: N/A

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] New and existing unit tests pass locally with my changes (N/A)
- [ ] video or image is included if visual changes are made (N/A)
- [ ] Relevant update to CHANGELOG.md (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below (N/A)

## Testing

N/A

## Further comments

N/A


JoeNemo:
I prefer that this becomes SLHAlloc2, by our current standard, adding an argument and having the original API function call a new one that has the new argument determine if the call should abort vs. return NULL.  Default is abort.  Also, is this code being called in a recovery.c recovery context.  That would handle the abort already with no DDOS.   *AND* when we have bugs that assert potential memory problems or DDOS, please make new files or subdirs under zowe-common-c/tests to repro and allow future regression testing.

ChongZhou-Broadcom:
@JoeNemo Thank you for your review! My response is as below.

> I prefer that this becomes SLHAlloc2, by our current standard, adding an argument and having the original API function call a new one that has the new argument determine if the call should abort vs. return NULL. Default is abort.

Totally agreed. Will change it soon.

> Also, is this code being called in a recovery.c recovery context. That would handle the abort already with no DDOS.

Please correct me if I'm wrong. The only `recoveryPush ... recoveryPop` section in httpserver.c is found in `handleHttpService`. According to `serviceLoop`, `processHttpFragment` is not called inside `handleHttpService` but before it, which means ABENDs happen in `processHttpFragment` is not recoverable.

> AND when we have bugs that assert potential memory problems or DDOS, please make new files or subdirs under zowe-common-c/tests to repro and allow future regression testing.

Will do.

c/utils.c
```diff
@@ -1562,6 +1564,34 @@ char *SLHAlloc(ShortLivedHeap *slh, int size){
   return (char *)data;
   }
 
+char *SLHAlloc_NoAbend(ShortLivedHeap *slh, int size) {
```
ifakhrutdinov: I think we should just factor out the logic out of  `SLHAlloc` into something like `SLHAllocInternal` and then modify the behaviour in the wrapper functions (`SLHAlloc` and e.g., `SLHAlloc2`).

This kind of leaks abstraction.

ChongZhou-Broadcom: Yes, I agree. @JoeNemo suggested something similar. I have updated the implementation.

ChongZhou-Broadcom:
@JoeNemo I implemented the "SLHAlloc2" according to your suggestion. I created "SLHAlloc_NoAbend" because I was thinking about c++'s "new(std::nothrow)" :)

~However, I didn't implement the test. To be honest, I created a test but I don't know how to build and run it. The code change is in `processHttpFragment` in httpserver.c, so if I want to unit test that function, I nearly have to rebuild a whole zss.~

ChongZhou-Broadcom:
@JoeNemo Hi Joe, as you requested earlier, I added two test cases in tests/. Please review it when you have a moment. Thanks.

c/utils.c
```diff
@@ -1495,10 +1495,10 @@ ShortLivedHeap *makeShortLivedHeap64(int blockSize, int maxBlocks){
 }
 
 char *SLHAlloc(ShortLivedHeap *slh, int size){
-  if (size <= 0) {
-    return NULL; 
-  }
```
JoeNemo, line 1543: Why is this null check gone when the code moved into SLHAlloc2

ChongZhou-Broadcom: This doesn't appear to have been an intentional removal. I will add it back right away.

JoeNemo (approved):
Approved.

PR #604, opened 2026-05-29
`pdsutil`: replace sprintf with snprintf for safety

Martin-Zeithaml:
`filenameBuffer` is limited, while `pdsName` has no length limit.

### DIFF
--- a/c/pdsutil.c
+++ b/c/pdsutil.c
@@ -46,7 +46,7 @@ int startPDSIterator(PDSIterator *iterator, char *pdsName){
   char filenameBuffer[50];
   memset(iterator,0,sizeof(PDSIterator));
 
-  snprintf(filenameBuffer,sizeof(filenameBuffer), "//'%s'",pdsName);
+  snprintf(filenameBuffer, sizeof(filenameBuffer), "//'%s'", pdsName);
   iterator->in = fopen(filenameBuffer,"rb");
   if (iterator->in){
     iterator->posInBlock = 0;
@@ -149,7 +149,7 @@ void listDirectory(char *pdsName){
   int blockCount = 0;
   int lastBlock = 0;
 
-  snprintf(filenameBuffer, sizeof(filenameBuffer), "//'%s'",pdsName);
+  snprintf(filenameBuffer, sizeof(filenameBuffer), "//'%s'", pdsName);
   in = fopen(filenameBuffer,"rb");
    zowelog(NULL, LOG_COMP_RESTDATASET, ZOWE_LOG_DEBUG, "fopen in=0x%x errno=%d\n",in,errno);
   while (!feof(in) && !lastBlock){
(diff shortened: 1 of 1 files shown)


c/pdsutil.c
```diff
@@ -46,7 +46,7 @@ int startPDSIterator(PDSIterator *iterator, char *pdsName){
   char filenameBuffer[50];
   memset(iterator,0,sizeof(PDSIterator));
 
-  sprintf(filenameBuffer,"//'%s'",pdsName);
+  snprintf(filenameBuffer, sizeof(filenameBuffer), "//'%s'", pdsName);
```
ifakhrutdinov, line 49: It's definitely better than the original code, but we should really verify the length because this could be a silent issue if the provided long data set name get truncated into a valid _unintended_ 44 character data set name.

Martin-Zeithaml: Good catch! I did not realize this. This will be probably more complicated than expected.

Martin-Zeithaml: Valid data set name is prefixed with double slash and enclosed in single apostrophes:

```
INPUT:     SYS1.PARMLIB
OUTPUT: //'SYS1.PARMLIB'

INPUT:     MAX45678.A2345678.A2345678.A2345678.A2345678
OUTPUT: //'MAX45678.A2345678.A2345678.A2345678.A2345678'
```

Invalid data set name of 45 chars is prefixed with double slash and enclosed in single apostrophes, but the attempt to open will fail:

```
INPUT:     WRONG678.CHARS#45.A2345678.A2345678.A23456789
OUTPUT: //'WRONG678.CHARS#45.A2345678.A2345678.A23456789'

INPUT:     WRONG678.CHARS#45.A2345678.A2345678.A2345678.
OUTPUT: //'WRONG678.CHARS#45.A2345678.A2345678.A2345678.'

INPUT:     WRONG678.CHARS#45.A2345678.A2345678.A234.A234
OUTPUT: //'WRONG678.CHARS#45.A2345678.A2345678.A234.A234'
```

Invalid data set name of 46+ is missing closing single apostrophe, so the attempt to open will fail:

```
INPUT:     WRONG678.CHARS#46.A2345678.A2345678.A2345678.A
OUTPUT: //'WRONG678.CHARS#46.A2345678.A2345678.A2345678.A

INPUT:     WRONG678.CHARS#47.A2345678.A2345678.A2345678.A2
OUTPUT: //'WRONG678.CHARS#47.A2345678.A2345678.A2345678.A
```

IMO no unintended trim can happen. 

I did not check yet, but if we were trying to open invalid data set name, we will do the same with this fix - the only change is, there is no buffer overflow.

JoeNemo:
Is this good to go??

Martin-Zeithaml:
> Is this good to go??

I am done here. I think the silent truncation can not happen, this is ready.

c/pdsutil.c
JoeNemo: Are we good to go here?

JoeNemo (approved):
Approved.

PR #605, opened 2026-05-29
`httpServer`: check if nativeVersion exists and not null

Martin-Zeithaml:
Any input parameter for `headerMatch` could be possibly null, add check for it.

Then used here:

https://github.com/zowe/zowe-common-c/blob/fcdc13c41e1e503842b63a2bf64e193dfde7bded/c/httpserver.c#L5518-L5521

### DIFF
--- a/h/http.h
+++ b/h/http.h
@@ -195,8 +195,6 @@ char *copyStringToNative(ShortLivedHeap *slh, char *s, int len);
 char* destructivelyNativize(char *s);
 void asciify(char *s, int len);
 
-int headerMatch(HttpHeader *header, char *s);
-
 char *toASCIIUTF8(char *buffer, int len);
 
 int writeFully(Socket *socket, char *buffer, int len);
--- a/h/http.h
+++ b/h/http.h
@@ -195,6 +195,8 @@ char *copyStringToNative(ShortLivedHeap *slh, char *s, int len);
 char* destructivelyNativize(char *s);
 void asciify(char *s, int len);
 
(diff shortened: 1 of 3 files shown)


h/http.h
```diff
@@ -195,8 +195,6 @@ char *copyStringToNative(ShortLivedHeap *slh, char *s, int len);
 char* destructivelyNativize(char *s);
 void asciify(char *s, int len);
 
-int headerMatch(HttpHeader *header, char *s);
```
ifakhrutdinov, line 198: You can't just remove a public function. 

You could (yes, integers for boolean logic aren't ideal, but this is the function's signature...):
```c
int headerMatch(HttpHeader *header, char *s) {
  if (header == NULL || header->nativeValue == NULL) {
    return 0;
  }
  return !strcmp(header->nativeValue,s);
}
```
If you want to make the code around less ugly.

Martin-Zeithaml: Your suggestion was my first implementation, but the ugly part remains the same: we are comparing structure with string - you have to see/know the function definition.

I don't like this solution, but the main task here was to check possible null values, which was achieved.

ifakhrutdinov: I agree. Also, we have inconsistent conventions with regards to bad arguments: in some functions it's garbage in garbage out and in some the args are validated... well, it is what it is, unfortunately.

ifakhrutdinov (approved):

JoeNemo (approved):
Approved.

PR #606, opened 2026-06-01
If possible do not use STCK

Martin-Zeithaml:
Use rather ICSF-based random number than STCK.


JoeNemo (dismissed):
Approved.

c/httpserver.c
```diff
@@ -1654,7 +1654,11 @@ static HttpServer *makeSecureHttpServerInner(STCBase *base, int port,
   server->config = (HttpServerConfig*)safeMalloc31(sizeof(HttpServerConfig),"HttpServerConfig");
   server->properties = htCreate(4001,stringHash,stringCompare,NULL,NULL);
   memset(server->config,0,sizeof(HttpServerConfig));
-  int64 now = getFineGrainedTime();
+  int64 now;
+  int icsfReason = 0;
+  if (icsfGenerateRandomNumber(&now, sizeof(now), &icsfReason) != 0) {
```
1000TurquoisePogs: this looks like a good improvement but shouldn't this also be a different variable name than "now"?

JoeNemo (approved):
Approved.

PR #607, opened 2026-06-02
Correct string length in calling compareIgnoringCase

ChongZhou-Broadcom:
## Proposed changes

In httpserver.c and httpclient.c, the code recognizes header keys in the following way:

```c
  if (!compareIgnoringCase(newHeader->nativeName,"Transfer-Encoding",parser->headerNameLength)){
    if (!compareIgnoringCase(newHeader->nativeValue,"chunked",parser->headerValueLength)){
      parser->isChunked = TRUE;
    }
}
```

It will recognize any prefixes of a header key. **It is usually not a very big issue.** But I think just a small change could make the match strict, so why not?

```c
  if (!compareIgnoringCase(newHeader->nativeName,"Transfer-Encoding",parser->headerNameLength + 1)){
    if (!compareIgnoringCase(newHeader->nativeValue,"chunked",parser->headerValueLength + 1)){
      parser->isChunked = TRUE;
    }
}
```

We can easily strengthen the string comparison by just taking the terminating '\0' into the comparison ... unless we intentionally want to match any prefixes.

I did not add unit tests because PR #603 introduced a working "little framework" for testing httpserver/httpclient. Without which, it's pretty hard to unit test these two units.

This PR addresses Issue: N/A

This PR depends upon the following PRs: N/A

## Type of change
- [ ] Refactor the code 

## PR Checklist
- [ ] I have made corresponding changes to the documentation (N/A)
- [ ] New and existing unit tests pass locally with my changes (N/A)
- [ ] video or image is included if visual changes are made (N/A)
- [ ] Relevant update to CHANGELOG.md (N/A)
- [ ] I have added tests that prove my fix is effective or that my feature works, or describe a test method below (N/A)

## Testing

N/A

## Further comments

N/A


ifakhrutdinov:
This looks like a good idea; I need to read the code carefully as I'm not sure whether those strings are always null-terminated (I know I know, they should be).

JoeNemo:
Please check the lengths explicity.  Hiding the length check in the compareXXX() is clever, but not clear.

c/httpclient.c
```diff
@@ -166,14 +166,14 @@ static void addResponseHeader(HttpResponseParser *parser) {
 #endif
 
   /* pull out enough data for parsing the entity body */
-  if (!compareIgnoringCase(newHeader->nativeName, "Transfer-Encoding", parser->headerNameLength)) {
-    if (!compareIgnoringCase(newHeader->nativeValue, "chunked", parser->headerValueLength)) {
+  if (!compareIgnoringCase(newHeader->nativeName, "Transfer-Encoding", parser->headerNameLength + 1)) {
```
Martin-Zeithaml: `nativeName` and `nativeValue` are created via `copyStringToNative`, which ensures null terminated string.
What about to replace `compareIgnoringCase` with `strcasecmp`?

In other words what is the reason for implementation of `compareIgnoringCase`?

ChongZhou-Broadcom: My guess: compareIgnoringCase only assumes strings to be in EBCDIC, while strcasecmp's behavior depends on the compiler option. To avoid potential impacts of compiler options, @1000TurquoisePogs created an EBCDIC only function and explicitly converted strings (e.g., header field keys) to EBCDIC before the comparison.

ChongZhou-Broadcom:
@JoeNemo I reimplemented it. Please review it again when you have a moment. Thank you so much.

JoeNemo (approved):
Approved.

PR #608, opened 2026-06-05
Strengthen ICSF encryption

ChongZhou-Broadcom:
## What's the problem?

Our session token encryption logic might have a cryptographic flaw.

Currently, the code uses AES-CFB mode via z/OS ICSF (CSNBSYE / CSNBSYD), but it relies on a static Initialization Vector (ICV) and lacks integrity verification. Because the encryption stream is predictable and unauthenticated, it makes our session tokens malleable. An attacker could potentially perform bit-flipping attacks to forge tokens or impersonate other users without knowing the key.

## How is it fixed?

I have refactored the encryption and decryption logic:

Switched to AES-GCM: Changed the ICSF processing rule from CFB to GCM. I also updated the rule keyword from INITIAL to ONLY to comply with IBM ICSF requirements for GCM mode.

Dynamic ICV: I use `icsfGenerateRandomNumber` to generate a unique 12-byte IV for every single encryption call.

Integrity Protection: We now leverage the 16-byte GCM authentication tag (keyParms in ICSF) to ensure data integrity. The decryption service will automatically validate this tag and fail if the ciphertext has been modified.

Token Structure Adjustment: Introduced a structured wire format (ICSFEncodingBlock) that packs the 12-byte random IV, the 16-byte GCM tag, and the actual ciphertext payload together.

## Test

Not done yet. This is a draft PR for now. The purpose is to better illustrate the idea by the code-diff view.

## References

https://www.ibm.com/docs/en/zos/2.5.0?topic=data-symmetric-key-encipher-csnbsye-csnbsye1-csnesye-csnesye1

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -1410,9 +1410,24 @@ static int encodeSessionToken(ShortLivedHeap *slh,
                               const HttpServerConfig *config,
                               const char *tokenText,
                               unsigned int tokenTextLength,
-                              char **result) {
+                              char **result,
+                              unsigned int *resultLength) {
 
-  unsigned int encodedTokenTextLength = tokenTextLength;
+  unsigned int encodedTokenTextLength;
+#ifdef __ZOWE_OS_ZOS
+
+  if (icsfEncipherOutputLength(tokenTextLength, &encodedTokenTextLength) != 0) {
+    zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG,
+      "Error: incorrect decoded session token length (size=%u)\n", tokenTextLength);
+    return -1;
+  }
+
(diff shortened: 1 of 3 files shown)


ChongZhou-Broadcom:
**I will consider closing this draft PR soon because the mentioned cryptographic flaw was proved to be non-exploitable in ZSS.**

According to the flaw, I simulated a "Bit-flipping attack" to ZSS in my test environment. Although I managed to inject the impersonated user ID, but the AES algorithm's "Avalanche Effect" screwed up the second AES window in the token, due to which, ZSS rejected the fake cookie because it failed to locate the second colon (which is in the second AES window).

Therefore, although `decodeSessionToken` does not check the integrity of the encrypted text, `sessionTokenStillValid` tries to extract all fields from the token, which also requires integrity.

PR #611, opened 2026-06-12
Change `RC_JWT_INSECURE` value from 14 to 16

Martin-Zeithaml:
Either last item has a typo (14 vs 16) or there is some hack, which is not explained:
```c
#define RC_JWT_UNKNOWN_CONTEXT_TYPE 14
#define RC_JWT_NOT_CONFIGURED    15
#define RC_JWT_INSECURE          14
#define MAX_JWT_RC               RC_JWT_INSECURE   // Sets max to 14 and excluding RC_JWT_NOT_CONFIGURED
```


JoeNemo (approved):
Approved.

PR #612, opened 2026-06-12
Add null check for password before length check

Martin-Zeithaml:
Impersonation calls directly with `NULL` as password couple times, for example:

https://github.com/zowe/zowe-common-c/blob/773d70bc3bac9c200ccbf89558cb0645d7a18a63/c/impersonation.c#L61-L66


JoeNemo (approved):

JoeNemo (approved):
Approved.

PR #613, opened 2026-06-15
Json conversion length

Martin-Zeithaml:
## Check the `len`
```c
  if (len > 0 && (long long)len * 2 + 7 <= INT_MAX) {
    conversionOutputLength = 2 * len;
    buffer = SLHAlloc(slh, conversionOutputLength);
  }
```

`SLHAlloc` rounds up the memory length to 8-byte boundary. This check ensures we will stay in `int` range.

## Local variable renamed
`convesionOutputLength` -> `conversionOutputLength`

## Update

The size logic was moved to `SLHAlloc` in #620 
Once reviewed and merged, this PR can be updated.

## Final change

The #620 was merged, so the changes:
* buffer initially NULL (just for sure)
* variable renamed

### DIFF
--- a/c/json.c
+++ b/c/json.c
@@ -2680,6 +2680,9 @@ Json *jsonParseUnterminatedUtf8String(ShortLivedHeap *slh, int outputCCSID,
   if (len > 0 && len <= (((INT_MAX) / 2) - 7)) {
     conversionOutputLength = 2 * len;
     buffer = SLHAlloc(slh, conversionOutputLength);
+  } else {
+    snprintf(errorBufferOrNull, errorBufferSize, "incorrect UTF8 string length %i", len);
+    return NULL;
   }
   if (buffer == NULL) {
     snprintf(errorBufferOrNull, errorBufferSize, "not enough memory");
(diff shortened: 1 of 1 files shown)


c/json.c
```diff
@@ -2673,23 +2673,30 @@ Json *jsonParseUnterminatedUtf8String(ShortLivedHeap *slh, int outputCCSID,
                                       char *jsonUtf8String, int len,
                                       char* errorBufferOrNull, int errorBufferSize) {
   int convRc, convRsn;
-  int convesionOutputLength = 2 * len;
-  char *buffer;
+  int conversionOutputLength = 0;
+  char *buffer = NULL;
 
-  buffer = SLHAlloc(slh, convesionOutputLength);
+  // SLHAlloc rounds up to a multiple of 8
```
ifakhrutdinov: Thinking about how much this leaks abstractions...

ifakhrutdinov: In my opinion, the ` + 7` part should be done by `SLHAlloc`; i.e. `SLHAlloc` should check for any integer overflow it may cause.

ifakhrutdinov: And to be precise, `SLHAlloc` does not always add 7; it adds from 1 to 7 depending on the provided length. In other words, `(long long)len * 2 + 7` is likely larger than what `SLHAlloc` ends up with.

Martin-Zeithaml: I agree, this fix is not perfect. The check should be included in `SLHAlloc`.

Martin-Zeithaml: See #620

ifakhrutdinov: Martin, I think #620 is a good enhancement.

JoeNemo (approved):
Approved.

PR #614, opened 2026-06-15
Add null check for forceEnabled in httpserver.c

Martin-Zeithaml:
`getQueryParam` can possibly return `NULL`.
Check `NULL` value before the string compare.


JoeNemo (approved):
Approved.

JoeNemo (approved):
Approved.

PR #615, opened 2026-06-16
Implement error check for `out_rc` in `rs_icsfp11`

Martin-Zeithaml:
Missing rc check and update for null check.


JoeNemo (approved):
Approved.

PR #616, opened 2026-06-16
Let GSK validate the CN and SAN domains of server certificates

ChongZhou-Broadcom:
Related PR: [ZSS: Support verifyCertificates modes: STRICT and NONSTRICT](https://github.com/zowe/zss/pull/873)

In `httpClientSessionInitv2()`, tls socket is initialized without CN/SAN validation.

GSK API is capable of validating the CN and SAN domains, but it's not a default behavior. The peer's hostname must be set to the socket handle by `gsk_attribute_set_buffer()`. The following code snippet is an example:

```c
   ...
   rc = rc || gsk_attribute_set_callback(socket->socketHandle, GSK_IO_CALLBACK, &ioCallbacks);
+  if (!isServer && peerHost) {
+    rc = rc || gsk_attribute_set_buffer(socket->socketHandle, GSK_REFERENCE_ID_DNS, peerHost, 0);
+    rc = rc || gsk_attribute_set_buffer(socket->socketHandle, GSK_REFERENCE_ID_CN, peerHost, 0);
+    rc = rc || gsk_attribute_set_enum(socket->socketHandle, GSK_WILDCARD_VALIDATION_ENABLE, +GSK_WILDCARD_VALIDATION_ENABLE_ON);
+  }
   rc = rc || gsk_secure_socket_init(socket->socketHandle);
   ...
```

@ifakhrutdinov Irek said he will be doing some research for this. I created this draft PR for him as a reference.

I tried this change with ZSS. I used the following configuration to test if the hostname validation really takes effect:

```yaml
  zss:
    enabled: true
    port: 7557
    crossMemoryServerName: ZWESIS_STD
    agent:
      jwt:
        fallback: true
      64bit: true
      mediationLayer:
        server:
          gatewayHostname: 1.2.3.4
          #gatewayHostname: my.real_domain.com
          gatewayPort: 7554
```

Without this change, ZSS worked with both `gatewayHostname` values. With this change, only the real host name worked, the IP address, which is not in my certificate, no longer works.


ifakhrutdinov:
@ChongZhou-Broadcom , thanks for opening this; we'll discuss the changes.

ChongZhou-Broadcom:
Now I see the real pain point of this change --- how do we make the domain validation optional?

In other words, how do we allow ZSS to handle the `zowe.verifyCertificates: STRICT | NONSTRICT`?

Theoretically, the best way is to add a boolean or bit-wise flag in `TlsSettings`. But we can't change already-released APIs. If we create `TlsSettings2`, it will require a lot of changes in both common-c and ZSS.

So I will leave it to you guys to decide. If you agree with creating `TlsSettings2` and updating both repos, I will be happy to do _(but I will do it in September due to my vacation plan)_. Otherwise if you have other better ideas, I will be very happy to know.

Thanks.

1000TurquoisePogs:
Correct @ChongZhou-Broadcom, we need to respect `zowe.verifyCertificates`.
zowe-common-c isn't supposed to know about yaml reading, that's what zss does instead.

So I believe to implement this security feature, which is good, you need to further enhance this to be 2 PRs

1) a new pr, in zss, in which the parameter is read, and passed down to
2) this pr, which is enhanced, to consider the value.

You have 
`tlsSocketInit2(TlsEnvironment *env, TlsSocket **outSocket, int fd, bool isServer, const char *peerHost)`
But I think you want to further enhance it like

```
#define VERIFY_CERTIFICATES_FULLY 0
#define VERIFY_CERTIFICATES_NO_DOMAIN_CHECK 1
#define VERIFY_CERTIFICATES_OFF 2
```

`tlsSocketInit2(TlsEnvironment *env, TlsSocket **outSocket, int fd, bool isServer, const char *peerHost, int certVerification)`

c/qjsnet.c
```diff
@@ -171,6 +173,15 @@ static JSValue netTLSClientConfig(JSContext *ctx, JSValueConst this_val,
   if (!JS_IsUndefined(label)){
     defaultTLSSettings->label = nativeString(ctx,label,NULL);
   }
+  defaultTLSSettings->certVerify = TLS_CERTVERIFY_DISABLED;
+  if (!JS_IsUndefined(certVerify)){
+    const char* certVerifyText = nativeString(ctx,certVerify,NULL);
+    if (certVerifyText && 0 == strcmp(certVerifyText, "strict")) {
+        defaultTLSSettings->certVerify = TLS_CERTVERIFY_STRICT;
+    } else if if (certVerifyText && 0 == strcmp(certVerifyText, "nonstrict")) {
```
Martin-Zeithaml: If this compare fine? The YAML key `zowe.verifyCertificates` is defined as
```
"enum": ["STRICT", "NONSTRICT", "DISABLED"]
```

So the values are uppercased.
Just double check this, if the values were lowercased somewhere, then it is fine.

ChongZhou-Broadcom: Thank you for the feedback. I will change it to case-insensitive comparison soon.

Martin-Zeithaml: Also the sonar warning seems to be right, there L181, `if` is twice.

ChongZhou-Broadcom: Fixed ... but these aren't the biggest problem at the moment :-( ... GSK doesn't seem to support "DISABLED". I'll have to discuss it with the squad.

ChongZhou-Broadcom:
@1000TurquoisePogs

Hi Sean, at the moment, I implemented what I can implement with GSK, but I don't think GSK allows to skip certificate verification, which means the "DISABLED" mode is quite hard to implement.

Theoretically, we can use libSSL for the "DISABLED" mode, but I'm not sure if it is worth doing --- paying a high price to implement a feature that few people need.

The current behavior for "DISABLED": force to "NONSTRICT" and write log.

1000TurquoisePogs (commented):
This looks like a solid improvement, but I'd like to test ip vs hostname when the build system returns.

PR #617, opened 2026-06-17
httpfileservice: check if double length is fine

Martin-Zeithaml:
* New constant for base64 oprations 256MB 
* Check if `contentLength` > 0 and < 256MB
* We can safely double the `contentLength`
* `DEBUG3` to `WARNING`


JoeNemo:
Notes from a closer look, so they do not get lost:

- The +7 in the check does not match the allocation, which stays 2 * contentLength. The doubleword rounding it anticipates happens inside GETMAIN/STORAGE OBTAIN, below the API, so the check can simply be (int64)contentLength * 2 <= INT_MAX. Better still, use a realistic request cap well under the 31-bit region: INT_MAX admits sizes that just turn the overflow into a failed malloc31, and these functions do not check safeMalloc results (#642).
- The output side has the same overflow one step later: resultBufferSize = (conversionLength * 3) / 4 wraps before the divide. Out of scope here, filed as #641.
- contentLength == 0 is not a compatibility concern. The old code also returned -1 for empty content (via the decodedLength > 0 check), and along the way did strlen over a zero-byte unzeroed buffer, so rejecting early is strictly better. Consider WARN instead of DEBUG for the rejection so a failed upload is visible to operators.

JoeNemo (approved):
Approved.

PR #618, opened 2026-06-17
Respect MAX_JSON_KEY size

Martin-Zeithaml:
Internally we have a limit `MAX_JSON_KEY` is `256`.

When deleting from JSON, working buffer is expecting `MAX_JSON_KEY` length, but this is not properly checked.


JoeNemo:
Verified on WSL under AddressSanitizer, linking the real `json.c` from this
branch and from `v3.x/staging`, driven through the public `jsonDelete()`.

The bug is real and the fix holds. Staging, deleting a bracketed key path:

```
  keyLen=400  ERROR: AddressSanitizer: stack-buffer-overflow   WRITE of size 401
                #1 deleteJson  c/json.c:2897  (the strncpy)
                #2 jsonDelete  c/json.c:3084
  keyLen=260  ERROR: AddressSanitizer: stack-buffer-overflow
  keyLen=255  ERROR: AddressSanitizer: stack-buffer-overflow
  keyLen=100  survived
```

With this PR, all four survive.

Worth recording the exact threshold, because it is one lower than it looks. The
buffer is `char workStr[MAX_JSON_KEY]` = 256, and

```c
strncpy(workStr, jsonTok, strlen(jsonTok));
```

is bounded by the **source** length, so a 256-character key fills the buffer
exactly and leaves no room for a terminator -- the `strlen(workStr)` on the next
line then runs off the end. So the first failing input is a key path of 256
characters, not 257.

Reachable from `cfgDeleteFromConfiguration()` in `c/configmgr.c:892`, which is
the only caller of `jsonDelete()` in either repo, so whether this is exposed
depends on where that key path originates.

Two non-blocking observations:

1. The two overflow guards bail differently. The first (`snprintf` into an empty
   buffer) `break`s the **outer** walk, so the delete does not happen at all.
   The second (the append) `break`s the **inner** loop and then falls through to
   the bracket-strip using a `workStr` that `snprintf` has already truncated. So
   an over-long composite key is used in partial form rather than abandoned. It
   cannot overflow either way, and a key that long will not match anything real,
   so this is a consistency note rather than a defect.

2. Both paths fail silently -- `cfgDeleteFromConfiguration()` documents that it
   always succeeds, so a caller cannot distinguish "key absent" from "key too
   long to process". That matches the existing contract, so changing it would be
   a separate discussion.

No test in the PR. I have the ASan harness above as a self-contained file if you
want it in `tests/` -- happy to hand it over rather than push it here, since it
is your PR.

LGTM.

JoeNemo (approved):
Approved.

PR #619, opened 2026-06-26
Fix swapped Min/Max TLS log labels

struga0258:
The getTlsMax() and getTlsMin() functions in c/tls.c had their log messages swapped, getTlsMax() was printing "Min TLS requested" and getTlsMin() was printing "Max TLS requested". This makes debugging TLS configuration issues unnecessarily confusing.


Martin-Zeithaml (approved):

ifakhrutdinov (approved):
lgtm

PR #620, opened 2026-06-30
Update `SLHAlloc` to check the limit

Martin-Zeithaml:
For cases like #613 

## Change
* `SLHAlloc` rounds the memory up to 8 byte boundary
* When calling `SLHAlloc` with size close to `INT_MAX`, it can possibly overflow and negative int will be later treated as unsigned int (large number).
* Before the change, caller should check, if the rounded size is <= INT_MAX
* After the change, this check is part of the `SLHAlloc`

### DIFF
--- a/c/utils.c
+++ b/c/utils.c
@@ -12,6 +12,7 @@
 
 #ifdef METTLE
 #include <metal/metal.h>
+#include <metal/limits.h>
 #include <metal/stddef.h>
 #include <metal/stdio.h>
 #include <metal/stdlib.h>
@@ -21,6 +22,7 @@
 #include "metalio.h"
 
 #else
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


c/utils.c
```diff
@@ -1493,10 +1495,18 @@ ShortLivedHeap *makeShortLivedHeap64(int blockSize, int maxBlocks){
 }
 
 char *SLHAlloc(ShortLivedHeap *slh, int size){
+  if (size <= 0) {
```
ifakhrutdinov, line 1498: I'm wondering if there is code out there which relies on the fact that size 0 would not return `NULL`.

JoeNemo: Do we really want an SLHAlloc to allocate more than 1GB.  Shouldn't the limit and warning be for something lower than INT_MAX.

Martin-Zeithaml: We can limit it, but IMO it should be done in V4 (this is public API).

JoeNemo: You are technically right.  And that's the best kind of right.  😄

JoeNemo (approved):
Approved.

PR #627, opened 2026-07-07
Use cookieNameLength in memcmp

Martin-Zeithaml:
* Bounds the `=` search to the current key/value segment (`semiPos`), fixing the boundary issue.
* Requires the candidate key's length to exactly equal `cookieNameLength` before comparing, then always compares exactly `cookieNameLength` bytes — eliminating both the **prefix-collision** false match and the **out-of-bounds** read.


JoeNemo (approved):
Approved.

PR #630, opened 2026-07-08
Fix zss#828 streaming charset conversion; route Open XL (ibm-clang64) to iconv

### LINKED ISSUE #828: Codepage-tagged file contents returned incorrectly (SHA256 mismatches)
ZSS /unixfile/contents/ endpoint returns incorrect or empty data for files with -tc (text conversion) or -mc (mixed conversion) codepage tags (ibm1047, utf8). SHA256 of response doesn't match the written file.

JoeNemo:
# PR: Fix zss#828 streaming charset conversion; route Open XL to iconv

Builds on cc#592/#593 (rebased onto v3.x/staging). Fixes zowe/zss#828 — a UTF-8
`/unixfile` GET returns an empty/truncated body — and an Open XL compiler-branch
bug found underneath it.

## #828 conversion mechanics — `charsets.c`, `httpserver.c`
- **`convertCharsetStreaming`** (new): converts fixed read buffers; a multibyte
  char straddling a buffer boundary is held and prepended to the next read; a
  char unmappable in the target is **substituted** (iconv SUB `0x1A` / glibc
  `?`) instead of aborting to an empty body.
- **`getCharsetName`**: `"ISO-8859-1"` → `"ISO8859-1"` — z/OS `iconv_open`
  rejects the dashed name (errno 121).
- **`streamTextForFile2`** uses it; an incomplete tail is dropped only at true EOF.

## Open XL branch fix — `charsets.c`  *(real root cause; bigger than #828)*
- ibm-clang64 sets `__clang__` but not `__IBMC__` → `__ZOWE_COMP_CLANG`, not
  `__ZOWE_COMP_XLCLANG`.
- The z/OS branch split tested `XLCLANG` only, so ibm-clang64 fell through to
  **metal/CUNLCNV** — consumes incomplete sequences, drops unmappable chars, no
  streaming. Now routed to **iconv** alongside xlclang.

## Tests
- `tests/charset-streaming` — exhaustive straddle sweep (every chunk size) +
  substitution + EOF/invalid-byte edges; under ASan.
- `tests/clang_readiness/test_charconv` (replaces `test_troo`) — behavioral
  multibyte/straddle/unmappable under **both** compilers; **fails on CUNLCNV**,
  so a wrong-branch regression can't hide.

## Validation
WSL **9/9** ASan · z/OS **9/9** real iconv (ibm-clang64) · Linux test_charconv 5/5.

## For reviewers
- **1000TurquoisePogs:** encoding *choice* stays cc#593 (source-aware:
  multibyte→UTF-8, single-byte→819); this fixes the *mechanics*.
- **ifakhrutdinov:** z/OS iconv is textbook POSIX here — fix leans on `EINVAL`
  (carry) + native SUB.
- **Deferred:** `unmappable=substitute|fail` param; base64 multi-chunk;
  `charset=UTF-8` response label. The `__ZOWE_COMP_CLANG` gap likely recurs
  elsewhere → separate common-c audit.


c/httpserver.c
```diff
@@ -4475,64 +4475,113 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+      }
+      if (targetParam == NULL) {
+        targetParam = getQueryParam(response->request, "target");
+      }
+
+      int callerSourceCCSID = 0; /* 0 = auto */
+      int callerTargetCCSID = 0; /* 0 = auto */
+
+      if (sourceParam != NULL && strcasecmp(sourceParam, "auto") != 0) {
+        callerSourceCCSID = parseEncodingValue(sourceParam);
```
ifakhrutdinov, line 4501: `parseEncodingValue` is zOS/AIX/Unix only, but our http server is buildable on Win. I think this will lead to a link error.

c/charsets.c
```diff
@@ -35,45 +35,163 @@
...
-  else if (!strcmp(localArray, "UTF-16BE")) {
-    return CCSID_UTF_16_BE;
+
+  /* "binary" is a special sentinel meaning CCSID_BINARY (0xFFFF) */
+  if (strlen(value) <= CHARSETNAME_SIZE) {
+    strcpy(localArray, value);
+    strupcase(localArray);
+    if (!strcmp(localArray, "BINARY")) {
+      return (unsigned short)CCSID_BINARY;
+    }
```
ifakhrutdinov: Side note: `stricmp` (case-insensitive `strcmp`) would be handy here; maybe we should implement that and add to the library.

c/charsets.c
```diff
@@ -35,45 +35,163 @@
...
+  int code = getCharsetCode(value);
+  if (code != -1) {
+    return code;
   }
-  else {
-    return -1;
+
+  /* Fall back to decimal integer string (e.g. "1047", "819") */
+  int n = 0;
+  if (sscanf(value, "%d", &n) == 1 && n >= 1 && n <= 65535) {
```
ifakhrutdinov: `sscanf` is quite unreliable as it will work for strings like `1047foo`. 

According to the webs, something like this is the safest way to parse an int:
```c
#include <stddef.h>
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>
#include <errno.h>
#include <string.h>

static int parseInt(const char *str, int *out) {
  if (str == NULL) {
    return -1;
  }

  errno = 0;
  char *end;
  long value = strtol(str, &end, 10);

  if (end == str || *end != '\0') {
    return -1;
  }

  if (errno == ERANGE || value < INT32_MIN || value > INT32_MAX) {
    return -1;
  }

  *out = value;
  return 0;
}

int main(int argc, char **argv) {

    int result = -1;
    int rc = parseInt(argv[1], &result);
    printf("%d, rc=%d\n", result, rc);

    return 0;
}

```

c/charsets.c
```diff
@@ -406,7 +528,7 @@ int convertCharset(char *input,
 static const char *getCharsetName(int ibmCode){
   switch (ibmCode){
   case CCSID_ISO_8859_1:
-    return "ISO-8859-1";
+    return "ISO8859-1"; /* NO dash after ISO: z/OS iconv rejects "ISO-8859-1" (errno 121); glibc accepts both */
```
ifakhrutdinov: Nit: not sure if we need a comment like this here; the commit log is probably a better place.

The reason is that it explains a bug fix (a change), but once it's fixed, such comments are not relevant to someone who's reading the code because changes and the purpose for those changes are better captured by git.

Or is there anything significant about `ISO-8859-1` specifically?

c/charsets.c
```diff
@@ -524,10 +646,131 @@ int convertCharset(char *input,
   }
   return result;
 }
-#else  
+
+/* Streaming charset conversion for the file-content path (USE_BUFFER only).
+ * Converts as much of [input, inputLength] as forms COMPLETE characters into
+ * the caller's output buffer, and reports BOTH how much output was produced
```
ifakhrutdinov: Using caps for both the words that do require that (like "EINVAL" and `//TRANSLIT`) and just normal words (like "BOTH", "COMPLETE", and "NOT") is confusing.

c/charsets.c
```diff
@@ -524,10 +646,131 @@ int convertCharset(char *input,
...
+  /* "//TRANSLIT" is a glibc transliteration extension: on glibc it gives clean
+   * per-character substitution for characters unmappable in the target. z/OS
+   * iconv does NOT implement it as a modifier -- it opens "NAME//TRANSLIT" as a
+   * distinct, malfunctioning converter (drops data, mishandles incomplete
+   * sequences) -- and z/OS iconv already substitutes unmappable characters
+   * natively (SUB, 0x1A). So attempt //TRANSLIT only on glibc; everywhere else
+   * use the plain converter, whose EILSEQ we substitute for ourselves below. */
+  iconv_t converter = (iconv_t) -1;
+#if defined(__ZOWE_OS_LINUX)
+  char translitCharset[64];
```
ifakhrutdinov: Why hardcode 64? I think you can use `CHARSETNAME_SIZE + strlen("//TRANSLIT") + 1` (I guess `//TRANSLIT` can be `#define`'d.

Something like this:
```c
#define TRANSLIT_OPT "//TRANSLIT"
char translitCharset[CHARSETNAME_SIZE + strlen(TRANSLIT_OPT) + 1];
snprintf(translitCharset, sizeof(translitCharset), "%s"TRANSLIT_OPT, outputCharset);
```

And you may want to check the return code from `snprintf` for overflow and output errors (negative rc).

h/charsets.h
```diff
@@ -93,9 +110,18 @@ int convertCharset(char *input,
...
                    ShortLivedHeap *slh, // optional
-                   int *conversionOutputLength, 
+                   int *conversionOutputLength,
                    int *reasonCode);
 
+/* Streaming variant used by the file-content path: converts only COMPLETE
+ * characters, reports input bytes consumed so the caller can carry a partial
+ * multibyte sequence forward across read buffers, and substitutes unmappable
+ * characters instead of failing. See convertCharsetStreaming in charsets.c. */
+int convertCharsetStreaming(char *input, int inputLength, int inputCCSID,
```
ifakhrutdinov, line 187: A nit, but I'd use doxygen format for the function doc as we can then render that as real documentation:
```c
/**
 * @brief Streaming charset conversion for the file-content path (USE_BUFFER only).
 *
 * Converts as much of @p input (up to @p inputLength bytes) as forms complete
 * characters into the caller's output buffer, reporting both how much output was
 * produced and how much input was consumed.
 *
 * A trailing incomplete multibyte sequence (iconv @c EINVAL) is intentionally
 * left unconsumed and is not treated as an error: the caller carries those
 * leftover bytes (@p inputLength minus @p *inputBytesConsumed) forward and
 * prepends them to the next read, which fixes multibyte characters split across
 * a read-buffer boundary.
 *
 * Characters the target cannot represent are substituted rather than aborting
 * the whole response (the issue #828 silent empty body): the @c //TRANSLIT
 * suffix handles characters unmappable in the target encoding, and an explicit
 * '?' is emitted for bytes that are invalid in the source encoding, so a single
 * bad byte cannot stall the stream.
 *
 * @param[in]  input                   Bytes to convert.
 * @param[in]  inputLength             Number of bytes available in @p input.
 * @param[in]  inputCCSID              CCSID of the source encoding.
 * @param[out] output                  Buffer receiving the converted bytes.
 * @param[in]  outputLength            Capacity of @p output in bytes.
 * @param[in]  outputCCSID             CCSID of the target encoding.
 * @param[out] conversionOutputLength  Number of bytes written to @p output.
 * @param[out] inputBytesConsumed      Number of bytes consumed from @p input.
 * @param[out] reasonCode              Receives the underlying errno on converter
 *                                     open failure; 0 otherwise.
 *
 * @retval CHARSET_CONVERSION_SUCCESS       (0)  Conversion progressed; a trailing
 *         incomplete multibyte sequence left unconsumed is reported here, not as
 *         an error.
 * @retval CHARSET_SHORT_BUFFER             (8)  Output buffer was filled before
 *         all input was consumed; drain @p output and call again.
 * @retval CHARSET_INTERNAL_ERROR           (12) iconv reported an unexpected errno.
 * @retval CHARSET_CONVERSION_UNIMPLEMENTED (20) Converter could not be opened;
 *         @p reasonCode holds the errno.
 * @retval CHARSET_UNKNOWN_CCSID            (24) @p inputCCSID or @p outputCCSID
 *         is not recognized.
 *
 * @note On platforms without a native iconv streaming converter, the portable
 *       fallback delegates to convertCharset and may additionally return
 *       CHARSET_INTERNAL_ERROR (12) or CHARSET_CONVERSION_ROUTINE_FAILURE (16).
 *       That fallback reports all input as consumed (no carry-forward).
 *
 * @return One of the CHARSET_* status codes above.
 */
int convertCharsetStreaming(char *input, int inputLength, int inputCCSID,
                            char *output, int outputLength, int outputCCSID,
                            int *conversionOutputLength, int *inputBytesConsumed,
                            int *reasonCode);
```

c/charsets.c
ifakhrutdinov: Actually, why not use `strcasecmp` as in `httpserver.c`?

Then the whole thing will be just:
```c
if (!strcasecmp((value), "binary")) {
  return (unsigned short)CCSID_BINARY;
}
```


No need for `localArray`, `strupcase`, and the length check.

c/httpserver.c
```diff
@@ -4475,64 +4475,113 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
-      streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, NATIVE_CODEPAGE, webCodePage, asB64);
-    }
-    else {
-	    zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG, "Sending with tagged conversion between %d and %d\n", 
-              ccsid, webCodePage);
-      streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, ccsid, webCodePage, asB64);
-    }
+
+      bool srcIsBinary = (callerSourceCCSID != 0 &&
+                          (unsigned short)callerSourceCCSID == (unsigned short)CCSID_BINARY);
```
ifakhrutdinov: Doesn't `(unsigned short)callerSourceCCSID == (unsigned short)CCSID_BINARY` imply `callerSourceCCSID != 0`?

And the same applies to `tgtIsBinary`.

ifakhrutdinov: In other words, this can be:
```c
bool srcIsBinary = (short)callerSourceCCSID == CCSID_BINARY;
```
_I think `CCSID_BINARY` is already a short, so no need to cast._

c/charsets.c
```diff
@@ -406,7 +528,7 @@ int convertCharset(char *input,
 static const char *getCharsetName(int ibmCode){
   switch (ibmCode){
   case CCSID_ISO_8859_1:
-    return "ISO-8859-1";
+    return "ISO8859-1"; /* NO dash after ISO: z/OS iconv rejects "ISO-8859-1" (errno 121); glibc accepts both */
   case CCSID_IBM1047:
     return "IBM-1047";
   case CCSID_UTF_8:
```
ifakhrutdinov, line 538: The new function `getCharsetCode` handles lots of charsets, but this one just 6. Should we make these symmetric? I'm trying to see if this can lead to issues otherwise.

c/httpserver.c
```diff
@@ -4749,64 +4798,73 @@ static int streamTextForFile2(HttpResponse *response, Socket *socket, UnixFile *
...
+
         if (TRACE_CHARSET_CONVERSION){
-          printf("convertCharset transLen=%d\n",translationLength);
+          printf("convertCharsetStreaming transLen=%d consumed=%d/%d rc=%d\n",
+                 translationLength, inputConsumed, inTotal, rc);
           dumpbuffer(translation,translationLength);
         }
         if (rc != 0){
-          zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG, "iconv rc = %d, bytesRead=%d xlateLength=%d\n",rc,bytesRead,translationLength);
+          zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG,
```
ifakhrutdinov: Looks like we're masking errors here.

If `convertCharsetStreaming` returns a non-zero RC, `inputConsumed` will be 0,`leftover` will be equal to `inTotal` skipping that `if`-body, and `translation` will have no content which will lead to sending an empty body.

The simplest case `convertCharsetStreaming` would return a bad RC is when `getCharsetName` returns `NULL` because we passed an unknown (to `getCharsetName`) CCSID, or if `iconv` returns `E2BIG`.

c/charsets.c
```diff
@@ -524,10 +646,131 @@ int convertCharset(char *input,
...
+
+  while (inLeft > 0){
+    size_t status = iconv(converter, &inPtr, &inLeft, &outPtr, &outLeft);
+    if (status != (size_t) -1){
+      break; /* all remaining input consumed */
+    }
+    if (errno == EINVAL){
+      break; /* incomplete trailing sequence -> caller carries it forward */
+    } else if (errno == E2BIG){
+      result = CHARSET_SHORT_BUFFER; /* output full; caller has what fit */
```
ifakhrutdinov, line 778: If this happens, we'll end up truncating the output in `streamTextForFile2` after the `convertCharsetStreaming` call. And I guess it can happen because the output buffer we pass is just twice the size of the input buffer, but the characters may probably get 3x - 4x of the original size in some target UTF8, no?

c/charsets.c
```diff
@@ -524,10 +646,131 @@ int convertCharset(char *input,
...
+      break;
+    } else if (errno == EILSEQ){
+      /* byte not valid in the source encoding: substitute and skip it so a
+         single bad byte cannot stall the whole stream. */
+      if (outLeft == 0){ result = CHARSET_SHORT_BUFFER; break; }
+      *outPtr++ = 0x3F; /* '?' as a numeric byte: correct in ASCII-family targets
```
ifakhrutdinov: Should we then use `0x6f` when it's EBCDIC?

ifakhrutdinov (commented):
@JoeNemo , @1000TurquoisePogs , please see my comments.

JoeNemo:
Pushed b6f5831 addressing the three main review findings:

1. **Masked converter errors -> empty body** (httpserver.c:4858): the resolved source/target pair is now validated **before the response status is committed**, via a new `isCharsetStreamingPairSupported()` (trial `iconv_open`, permissive on builds whose converter takes numeric CCSIDs). A forced unusable pair (e.g. `?targetEncoding=IBM-37`) is rejected with **400**; a default GET of a file whose CCSID tag cannot be converted on the build falls back to raw binary streaming with a warning; and a mid-stream hard error now logs at SEVERE and aborts the stream instead of being ignored.

2. **E2BIG / expansion past the 2x buffer -> truncation** (charsets.c:713): the streaming loop now drains-and-recalls on `CHARSET_SHORT_BUFFER`, emitting once per converter call, so a target that expands beyond the translation buffer cannot lose data. That is also what makes it safe to ever extend `getCharsetName` with targets that expand more than 2x (e.g. IBM-838 Thai -> UTF-8).

3. **getCharsetCode vs getCharsetName asymmetry** (charsets.c:534): kept deliberately asymmetric - the code table parses caller-supplied names, while the name table only holds spellings verified against BOTH z/OS iconv and glibc - but that contract is now documented on the function, and the pair check above turns the asymmetry from a reachable empty body into the 400.

Tests: `tests/charset-streaming/error-paths-test.c` - 13 checks covering the pair-guard contract, hard-error surfacing (the model server loop returns a failure instead of a quiet empty success), drain-and-recall delivering all 6 bytes where the old loop delivered 4, and a straddle regression - wired into `run.sh` (ASan) and `run-zos.sh`. Existing suites unchanged and green: charset-streaming 9/9, getcharsetcode 136/136, clang_readiness charconv 5/5.

End-to-end on a native-Linux ZSS built from this branch: `?source=1208&target=37` now returns **HTTP 400** ("sourceEncoding/targetEncoding pair is not supported for conversion") where it previously returned 200 with an empty body; identity, lossy-819, and default GETs are byte-identical to before.

The smaller comments (sscanf parsing, //TRANSLIT buffer sizing, the 0x3F-vs-0x6F substitute on EBCDIC targets, doxygen, strcasecmp, the Windows link concern) are queued as follow-ups to keep this change reviewable.

JoeNemo:
z/OS validation of b6f5831 (Marist, ibm-clang64, real z/OS iconv), via tests/charset-streaming/run-zos.sh:

```
charset-streaming-test   ALL PASS  (9/9)
getcharsetcode-test      ALL PASS  (136/136)
error-paths-test         ALL PASS  (13/13)   <- new: pair guard, hard-error surfacing, SHORT_BUFFER drain
```

Same results as Linux/ASan. With the earlier end-to-end check (forced unusable pair -> 400 through a live server), all three review findings are covered by tests on both platforms.

c/httpserver.c
```diff
@@ -4475,64 +4475,139 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
...
+        /* both explicit: use caller's conversion directly */
+        zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG,
+                "Streaming with caller-supplied conversion %d->%d for %s\n",
+                callerSourceCCSID, callerTargetCCSID, absolutePath);
+        streamTextForFile2(response, NULL, in, ENCODING_CHUNKED, callerSourceCCSID, callerTargetCCSID, asB64);
+      } else if (callerSourceCCSID != 0) {
+        /* source explicit, target auto: convert from caller source to webCodePage */
+        zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG,
+                "Streaming with caller source %d->webCodePage %d for %s\n",
+                callerSourceCCSID, webCodePage, absolutePath);
```
ifakhrutdinov, line 4586: Here we use `webCodePage` as the target, and since `webCodePage` is derived from the source file tag, we use the file tag as the target encoding; however, we also have an explicit source. Wouldn't it make more sense to derive the target from the explicit source and not the file tag when the target is auto?

c/httpserver.c
```diff
@@ -4548,12 +4545,41 @@ void respondWithUnixFile2(HttpService* service, HttpResponse* response, char* ab
 #error Unknown OS
 #endif
 
-      bool srcIsBinary = (callerSourceCCSID != 0 &&
-                          (unsigned short)callerSourceCCSID == (unsigned short)CCSID_BINARY);
-      bool tgtIsBinary = (callerTargetCCSID != 0 &&
-                          (unsigned short)callerTargetCCSID == (unsigned short)CCSID_BINARY);
+      /* CCSID_BINARY is (short)0xFFFF, so the cast comparison alone suffices:
+         0 (auto) can never equal it (review simplification). */
```
ifakhrutdinov: We don't need comments about issues introduced and fixed within the same PR.

JoeNemo:
Confirmed and fixed in 48d3ef8 -- the base64 finding was real, and older than the drain loop: a single conversion emit already had arbitrary length, so any converted response larger than one read buffer could carry mid-stream padding; the drain loop just made it easier to hit. The %3 read sizing only ever protected the identity path.

Fix: the base64 path now emits only the 3-aligned prefix of each span and holds the 0-2 byte tail in a small carry, flushed once at end of stream, so the final group carries the body's only padding. Identity and conversion paths share the carry (which also fixes the identity path's short-read case); RAW is untouched.

Demonstrated A/B on a live server with a 1,224,000-byte file converted 819->UTF-8 (first-read output = 2 mod 3): before, "=" appeared at offset 1,365,334 of the body and strict decoders fail; after, the body is strictly valid base64 and decodes byte-for-byte -- 1,224,002 bytes with the expansion exactly at the read-boundary offset. Nice catch.

ifakhrutdinov (approved):
I'm approving this, but please, let's keep the PR size smaller next time. And when PRs are smaller, we can even avoid using LLMs for writing code (the tests I'm fine with).

It's really difficult to review LLm-written pull-requests; just the comment verbosity is staggering.

PR #631, opened 2026-07-08
Bugfix: run the JS async continuation loop after configmgr script eval

JoeNemo:
## Problem

configmgr evaluates a top-level script (`ejsEvalFile` — used by `-script`, `-embeddedScript`, and any embedder such as a launcher) and returns **without ever running the QuickJS event loop**: `js_std_loop()` is declared but never called. QuickJS dispatches `os.signal()` handlers, `os.sleepAsync()` timers, and Promise jobs **only** inside `js_os_poll()`, which is reached only from `js_std_loop()`. So every asynchronous continuation scheduled during eval is silently dropped — a JS component that registers a `SIGTERM` handler for graceful shutdown never sees the signal.

This is **platform-independent** (reproduces on Linux, not only z/OS): a synchronous script "works" only because all of its work happens during the straight-line eval.

## Fix

One call — `js_std_loop(ejs->ctx)` after the eval in `ejsEvalFile()` (`c/embeddedjs.c`) — fulfilling the single-threaded JS event-loop contract: after top-level eval, pump microtasks + macrotasks until the loop is idle.

- Purely synchronous script → nothing pending → returns immediately (no behavior change).
- Async / signal-waiting script → driven to completion.

## Tests (`tests/js/`)

- **`async-sleep.js`** — the correct pattern: an `await os.sleepAsync` loop that yields to the event loop each tick. Catches `SIGTERM` → **PASS** with the fix (and produces *no ticks at all* without it, since the loop never runs).
- **`sync-sleep.js`** — negative control: a synchronous `os.sleep` loop blocks the one JS thread and starves the loop, so the handler can never fire in time — true on **any** JS host (Node/Deno/Bun included), and true even with this fix. Kept to document the anti-pattern.

## Verification (native Linux/WSL, `build_cmgr_clang.sh linux`)

| build × script | result |
|---|---|
| **fixed** × async | ticks 1/s, `*** SIGTERM handler fired ***`, **PASS** |
| **fixed** × sync | 10 ticks, `stop` never flips → **FAIL**; handler fires only in the post-eval drain (too late) |
| **unfixed** × async | pid line only, **no ticks** — loop never runs |

A long-standing design-level gap: QuickJS underlies Zowe install/config, but nothing had needed the *asynchronous* half of the JS contract until signal-driven graceful shutdown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Zowe Common C Changelog
 
 ## `3.6.0`
+- Bugfix: configmgr now runs the JavaScript asynchronous continuation loop (`js_std_loop`) after evaluating a script, so `os.signal` handlers, `os.sleepAsync` timers, and Promise jobs scheduled during eval are dispatched instead of silently dropped. This lets JS-based components (e.g. a launcher) catch `SIGTERM` for graceful shutdown; a purely synchronous script is unaffected (the loop returns immediately when nothing is pending). [(#631)](https://github.com/zowe/zowe-common-c/pull/631)
 - Bugfix: Internal Short Lived Heap allocation routine `SLHAlloc` checks the size [(#620)](https://github.com/zowe/zowe-common-c/issues/620)
 - Bugfix: Return code 414 (`HTTP_STATUS_URI_TOO_LONG`) for too long URI [(#597)](https://github.com/zowe/zowe-common-c/issues/597)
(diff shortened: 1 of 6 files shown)


Martin-Zeithaml (approved):

Gautham-coder (approved):
Tested it and the signal handler is called correctly.

PR #636, opened 2026-07-10
Validate pointer after alloc in jsonParserAlloc and jsonTokenizerAlloc

ChongZhou-Broadcom:
Just a very very small fix, so I will skip unit test hints this time. Thanks.


JoeNemo (approved):
Approved

PR #638, opened 2026-07-14
Restrict key type match in checkSignature

ChongZhou-Broadcom:
`checkSignature()` is vulnerable to "algorithm-confusion" attack: if a client uses the public key and a symmetric algorithm to fake a signature, and tempers the "sig" field of the JWT to the used symmetric algorithm, then `checkSignature()` might successfully validate the tempered token.

Although ICSF will probably refuse to validate it because it will probably find out the provided key handle is not a symmetric key, but this means the safety is riding on ICSF's key-class check, the behavior of which is not guaranteed in the future.

With some small code changes, we can make it guaranteed by our own code. I think it's worth doing.


JoeNemo:
How are we guaranteeing the exhaustiveness of these (very good) additional tests against all length variants of the algorithms?

ChongZhou-Broadcom:
> How are we guaranteeing the exhaustiveness of these (very good) additional tests against all length variants of the algorithms?

I moved those tests out of the switch..case. Now it checks algorithm id ranges.

JoeNemo (approved):
Approved.

PR #640, opened 2026-07-15
Added extra length limit to http chunk receiving

ChongZhou-Broadcom:
When `processHttpFragment()` is receiving HTTP content chunks, it allocates space from a caller-provided SLH. Theoretically it could be very large, so the following addition is not guaranteed to be less than INT_MAX:

```c
parser->content = SLHAlloc(parser->slh, parser->specifiedContentLength + parser->chunkSize);
```

So, I added a check to prevent the addition from overflowing `INT_MAX`.

Hints for unit test is skipped because this is just a very rare corner case.


ChongZhou-Broadcom:
**Action plan:** per the discussion in the squad meeting, I will try to create a scenario where the added if-branch can be entered.

ifakhrutdinov:
@ChongZhou-Broadcom , is this something that can happen? Would it be more realistic to put a limit on the body size somewhere upstream?

Why pick this particular place? I'm thinking that there are probably lots of potential integer overflows in this code base and we may want to address that in some coordinated manner. Or perhaps there aren't any integer overflows left..

ChongZhou-Broadcom:
> @ChongZhou-Broadcom , is this something that can happen? Would it be more realistic to put a limit on the body size somewhere upstream? 

There is a similar check in the "upstream" written by Sean many years ago:

https://github.com/zowe/zowe-common-c/blob/3f7368d8bb7d2b3afba463168ab90b034a27f787/c/httpserver.c#L2385-L2394

However, it checks whether the new chunk length is about to overflow.

After that, when it sees a CR, it will move the state to `HTTP_STATE_CHUNK_SIZE_CR_SEEN`. In the next iteration, the state machine will increase the body buffer to get ready for receiving the next chunk of body. So logically it is a correct place to check whether the body length is about to overflow.

> Why pick this particular place?

Because it is where the body length is updated and overflow (integer wrap-around) might happen.

ifakhrutdinov:
Thanks for the details.

You haven't answered the following:

> is this something that can happen?

Regarding this:
>> Why pick this particular place?

> Because it is where the body length is updated and overflow (integer wrap-around) might happen.

I meant why out of all integer overflows pick this one? Or was this the only one you had found? Or are you going to open a PR for every place with integer overflow?

ChongZhou-Broadcom:
@ifakhrutdinov 

For the first question: no, this won't actually happen in the current version of ZSS due to the space limit of the SLH. It might happen in theory. But I think, since it's a library function, we should not assume the SLH space is always limited, should we? Regarding the second question, I'll answer it in Slack. Thank you for the comment.

ifakhrutdinov (approved):

PR #645, opened 2026-07-21
`httpserver`: check webSocketKey for null

Martin-Zeithaml:
`webSocketKey` or `webSocketKey->nativeValue` could be `NULL`.

Adding check for this situation.


JoeNemo (approved):
Approved.

PR #646, opened 2026-07-21
`util.c`: return -1 if the string has no space for null terminator

Martin-Zeithaml:
## Change
* `nullTerminate` is expecting the caller to provide space for null terminator. 
* It returns a length of null terminated string
* It returns -1
  * If there is no space for NULL terminator (full buffer)
  * If string is NULL
  * If length is <= 0

If the string is NULL, it returns -1, because it is an error.
* We need to differentiate between wrong (NULL) and good (empty) buffer
* If buffer is full of for example spaces, `nullTerminate` returns 0 - it means the buffer is empty and accessible
* If buffer is NULL, it has to return -1 to indicate error

## Simple unit test

See Case 2, the buffer is `HELLO`
* without this fix the buffer will be `HELL` + `0x0`
* with the fix, the buffer is unchanged and returned length is -1

```c
#include <stdio.h>
#include <assert.h>
#include <string.h>

#define SPACE 0x20

int nullTerminate(char *str, int len){
  int i;

  if (str == NULL || len <= 0) {
    return -1;
  }

  for (i = len - 1; i >= 0; i--) {
    if ((str[i] != ' ') && (str[i] != 0)) {
      if (i == len - 1) {  // No room for a null-terminator; return an error code.
        return -1;
      }
      str[i+1] = 0;
      return i + 1;
    }
  }
  str[0] = 0;
  return 0;
}


void test_nullTerminate() {
    // Case 1: Trailing EBCDIC spaces are stripped correctly
    char buf1[10] = {'A', 'B', 'C', SPACE, SPACE, SPACE, SPACE, SPACE, SPACE, SPACE};
    int res1 = nullTerminate(buf1, 10);
    assert(res1 == 3);
    assert(buf1[3] == 0);
    assert(buf1[0] == 'A' && buf1[1] == 'B' && buf1[2] == 'C');

    // Case 2: Buffer completely full of data (Boundary Check)
    char buf2[5] = {'H', 'E', 'L', 'L', 'O'};
    int res2 = nullTerminate(buf2, 5);
    assert(res2 == -1);

    // Case 3: Buffer completely full of EBCDIC spaces
    char buf3[5] = {SPACE, SPACE, SPACE, SPACE, SPACE};
    int res3 = nullTerminate(buf3, 5);
    assert(res3 == 0);
    assert(buf3[0] == 0);

    // Case 4: Already null-terminated data
    char buf4[6] = {'X', 'Y', 0, SPACE, SPACE, SPACE};
    int res4 = nullTerminate(buf4, 6);
    assert(res4 == 2);
    assert(buf4[2] == 0);

    // Case 5: Empty buffer / zero length
    char buf5[1] = {'A'};
    int res5 = nullTerminate(buf5, 0);
    assert(res5 == -1);

    // Case 6: Incorrect length
    char buf6[1] = {'B'};
    int res6 = nullTerminate(buf5, -42);
    assert(res6 == -1);
    
    // Case 7: NULL string
    int res7 = nullTerminate(NULL, -42);
    assert(res7 == -1);

    printf("All nullTerminate tests passed successfully!\n");
}

int main() {
    test_nullTerminate();
    return 0;
}
```


c/utils.c
```diff
@@ -120,13 +120,37 @@ int parseIntSafely(const char *str, int *out){
...
-  for (i=len-1; i>=0; i--){
-    if ((str[i] != 0x40) && (str[i] != 0)){
+  if (str == NULL || len <= 0) {
+    if (str) str[0] = 0;
+    return 0;
+  }
+
+  for (i = len - 1; i >= 0; i--) {
+    if ((str[i] != ' ') && (str[i] != 0)) {
+      /* No extra byte available: the buffer is fully packed with data, so
```
ifakhrutdinov: Not a fan of essays like this for simple logic; you could have just said "No room for a null-terminator; return an error code."

c/utils.c
```diff
@@ -120,13 +120,37 @@ int parseIntSafely(const char *str, int *out){
...
+   Returns the resulting string length (the index of the null byte written),
+   or -1 if str[len-1] itself holds real data: with no blank/null anywhere
+   in the field there is no spare byte to hold the terminator, so the
+   buffer is left untouched rather than overwriting that last byte of data. */
 int nullTerminate(char *str, int len){
   int i;
 
-  for (i=len-1; i>=0; i--){
-    if ((str[i] != 0x40) && (str[i] != 0)){
+  if (str == NULL || len <= 0) {
```
ifakhrutdinov, line 210: In my opinion, it's cleaner and more readable to have separate checks:
```c
if (str == NULL) {
  return 0;
}
if (len <= 0) {
  str[0] = '\0';
  return 0;
}

```

Yeah, there are more lines, but it's ok I think.

ifakhrutdinov, line 210: Now I'm wondering why we'd set `str[0]` to 0 when `len` is equal or smaller than 0? The length indicates the buffer size we want to null-terminate and 0 is clearly "no space" (-1 seem invalid).

c/utils.c
```diff
@@ -120,13 +120,37 @@ int parseIntSafely(const char *str, int *out){
   return 0;
 }
 
+/* Null-terminates a fixed-width, blank-padded field in place. A blank is the
```
ifakhrutdinov: Could this be more concise? 

Consider
```c
/**
 *
 * Null-terminate a blank-padded field in place. A null-terminator is written in
 * the position after the last non-blank or non-null character in the string, or
 * position 0 if none are found.
 *
 * @param[in,out] str the string to null-terminate; if @c NULL or there is no
 * room for a null-terminator (no padding), the function does nothing the
 * string.
 * @param[in] len the length of the string; if the length <= 0, the function
 * does nothing to the string.
 * @returns the length of the resulting C-string or -1 if there was no room
 * for a null-terminator or the length specified was <= 0.
 * */

```
_I assumed `len <= 0` is not valid._

c/utils.c
Martin-Zeithaml: This is exactly the same what the old code would do:
```c
int nullTerminate(char *str, int len){
  int i;
  // Len = -5 ==> skip for loop and set str[0] = 0
  // And does not check, if str is NULL
  for (i=len-1; i>=0; i--){
    if ((str[i] != 0x40) && (str[i] != 0)){
      str[i+1] = 0;
      return i+1;
    }
  }
  str[0] = 0;
  return 0;
}
```

ifakhrutdinov: Yes, but maybe the old code is bad anyway. Have you found any invocations with length <= 0? I'm not asking to change that at this point; I'm just thinking the original behaviour could be improved potentially.

Martin-Zeithaml: The invocation is solid, all I found is like this example, always with extra room for null term and length defined by `sizeof`:
```c
  char ownerNameNullTerm[ZIS_USER_ID_MAX_LENGTH + 1] = {0};
  // update ownerNameNullTerm with memcpy | snprintf | ...
  nullTerminate(ownerNameNullTerm, sizeof(ownerNameNullTerm) - 1);
```

ifakhrutdinov: Thanks, let me discuss this with the team today and I'll let you know if the behaviour needs to change.

ifakhrutdinov (approved):
lgtm

PR #647, opened 2026-07-22
Check decoded length and colon position

Martin-Zeithaml:
* Check `decodeBase64` returning -1 before it is used to index `authString[decodedLength]`
* Changed `if (colonPos)` to `if (colonPos != -1)` for the missing colon
* `memcpy` length from `decodedLength-colonPos+1` to `decodedLength-colonPos` - this now copies exactly the password bytes plus the null terminator already written at `authString[decodedLength]`, which fits the `decodedLength` - sized `password` buffer with no slack, even for the empty username edge case.


JoeNemo:
You are copying one fewer bytes in the memcpy.  Is that your intention?   Were we copy too many or too few bytes into password?

JoeNemo (approved):
Approved.

PR #648, opened 2026-07-22
Check (SSCTFLG1 & SSCTLDEL) before dereferencing SSCTSUSE

### LINKED ISSUE #624: List of SoftwareInstance is obtained but not used in findSessions
SoftwareInstance list is obtained at the beginning of findSessions, but till the end of the function, the list is never used.

https://github.com/zowe/zowe-common-c/blob/97ef2c262433cfc9025123f4eadc347a7ea67b33/c/discovery.c#L558-L565

I noticed this minor issue because ABEND happened in the following calling route:

[0] [visitSSCTEntry](https://github.com/zowe/zowe-common-c/blob/97ef2c262433cfc9025123f4eadc347a7ea67b33/c/discovery.c#L361)
[1] discoverSubsystems
[2] scanSlowSoftware
[3] zModelGetSoftware
[4] findSessions

Although the ABEND in `visitSSCTEntry` should be investigated and fixed, but since `findSessions` doesn't seem to need the software instance list, why not just remove the call?

ChongZhou-Broadcom:
Fix [issue 624](https://github.com/zowe/zowe-common-c/issues/624)

In a customer system, ZSS crashed when accessing SSCVT.SSCTSUSE. The root cause is not clear at the moment. The customer refused to continue the investigation.

However, we could try to skip unavailable items in the SSCVT chain by checking the SSCVT. SSCTFLG1 against mask SSCTLDEL. If the masked bit is 1, then the item should be skipped. _(suggested by @JoeNemo)_

[IBM's document](https://www.ibm.com/docs/en/zos/3.1.0?topic=xtl-sscvt-information) explains SSCTLDEL as below:

> "X'10'" Subsystem is logically deleted. When this flag is set, subsystem is removed from the SSCVT chain

In addition to checking the flag field, I also noticed in my test that, if I temporarily delete a subsystem by the command `SETSSI DELETE,SUBNAME=xxx,FORCE`, the subsystem name will become "!DEL", so the checking finally becomes:

```c
if((ssctChain->flags & SSCTFLG1_SSCTLDEL) || !memcmp(ssctChain->sname, "!DEL", 4)) {
    /* skip the current item */
}
```

However, I cannot fully reproduce the customer's symptom --- when I deleted a subsystem, its SSCTSUSE field was immediately set to 0 by the system, which will not pass the `isPointerCommon` test. So the problematic subsystem (QMR) in the customer's system must have experienced a different situation (such as hung up by XDC).

Therefore, this PR will not necessarily fix the issue in the customer's system, but it might make ZSS stronger, at lease it is no harm according to my internal tests.

The following is the instrumental log from my test: (this log is not included in the PR)

```
checking unit. sname=JES2, flags=0xa0, suse=0xcce6fc
checking unit. sname=MSTR, flags=0x0, suse=0x0
checking unit. sname=IZUG, flags=0x0, suse=0x0
checking unit. sname=RACF, flags=0x0, suse=0x0
checking unit. sname=SMS , flags=0x0, suse=0x0
checking unit. sname=TNF , flags=0x0, suse=0xce91a8
checking unit. sname=VMCF, flags=0x0, suse=0xce90c0
checking unit. sname=BLSR, flags=0x0, suse=0x0
checking unit. sname=IRLM, flags=0x0, suse=0x0
checking unit. sname=I10A, flags=0x0, suse=0x0
checking unit. sname=JRLM, flags=0x0, suse=0x0
checking unit. sname=!DEL, flags=0x0, suse=0x0
checking unit. sname=GSVX, flags=0x0, suse=0x0
checking unit. sname=RIM , flags=0x0, suse=0x0
checking unit. sname=RRS , flags=0x0, suse=0x1296bf80
checking unit. sname=AXR , flags=0x0, suse=0x0
checking unit. sname=!DEL, flags=0x0, suse=0x0
checking unit. sname=HASP, flags=0xa0, suse=0xcce6fc
checking unit. sname=!DEL, flags=0x0, suse=0x0
checking unit. sname=CICS, flags=0x0, suse=0x0
checking unit. sname=PDSM, flags=0x0, suse=0x0
```

### DIFF
--- a/c/discovery.c
+++ b/c/discovery.c
@@ -348,11 +348,16 @@ static void visitSSCTEntry(DiscoveryContext *context,
   void *usr1 = (void*)INT2PTR(ssctChain->ssctsuse);
   zowelog(NULL, LOG_COMP_DISCOVERY, ZOWE_LOG_DEBUG, "user pointer at 0x%x COMMON?=%s\n",usr1,isPointerCommon(gda,usr1) ? "YES" : "NO");
   if (isPointerCommon(gda,usr1)){
+    char *sname = &(ssctChain->sname[0]);
+    char *usrData = (char*)usr1;
+    bool deleted = ((ssctChain->flags & SSCTFLG1_SSCTLDEL) != 0);
+    if (deleted || !memcmp(sname, "!DEL", 4)) {
+      zowelog(NULL, LOG_COMP_DISCOVERY, ZOWE_LOG_DEBUG, "skipping logically deleted unit. sname=%4.4s, flags=0x%x\n",sname,ssctChain->flags);
+      return;
+    }
     if (context->ssctTraceLevel >= 1){
       dumpbuffer((char*)usr1,48);
     }
-    char *sname = &(ssctChain->sname[0]);
-    char *usrData = (char*)usr1;
     zowelog(NULL, LOG_COMP_DISCOVERY, ZOWE_LOG_DEBUG, "sname=%4.4s\n",sname);
     if (!memcmp(sname,"CICS",4) && (context->ssctTraceLevel >= 1)){
       dumpbuffer(usrData+0x08,6);
--- a/h/zos.h
+++ b/h/zos.h
@@ -1287,6 +1287,10 @@ typedef struct SSCT_tag{
   struct SSCT_tag *__ptr32 scta;  /* chain pointer */
   char sname[4];          /* 4 letter name */
   unsigned char flags;    /* equates */
+#define SSCTFLG1_SSCTSFOR 0x80
+#define SSCTFLG1_SSCTUPSS 0x40
+#define SSCTFLG1_SSCTARDR 0x20
+#define SSCTFLG1_SSCTLDEL 0x10
   unsigned char ssid;     /* JES2 or 3 */
   unsigned char reserved1[2];
   Addr31 ssvt;              /* SUBSYSTEM VECTOR TABLE POINTER */


c/discovery.c
```diff
@@ -348,11 +348,16 @@ static void visitSSCTEntry(DiscoveryContext *context,
   void *usr1 = (void*)INT2PTR(ssctChain->ssctsuse);
   zowelog(NULL, LOG_COMP_DISCOVERY, ZOWE_LOG_DEBUG, "user pointer at 0x%x COMMON?=%s\n",usr1,isPointerCommon(gda,usr1) ? "YES" : "NO");
   if (isPointerCommon(gda,usr1)){
+    char *sname = &(ssctChain->sname[0]);
+    char *usrData = (char*)usr1;
+    bool deleted = ((ssctChain->flags & SSCTFLG1_SSCTLDEL) != 0);
```
Martin-Zeithaml, line 353: Where is the `SSCTFLG1_SSCTLDEL` flag defined?

c/discovery.c
```diff
@@ -348,11 +348,16 @@ static void visitSSCTEntry(DiscoveryContext *context,
   void *usr1 = (void*)INT2PTR(ssctChain->ssctsuse);
   zowelog(NULL, LOG_COMP_DISCOVERY, ZOWE_LOG_DEBUG, "user pointer at 0x%x COMMON?=%s\n",usr1,isPointerCommon(gda,usr1) ? "YES" : "NO");
   if (isPointerCommon(gda,usr1)){
+    char *sname = &(ssctChain->sname[0]);
```
Martin-Zeithaml, line 351: PR title says "Check before SSCTSUSE", but the code is checking it after SSCTSUSE.

Should we move this check?

ChongZhou-Broadcom: Ah sorry, my bad. The title was not accurate. I changed it from "...before reading..." to "...before dereferencing...". `isPointerCommon` only check the pointer's value against some ranges.

c/discovery.c
ChongZhou-Broadcom: Good catch, but I am surprised... you did not find the definition because I coincidently omitted it in the commit, but how did it pass all the auto tests? Wired. But anyway I will add it into the PR. Thanks.

ChongZhou-Broadcom: Oh... I took a quick look at those tests. It seems they do not compile the code. Silly me, they do not have an lpar, why did I assume they compile it.

c/discovery.c
```diff
@@ -348,11 +348,16 @@ static void visitSSCTEntry(DiscoveryContext *context,
...
+    char *usrData = (char*)usr1;
+    bool deleted = ((ssctChain->flags & SSCTFLG1_SSCTLDEL) != 0);
+    if (deleted || !memcmp(sname, "!DEL", 4)) {
```
JoeNemo, line 354: Where is the "!DEL" documented as an "interesting" subsystem name.   I only saw the SSCTLDEL flag in my research.

ChongZhou-Broadcom: No, it's undocumented. I saw it in my test.

If you could look at the "checking unit..." logs again in the description, you will found three of them. When I deleted a subsystem by `SETSSI DELETE,SUBNAME=xxx,FORCE`, the name became "IDEL".

I am not sure if it would be better to also check this special name or not. What would you say? Thanks.

Martin-Zeithaml: _The subsystem name. The name can be up to 4 characters long; it must begin with an alphabetic or national character (#, @, or $), and the remaining characters (if any) can be alphanumeric or national._ ([ibm doc](https://www.ibm.com/docs/en/zos/3.2.0?topic=form-statements-parameters-iefssnxx))

If we agree that delete flag is right and subsystem name syntax is right, the extra condition `!memcmp(sname, "!DEL", 4)` should not affect anyone, but it will work at least on one Broadcom machine.

JoeNemo: Can a subsystem legitimately have the name "!DEL"?  I don't know the answer yet, but that could be an issue?  And also if this SETSSI command does not set the logical deletion flag in SSCT entry what does?   Further do you know that the SUBNAME=xxx was in the particular test you did was?

JoeNemo: Suppose a customer legitimately names their subsystem "!DEL"?   It could affect such a system.  I am googling/chatgpting/clauding as hard as I can to find evidence of what the subsystem name becomes and I see no evidence of the "!DEL", not to say that it might be sitting there in the MVS nucleus.   Also, this SETSSI DELETE test implies that the logical deletion bit did NOT get set when this command is run, which is a second mystery.  Still investigating...

JoeNemo: Eureka!  I found it in https://www.ibm.com/docs/en/zos/2.2.0?topic=subsystem-naming-your:

The name you use for your subsystems depends on how your subsystem is defined to MVS™. Use one of the following naming conventions:

    If your subsystem is defined to MVS through the IEFSSNxx parmlib member processing at IPL, the subsystem name can be no more than four characters long, beginning with an alphabetic character or #, @, or $. The remaining characters can be alphabetic, numeric, or #, @, or $.
    If your subsystem is defined to MVS through the IEFSSI macro, the subsystem name can be no more than four characters long, containing any character other than blanks or nulls.
    If you subsystem is defined to MVS through the SETSSI command, the subsystem name can contain any character other than blanks or nulls that is valid for system commands. See [z/OS MVS System Commands](https://www.ibm.com/docs/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.ieag100/toc.htm) for more information on the valid characters.

You cannot use the following names for your subsystems:

    APPC
    ASCH
    MSTR
    OMVS
    STC
    SYS
    TSO
    !DEL
    !DMY

JoeNemo (approved):
Approved.

PR #654, opened 2026-07-28
Consistently use USS_MAX_PATH_LENGTH instead of PATH_PAX

ChongZhou-Broadcom:
Due to the following reasons:

1. We already have `USS_MAX_PATH_LENGTH` defined in `unixfile.h` and this header file is the common dependency of all USS file related modules (such as httpfileservice.c, zosfile.c, and etc.);
https://github.com/zowe/zowe-common-c/blob/ddfef95eedf2d2980a0e8f0834f58dd7bf1a20fd/h/unixfile.h#L494

2. The value 1023 is a better choice than 256 in today's z/OS.

We should consistently use USS_MAX_PATH_LENGTH instead of PATH_PAX.

### DIFF
--- a/c/httpfileservice.c
+++ b/c/httpfileservice.c
@@ -41,14 +41,6 @@
 #define ENOENT 129   
 #endif
 
-#ifndef PATH_MAX
-# ifdef _POSIX_PATH_MAX
-#   define PATH_MAX  _POSIX_PATH_MAX
-# else
-#   define PATH_MAX  256
-# endif
-#endif
-
 #define MAX_BASE64_CONTENT_LENGTH (256 * 1024 * 1024)
 
 #ifdef __ZOWE_OS_ZOS
@@ -770,7 +762,7 @@ void respondWithUnixFileMetadata(HttpResponse *response, char *absolutePath) {
 
 void directoryChangeOwnerAndRespond(HttpResponse *response, char *path,
         char *user, char *group, char *Recursive, char *pattern) {
-# define RETURN_MESSAGE_SIZE (PATH_MAX + 50)
+# define RETURN_MESSAGE_SIZE (USS_MAX_PATH_LENGTH + 50)
 
   char message[RETURN_MESSAGE_SIZE] = {0};
   UserInfo  userInfo = {0};


c/zosfile.c
```diff
@@ -1426,7 +1425,7 @@ int directoryMakeDirectoryRecursive(const char *pathName,
   const char * nextField, *tempField, *endField;
   FileInfo info = {0};
   int done = 0;
-  char Path[PATH_MAX];
+  char Path[USS_MAX_PATH_LENGTH];
```
Martin-Zeithaml: I am not sure about this...
The unix path is up to 1023 characters, but the path is compound from segments, each up to 256 characters.
```
/segment1_max256chars/segment2_max_256chars/...
```

The code here splits the path into segments, if the segment is over 256 characters, it has to report an error.
If the entire path is over 1023 characters, it has to report an error.

If you increase the segment to 1023, then you are buffer overflow safe, but you need to check the segment length.

Otherwise this code would try to create invalid directory.

ChongZhou-Broadcom: Make sense.

But I didn't "increase the segment to 1023". The char array `path` is an accumulating buffer, not for one segment, so what I enlarged is the array for the whole path not for single segment.

You pointed out 2 requirements:
- If the segment is over 256 characters, it has to report an error.
- If the entire path is over 1023 characters, it has to report an error.

For the 1st one, we can check the `endField - nextField`, but we don't have to. `BPXMKD` will return error.

For the 2nd one, in recursive mode, since the `path` will eventually hold the entire path, I think we can simply check `strlen(pathName)` against USS_MAX_PATH_LENGTH at the very beginning.

In non-recursive mode, it's hard to say which segment is the new directory. It could be first one or the last one, so in my opinion, we just check `strlen(pathName)` too.

So in conclusion, I would add the following, let me know what you think:

```c
/* 
 * Recursively, make directory tree.
 */
int directoryMakeDirectoryRecursive(const char *pathName, 
                                   char * message, int messageLength,
                                   int recursive, int forceCreate){
  int returnCode = 0, reasonCode = 0, status = 0;
  int returnValue = 0;
  const char * nextField, *tempField, *endField;
  FileInfo info = {0};
  int done = 0;
  char Path[USS_MAX_PATH_LENGTH];
  char *path = Path;
  path[0] = '\0';
  nextField = pathName;

  // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
  if (strlen(pathName) >= USS_MAX_PATH_LENGTH) {
    errno = ENAMETOOLONG;
    return -1;
  }
  // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<

  /* Determine if absolute path or relative path */
  if (0 != strncmp (nextField,"/",1)) {
  ... ...
```

ChongZhou-Broadcom: The disadvantage of partially relying on `BPXMKD` to find and report error is that some-but-not-all new directories will be created if the given path is too long.

Do you think it's necessary to check the path and segments without creating directories at the beginning?

Martin-Zeithaml:
For your awareness, #655 is updating `directoryMakeDirectoryRecursive`.
It should be good to sync the work.

ChongZhou-Broadcom:
> For your awareness, #655 is updating `directoryMakeDirectoryRecursive`. It should be good to sync the work.

Thank you so much for pointing out this. #655 fixed the potential buffer overrun issue as you pointed out earlier.

So this PR will only focus on fixing the inconsistent definitions of PATH_MAX ---> we now consistently use USS_MAX_PATH_LENGTH.

ChongZhou-Broadcom:
This PR should go after #655 due to conflicts in code changes. (I will resolve the conflicts after #655 is merged)

ChongZhou-Broadcom:
The main problem was also fixed by PR #655. What's left in this PR is not quite worth a fix. PR closed.

PR #655, opened 2026-07-28
Fix stack buffer overflow in directoryMakeDirectoryRecursive for long paths

struga0258:
## Bug
`directoryMakeDirectoryRecursive()` assembled the target path in a fixed 256-byte stack buffer (`char Path[PATH_MAX]`) using unbounded `strcat`/`strncat`. A path longer than that buffer overran the stack, causing a 0C4 protection ABEND. Via the ZSS `/unixfile/mkdir` service this let any client crash the mkdir handler by creating a directory with a path over 255 characters (zss#2094).

## Change
Reject paths that don't fit the buffer up front (accounting for a possible `./` prefix, trailing `/`, and null terminator), returning -1 with an empty message so the caller responds with a normal error instead of ABENDing.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Zowe Common C Changelog
 
 ## `3.6.0`
+- Bugfix: `directoryMakeDirectoryRecursive()` now rejects a path that does not fit its internal buffer instead of overrunning a fixed 256-byte stack buffer, which caused a 0C4 ABEND (crash) when creating a directory with a path longer than 255 characters. [(zss#2094)](https://github.com/zowe/zss/issues/2094)
```
ChongZhou-Broadcom: But is the path length really up to 255 in z/OS?

I didn't find an official document that directly/clearly states the maximum length of a path or a file name, but I found the FTP error code-550. It says:

> Note: The maximum length for a file name is 255. The total path name maximum length is 1023.

https://www.ibm.com/docs/en/zos/3.2.0?topic=codes-550-name-length-error-pathname-pathname

Even worse, if you search for "PATH_PAX" in common-c, you will find inconsistent definitions:

The following seems like a "careless whisper":
https://github.com/zowe/zowe-common-c/blob/4e27abb2472635c8109d5f5ae8e3d94233e7a304/c/zosfile.c#L1417-L1429

The following probably means "either the limits.h or the build script should define it, if not, use 256 by default":
https://github.com/zowe/zowe-common-c/blob/4e27abb2472635c8109d5f5ae8e3d94233e7a304/c/httpfileservice.c#L44-L50

The following gave 2 correct values, but it used a different name for path limit.
https://github.com/zowe/zowe-common-c/blob/4e27abb2472635c8109d5f5ae8e3d94233e7a304/h/unixfile.h#L494-L495

I created another [PR](https://github.com/zowe/zowe-common-c/pull/654/changes) yesterday in order to fix the inconsistency. If you fix it in your PR, I'll cancel mine.

ChongZhou-Broadcom (changes requested):

CHANGELOG.md
struga0258: Please check latest push, instead of a hardcoded PATH_MAX 256, the buffer is now sized to the real zos uss path maximum (USS_MAX_PATH_LENGTH = 1023, already used elsewhere in the file), with +4 bytes for the "./" prefix, trailing /, and null terminator, and the guard rejects only paths that actually exceed that limit. So it fixes the 0C4 crash without wrongly rejecting valid long paths.

ChongZhou-Broadcom (dismissed):
Now it looks good to me.

JoeNemo:
Thanks for this -- the sizing is right and it is the part that is easy to get
wrong. `Path[USS_MAX_PATH_LENGTH + 4]` with a `strlen(pathName) >
USS_MAX_PATH_LENGTH` reject exactly covers the worst case: the relative branch
prepends `"./"` without consuming a character and the loop appends a trailing
`/` after the final segment, so the assembled string reaches
`strlen(pathName) + 3` and needs one more byte for the terminator. I derived the
same `+4` independently before reading this, and only got there after testing
relative paths -- an entry check reasoned out from absolute paths alone (`+2`)
still overflows. Worth saying because the guard proposed on #654 has that bug.

I have pushed two commits here. Happy to drop or rework either.

**The buffer contract.** Raising the internal buffer moves the problem to the
caller. `strncpy(message, path, messageLength)` does not terminate when the path
is at least `messageLength` characters, and `createUnixDirectoryAndRespond`
calls `strlen(message)` on the next line. That used to be structurally
impossible -- `Path` was 256 bytes and the caller's buffer is `PATH_MAX`, 1024
on z/OS -- so the size mismatch was doing the work. With both at the USS
maximum it is reachable on a legal 1023-character path.

Guaranteeing termination is not enough on its own: a truncated path names a
directory that need not exist, and this value is reported to the user as
"Successfully created directory: ...". So instead the required size is now
published as `USS_MKDIR_PATH_BUFFER_SIZE` and a smaller `messageLength` is
rejected at entry, before anything is created, so -1 unambiguously means nothing
was done. The same constant sizes the internal buffer and the caller's
declaration, so no call site repeats the arithmetic. `message == NULL` stays
legal and the copy is now guarded for it.

The only caller anywhere in the org is `createUnixDirectoryAndRespond`, reached
from zss `serveUnixFileMakeDirectory`, so the blast radius is small -- but this
does change behaviour on a function in a public header, hence the CHANGELOG
line.

**Tests.** `tests/zosfile-mkdir/` has four checks and a `run-zos.sh` following
the convention in `tests/charset-streaming/`. Two guard your fix (over-long and
NULL paths rejected); two guard the contract (an undersized buffer is rejected;
a correctly sized one receives a path that is complete and actually exists on
disk). It walks only existing directories and passes paths that are rejected
before any BPXMKD, so it creates nothing and writes nothing, and every case
prints its arguments and results so a reader can check the verdict.

Run on z/OS (Marist, ibm-clang64), before and after the contract commit:

```
[3] a messageLength below the required minimum is rejected
    messageLength : 16  (minimum is 1027)
    returned      : 0 (expected -1)
    buffer written: YES -- a partial path was written
    VERDICT       : FAIL
1 failure(s)
```

and with the fix:

```
[3] ... returned : -1 (expected -1)   buffer written: no, still all 'X'   PASS
[4] reported path : "/ZOWE/joezowe/scratch202607/pr655" (33 chars)
    equals input  : yes    exists on disk: yes                            PASS
0 failure(s)
```

**One heads-up for #654:** this touches the same lines of
`c/httpfileservice.c` that #654 converts from `PATH_MAX`, so they will
conflict. Whichever lands second should take `USS_MKDIR_PATH_BUFFER_SIZE` for
this buffer; the `PATH_MAX` uses in `directoryChangeOwnerAndRespond` are
untouched here and still belong to #654.

JoeNemo:
Triaged the SonarCloud findings. One was worth acting on; I have pushed a fix
for it (c8cb38c). Recording the rest so nobody re-litigates them later.

**Real: `strncat` bound (c:S5815 hotspot, and c:S3519 BLOCKER) at
httpfileservice.c:158,165.** `strncat`'s third argument is the maximum number
of characters to *append*, not the size of the destination, and we were passing
the destination size. Now bounded by the space actually remaining.

The BLOCKER's claim itself is false, though. "String concatenation function
overflows the destination buffer" -- it does not: the longest prefix (37) plus
a maximal path (1026) plus the terminator is 1064 bytes into a 1077-byte
buffer, 13 bytes of headroom, confirmed under AddressSanitizer with the two
buffers replicated exactly. Sonar is reasoning from the third argument alone
without correlating the size of `message`, so it reports the worst case the
argument *permits* as though it were reachable. Right to complain, wrong about
the consequence -- and it is worth fixing anyway, because code that is safe
only by arithmetic no reader checks is one edit away from not being safe.

**Not real, no action:**

- `c:S1874` "'strncat' is deprecated ... consider strncat_s" -- that is the
  Microsoft/Annex K deprecation. `strncat_s` is not available here.
- `c:S5813` "Make sure use of strlen is safe" x5. These are hotspots, i.e. a
  prompt to look, not a defect claim. Having looked: in the test the arguments
  are a local array this file just terminated, a `getcwd` result checked for
  failure, and -- at line 203 -- a buffer whose terminator the preceding line
  explicitly verifies. The one in zosfile.c is the guard's own
  `strlen(pathName)`, short-circuited behind a NULL check; bounding it would
  want `strnlen`, which needs `_POSIX_C_SOURCE >= 200809L` and this build sets
  `_XOPEN_SOURCE=600`, so it is not a free change.
- `c:S5955` "declare i inside the loop" x3 and `c:S1659` "one identifier per
  statement" -- both cut against the prevailing style in `c/zosfile.c` itself,
  and consistency with the file wins.

Worth noting for anyone touching `c/httpfileservice.c`: no build in this repo
compiles it -- `h/httpserver.h` pulls in `jwt/jwt/jwt.h`, which needs
`rscrypto/`, which lives outside this repository. So its CI signal here is
Sonar and nothing else. I verified the change's behaviour by replicating both
buffers under ASan rather than by building; a real compile needs a ZSS build.

JoeNemo (dismissed):
Approved.

JoeNemo (approved):
Approved.

PR #659, opened 2026-08-03
Minor fix about using SHLAlloc in processHttpFragment

ChongZhou-Broadcom:
This is a tiny patch about using `SLHAlloc2` in `processHttpFragment()`.

In PR #603, `SLHAlloc2` was implemented to allow a caller to choose if it should ABEND or return NULL when the heap is out of space. The function `processHttpFragment()` was also changed to only use `SLHAlloc2` in "return NULL" mode because it is not protected by a recovery routine.

However, I seemed to make a mistake in PR #640. Once of `SLHAlloc2` was coincidentally changed back to `SLHAlloc`. This tiny PR corrects the mistake. I apologize for the inconvenience!

Additionally, I also defined 2 macros: `SLHALLOC2_NO_ABEND` and `SLHALLOC2_MAY_ABEND` for the third argument of `SLHAlloc2` in order for a better readability.

### DIFF
--- a/c/httpserver.c
+++ b/c/httpserver.c
@@ -2348,7 +2348,7 @@ int processHttpFragment(HttpRequestParser *parser, char *data, int len){
         } else{
           zowelog(NULL, LOG_COMP_HTTPSERVER, ZOWE_LOG_DEBUG3, "_____ END OF MESSAGE HEADER _________\n");
           parser->state = HTTP_STATE_READING_FIXED_BODY;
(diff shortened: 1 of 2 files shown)


ChongZhou-Broadcom:
@JoeNemo Since this is a very tiny fix, I didn't update the change log. Give me a shout if you want me to update the change log for this PR. Again, I am sorry for giving you this extra work!

JoeNemo (approved):
Approved.

PR #666, opened 2026-08-14
Implement fileGetINode/fileGetDeviceID off z/OS

JoeNemo:
`h/unixfile.h` declares `fileGetINode()` and `fileGetDeviceID()` unconditionally,
and `c/httpserver.c:4263` calls `fileGetINode()` from `makeFileEtag()`. But only
`c/zosfile.c` ever implemented the pair. `platform/windows/winfile.c` implements
one of the two; `platform/posix/psxfile.c` implements neither. So any off-z/OS
build that reaches a file's identity fails to link on a function the header
offers it.

This implements both over `struct stat` (`st_ino` / `st_dev`), beside the rest
of the accessor family already in `psxfile.c` -- `fileInfoSize`,
`fileInfoIsDirectory`, `fileInfoOwnerUID` and friends -- and adds the missing
`fileGetDeviceID()` to the Windows port so the declared pair exists everywhere
it is declared. **The z/OS implementations are not touched.**

### Why now

ZSS keys its upload-session table on the `(inode, deviceID)` pair, in
`c/unixFileService.c`, and reaches the fields directly rather than through these
accessors:

```
c/unixFileService.c:281: no member named 'inode' in 'struct stat'
c/unixFileService.c:282: no member named 'deviceID' in 'struct stat'
c/unixFileService.c:475: no member named 'inode' in 'struct stat'
c/unixFileService.c:476: no member named 'deviceID' in 'struct stat'
```

Those four lines are the entire reason the `/unixfile` service cannot be
compiled anywhere but z/OS, which in turn is why it has no off-platform test
coverage. A companion ZSS change converts them to these accessors. That is worth
saying plainly: this is not a new abstraction invented for portability, it is
the abstraction the header already documents at line 429 ("FileInfo is a
generic, opaque typedef") being finished.

### On the return type

`int`, because BPXYSTAT's `inode` field is an `int`. On hosts where `ino_t` and
`dev_t` are wider, these narrow, and the implementations say so. Callers use the
values as a hash input (`makeFileEtag`) or as one half of an identity pair
(ZSS's `FileID`), never as something handed back to the filesystem, so the
narrowing costs uniqueness rather than correctness. Widening the signature would
be an API change across three platforms and every caller; it did not seem worth
bundling into this.

### Tests

`tests/fileinfo-identity/` checks the contract both platforms have to honour:

1. a real file has a non-zero inode and device
2. distinct files differ in inode, and agree on device within one filesystem
3. a hard link matches its target in **both** halves of the pair
4. the pair is stable across repeated `fileInfo()` calls

`run.sh` builds and runs it on Linux/WSL under AddressSanitizer and
UndefinedBehaviorSanitizer:

```
  path           inode        deviceID
  fiid-a.tmp     532318       2096
  fiid-b.tmp     532319       2096
  fiid-link.tmp  532318       2096

  case 1: a real file has a non-zero identity            ok, ok
  case 2: distinct files are distinguishable             ok, ok
  case 3: a hard link is the same file                   ok, ok
  case 4: the identity is stable                         ok

7 checks, 0 failed
```

The negative control matters more than the pass here. Linking the same test
against `v3.x/staging`'s `psxfile.c` does not produce a failing assertion -- it
does not link at all:

```
fileinfo-identity-test.c:102: undefined reference to `fileGetINode'
fileinfo-identity-test.c:106: undefined reference to `fileGetDeviceID'
```

`run-zos.sh` runs the same cases under `ibm-clang64` against the z/OS
implementations, using the compiler flags lifted verbatim from
`build/build_cmgr_clang.sh`, as `tests/zosfile-mkdir/run-zos.sh` does. It also
compiles `psxfile.c` on z/OS, which is what proves the file's platform guard
still excludes these two rather than colliding with the `zosfile.c` versions.
**Not yet run on Marist** -- I will post the output before this merges.


ifakhrutdinov (approved):

JoeNemo:
Closing the loop on this: the PR body said `tests/fileinfo-identity/run-zos.sh`
had not been run on z/OS and that I would post the output. It merged before I
did, so here it is, run against `v3.x/staging` (a fresh clone, so it includes
the merge) with `ibm-clang64` on Marist.

All 15 translation units compile, including `psxfile.c`, which is the check that
its platform guard still excludes the two new functions on z/OS rather than
colliding with the `zosfile.c` versions:

```
== compiling with ibm-clang64 (flags lifted from build_cmgr_clang.sh zos) ==
  OK   zosfile.c ... OK   psxfile.c ... OK   fileinfo-identity-test.c

ALL COMPILED under ibm-clang64.

== linking + running ==
fileGetINode / fileGetDeviceID over FileInfo

  path           inode        deviceID
  fiid-a.tmp     56829        47
  fiid-b.tmp     56830        47
  fiid-link.tmp  56829        47

  case 1: a real file has a non-zero identity
    ok    inode of fiid-a.tmp is non-zero
    ok    deviceID of fiid-a.tmp is non-zero
  case 2: distinct files are distinguishable
    ok    fiid-a.tmp and fiid-b.tmp have different inodes
    ok    same-filesystem files share a deviceID
  case 3: a hard link is the same file
    ok    hard link has the inode of its target
    ok    hard link has the deviceID of its target
  case 4: the identity is stable
    ok    a second fileInfo() reports the same inode

7 checks, 0 failed
(exit 0)
```

Same 7/7 as the Linux run in the PR body, over BPXYSTAT instead of `struct stat`
and in EBCDIC char mode with the `__LONGNAME__` mappings active. The values
differ as you would expect for a different filesystem -- inode 56829 and device
47 here against 532318 and 2096 on the Linux run -- and the relationships the
test actually asserts hold identically on both.

That makes the contract in `tests/fileinfo-identity/` verified on both sides of
the platform split rather than on one.

PR #667, opened 2026-08-18
logging: test the loop bound before indexing the component ID

JoeNemo:
`logConfigureComponent()` and `getComponent()` walk a component ID as four
unsigned shorts aliased over a `uint64`:

```c
unsigned short *id = (unsigned short *)&compID;
for (int i = 1; id[i] != 0 && i < 4; i++) {
```

`id[i]` is evaluated before `i < 4` is tested, so reaching `i == 4` reads two
bytes past `compID`. This swaps the operands so the bound is tested first. Four
identical loops, four identical edits; nothing else about them changes.

### Why it has gone unnoticed

It depends on byte order. Every component ID in `logging.h` ends in a zero low
short -- I checked all 23 -- so on a **big-endian** host `id[3]` is `0` and the
walk stops at `i == 3` before the bound is ever the thing that ends it. It is
latent there, waiting for someone to define a four-level component ID.

On a **little-endian** host the vendor ID lands in `id[3]` instead, which is
never zero, so every call runs off the end. Taking
`LOG_COMP_HTTPSERVER = 0x008F000100090000`:

| | `id[0]` | `id[1]` | `id[2]` | `id[3]` | old loop stops at |
|---|---|---|---|---|---|
| big-endian | `008F` | `0001` | `0009` | `0000` | `i=3`, on `id[3]==0` |
| little-endian | `0000` | `0009` | `0001` | `008F` | `i=4`, after reading `id[4]` |

AddressSanitizer on the real macro:

```
ERROR: AddressSanitizer: stack-buffer-overflow
READ of size 2 at 0x74c953a00028
  [32, 40) 'compID' <== Memory access at offset 40 overflows this variable
```

### Scope

No test is added. The condition is unreachable on z/OS with the IDs we ship, and
reproducing it elsewhere needs a sanitizer to observe a two-byte overread that
is otherwise silent -- so a test would assert on the tool rather than on the
code. The ASan output above is the evidence.

The endian-dependence visible in that table is a **separate** defect and is not
addressed here: on a little-endian host the walk reads the shorts in reverse, so
the component hierarchy lookup is wrong regardless of the bound. `(compID >> 48)`
on the line above already does it the portable way. Filed as #668 rather than folded in,
to keep this reviewable at a glance.


ifakhrutdinov (approved):
lgtm

PR #670, opened 2026-08-21
On z/OS evaluate the embedded JS always in KEY 8 and PROBLEM STATE

Martin-Zeithaml:
## Change - z/OS only
```c
Json *evaluateJsonTemplates(EmbeddedJS *ejs, ShortLivedHeap *slh, Json *json){
  TemplateEvalPrivilege savedPrivilege = dropToUnauthorizedState();
  Json *result = evaluateJsonTemplatesUnprotected(ejs, slh, json);
  restoreAuthorizedState(savedPrivilege);
  return result;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Zowe Common C Changelog
 
 ## `3.6.0`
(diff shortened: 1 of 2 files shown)


JoeNemo:
I would rather throw an error.

Martin-Zeithaml:
Will be replaced by new PR.

PR #674, opened 2026-08-26
Throw an error for embedded JS on z/OS if key != 8 or supervisor state

Martin-Zeithaml:
## Change - z/OS only

* New function `isKey8ProblemState`
* If the caller is not in key 8 or not in problem state, do not continue with processing of embedded JS.

AI-Assisted: Claude Sonnet 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Zowe Common C Changelog
 
 ## `3.6.0`
+- Bugfix: Embedded JS is expected to run in key 8 and problem state. [(#674)](https://github.com/zowe/zowe-common-c/pull/674)
 - Bugfix: `icsfDigestInit` in `icsf.c` contained a typo in the SHA-256 ICSF rule array keyword (`"SHA246"` instead of `"SHA256"`), causing SHA-256 hashing to fail. [(#594)](https://github.com/zowe/zowe-common-c/issues/594)
 - Bugfix: `ICSFDigest.hash` buffer in `icsf.h` was 32 bytes, too small for SHA-384 (48 bytes) and SHA-512 (64 bytes). Increased to 64 bytes to prevent buffer overflow. [(#594)](https://github.com/zowe/zowe-common-c/issues/594)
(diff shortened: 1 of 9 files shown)


JoeNemo:
The precondition itself looks right to me. The key extraction is consistent with
the constant already in the file:

```c
uint32 currentKey = (currentPSW >> 20) & 0x0000000F;
bool isProblemState = (currentPSW & PROBLEM_STATE) ? true : false;
```

`PROBLEM_STATE` is `0x00010000`, i.e. PSW bit 15 in the first word, so the same
word has the key at bits 8-11 -- which is `>> 20 & 0xF`. The two agree, so the
shift is not a guess.

**But `makeConfigManager()` can now return NULL, and two callers do not check
it.** The PR updates `simpleMain()` and `REXXCMGR()`, and misses these:

1. **`c/configmgr.c:1496`**, in this repo:

```c
static void *makeConfigManagerWrapper(void *userData, EJSNativeInvocation *invocation){
  return makeConfigManager();
}
```

   The NULL goes straight out to the embedded-JS layer as a native pointer.

2. **`zss/c/zss.c:1718`**, the other repo:

```c
ConfigManager *configmgr = makeConfigManager();
CFGConfig *theConfig = addConfig(configmgr,ZSS_CFGNAME);
```

   `cfgAddConfig()` calls `getConfig(mgr,...)`, whose first statement is
   `CFGConfig *config = mgr->firstConfig;`.

The second one is the one I would not merge without. It is cross-repo, so it
will not show up in this repo's CI -- ZSS picks it up on the next submodule
bump, and by then it is someone else's failure to diagnose.

**And on z/OS the failure mode is the bad kind.** Address 0 is readable there,
so `mgr->firstConfig` does not ABEND -- it reads whatever is in the PSA and
walks on with a garbage pointer. Instead of "ZSS died at startup with a clean
0C4" you get "ZSS started and then behaved strangely." That is a harder problem
to hand a customer than the state this PR is trying to prevent.

Worth deciding explicitly: is `wtoMessage` plus a NULL return the intended
contract for a wrong-state caller, or should this be fatal at the point of
detection? If the answer is that no correct caller can ever be in the wrong
state, then failing loudly and immediately is easier to support than a NULL that
every caller must now remember to test.

I have not run this on z/OS -- the PSW read needs the real thing, and I have
only reasoned about the arithmetic above. Happy to run it on Marist if that
would help.

JoeNemo:
Follow-ups filed for the unchecked callers, so they are tracked rather than
blocking this: zowe/zowe-common-c#676 (`makeConfigManagerWrapper`), zowe/zss#868
and zowe/launcher#219.

Worth noting this PR does not regress any of them. Before it,
`makeConfigManager()` discarded the result of `configureEmbeddedJS()`, so a
failed JS-context allocation stored the pointer the callee had just freed and
returned a `ConfigManager` with a dangling `ejs`. A NULL is strictly better, and
checking that return value is a robustness fix in its own right.

JoeNemo (approved):
Approved.

PR #679, opened 2026-09-03
`getCAA` - all offsets in the same format

Martin-Zeithaml:
Current implementation is mixing decimal and hexadecimal offsets.


JoeNemo (changes requested):
Can you introduce a couple of defines in "zos.h" and include that file where needed when:

`#ifdef __ZOWE_OS_ZOS`

Martin-Zeithaml:
# `DSAB_tag` comparison: `h/zos.h` vs. removed copy in `c/pdsutil.c`

Context: `c/pdsutil.c` defined its own local `DSAB_tag`/`DSAB` typedef (pre-existing duplicate of the one in `h/zos.h`). It was removed and `pdsutil.c` now includes `zos.h` (guarded by `#ifdef __ZOWE_OS_ZOS`) and uses the shared definition instead. 

## Packing

| | Packing pragma | Effect |
|---|---|---|
| `h/zos.h` `DSAB_tag` | Defined inside `ZOWE_PRAGMA_PACK` / `ZOWE_PRAGMA_PACK_RESET` (`#pragma pack(1)` / `pack(packed)` depending on compiler) | Byte-packed, no padding anywhere |
| `c/pdsutil.c` `DSAB_tag` (removed) | Defined *before* the file's own `ZOWE_PRAGMA_PACK` (that pragma appeared later in the file, around the original line 367) | Natural/default alignment |

Even though the removed `pdsutil.c` copy was *not* under a pack pragma, every field in the struct happens to fall on a naturally aligned boundary already (see the running offset column below), so default alignment inserts **zero** padding bytes. The two layouts come out identical regardless of packing mode.

## `Addr31` vs `int`

`h/zos.h` types the four DSECT pointer fields (`dsabfchn`, `dsabbchn`, `dsabtiot`, `dsabsiot`) as `Addr31`. `Addr31` is defined in `h/zowetypes.h`:

```c
#if defined(__ZOWE_ARCH_Z) && !defined(__ZOWE_OS_LINUX)
typedef void *__ptr32 Addr31;   /* z/OS: a real 31-bit pointer, 4 bytes */
#else
typedef uint32_t Addr31;        /* everywhere else: 4 bytes */
#endif
```

The removed `pdsutil.c` copy typed those same four fields as plain `int` (4 bytes). Both are 4 bytes wide, so this is a naming/typing improvement only — it does not change field size, offset, or the struct's total size. Existing call sites in `pdsutil.c` already treated these fields as raw addresses via explicit casts (e.g. `(TIOT*)dsab->dsabtiot`, `((int*)dsab->dsabtiot)[3]`), which compile identically against `Addr31` or `int`.

## Field-by-field layout

| Field | Offset (hex) | Offset (dec) | Size (bytes) | `zos.h` type | `pdsutil.c` type (removed) | Same layout? |
|---|---|---|---|---|---|---|
| `eyecatcher` | 0x00 | 0 | 4 | `char[4]` | `char[4]` | Yes |
| `dsabfchn` | 0x04 | 4 | 4 | `Addr31` | `int` | Yes (both 4 bytes) |
| `dsabbchn` | 0x08 | 8 | 4 | `Addr31` | `int` | Yes (both 4 bytes) |
| `dsablnth` | 0x0C | 12 | 2 | `short` | `short` | Yes |
| `dsabopct` | 0x0E | 14 | 2 | `short` | `short` | Yes |
| `dsabtiot` | 0x10 | 16 | 4 | `Addr31` | `int` | Yes (both 4 bytes) |
| `dsabssva` | 0x14 | 20 | 4 | `int` | `int` | Yes |
| `dsabxsva` | 0x18 | 24 | 4 | `int` | `int` | Yes |
| `dsabanmp` | 0x1C | 28 | 4 | `int` | `int` | Yes |
| `dsaborg1` | 0x20 | 32 | 1 | `char` | `char` | Yes |
| `dsaborg2` | 0x21 | 33 | 1 | `char` | `char` | Yes |
| `dsabflg1` | 0x22 | 34 | 1 | `char` | `char` | Yes |
| `dsabflg2` | 0x23 | 35 | 1 | `char` | `char` | Yes |
| `dsabflg3` | 0x24 | 36 | 1 | `char` | `char` | Yes |
| `dsabflg4` | 0x25 | 37 | 1 | `char` | `char` | Yes |
| `dsabdext` | 0x26 | 38 | 2 | `short` | `short` | Yes |
| `dsabtcbp` | 0x28 | 40 | 4 | `int` | `int` | Yes |
| `dsabpttr` | 0x2C | 44 | 4 | `int` | `int` | Yes |
| `dsabssnm` | 0x30 | 48 | 4 | `char[4]` | `char[4]` | Yes |
| `dsabsscm` | 0x34 | 52 | 4 | `int` | `int` | Yes |
| `dsabdcbm` | 0x38 | 56 | 6 | `char[6]` | `char[6]` | Yes |
| `dsabtctl` | 0x3E | 62 | 2 | `short` | `short` | Yes |
| `dsabsiot` | 0x40 | 64 | 4 | `Addr31` | `int` | Yes (both 4 bytes) |
| `dsabtokn` | 0x44 | 68 | 4 | `int` | `int` | Yes |
| `dsabtct2` | 0x48 | 72 | 4 | `int` | `int` | Yes |
| `dsabsiox` | 0x4C | 76 | 4 | `int` | `int` | Yes |
| `dsabfcha` | 0x50 | 80 | 4 | `int` | `int` | Yes |
| `dsabbcha` | 0x54 | 84 | 4 | `int` | `int` | Yes |
| `dsabflg5` | 0x58 | 88 | 1 | `char` | `char` | Yes |
| `reserved` | 0x59 | 89 | 7 | `char[7]` | `char[7]` | Yes |
| **Total size** | | | **0x60 (96 bytes)** | | | **Identical** |

No gaps: every multi-byte field's offset is already a multiple of its own size at the point it appears (e.g. `dsabdext` at offset 38 is 2-aligned; `dsabtcbp` at offset 40 is 4-aligned), so natural alignment produces the same byte layout as the explicit `pack(1)`. The struct's total size (96 bytes) is also already a multiple of 4, so there's no compiler-added trailing pad
either.

## Conclusion

The two `DSAB_tag` definitions are byte-for-byte identical: same field order, same offsets, same sizes, same total size (96 bytes / 0x60). The only difference is that `zos.h` uses the semantically-typed `Addr31` (a 4-byte 31-bit pointer type) for four fields where `pdsutil.c`'s copy used plain `int` (also 4 bytes) — a documentation-level improvement with no ABI impact.
Deleting the `pdsutil.c` copy in favor of the shared `zos.h` definition is safe.

JoeNemo (approved):
Approved.

PR #680, opened 2026-09-03
Bugfix: configmgr -script reports uncaught script errors and exits non-zero (#585)

### LINKED ISSUE #585: No errors in z/OS 3.2
It sounds like a wonderful world without errors...

This is z/OS 3.1:

```
/zowe/bin/utils: >cat ./test.js
asdf();

/zowe/bin/utils: >./configmgr -script ./test.js
ReferenceError: 'asdf' is not defined
    at <anonymous> (./test.js:1)
```

And z/OS 3.2:
```
/zowe/bin/utils: >cat ./test.js
asdf();

/zowe/bin/utils: >./configmgr -script ./test.js
```

Any idea? Some new z/OS 3.2 setting? 

`shell`, `bash`, `2>&1` no difference.

### LINKED ISSUE #3639: Failing QuickJS script is returning 0
`ok.js` is very simple: `console.log("Hello!");`
`wrong.js:` is also simple, but does not work: `aasdf;`

Let's call both scripts... 

```
./utils/configmgr -script ./ok.js
echo $?
./utils/configmgr -script ./wrong.js
echo $?
```
...and print return value:
```
Hello!
0
ReferenceError: 'aasdf' is not defined
    at <anonymous> (./wrong.js:1)

0
```

The node is returning in such cases `0|1` (MS Batch, Ubuntu, Debian) or `True|False` (MS Powershell).

JoeNemo:
Fixes #585. Fixes zowe/zowe-install-packaging#3639.

## Proposed changes

`configmgr -script` swallows a script that fails. On current staging a script that throws at top level, or that references an undefined name, prints nothing and exits 0. #585 reported "no errors" on a z/OS 3.2 system and zowe-install-packaging#3639 reported exit code 0 for a failing script; they are the same defect, and it is not about z/OS 3.2.

**Cause.** Since the QuickJS 2024-01-13 update (#573), evaluating a module returns a promise. A throw at module top level rejects that promise instead of coming back as `JS_EXCEPTION`. `ejsEvalBuffer1()` only checked `JS_IsException()` on the value `JS_EvalFunction()` returned, so the rejection was never seen, never printed, and `configmgr` fell off the end of `main()` with status 0. `-script` always evaluates as a module (`EJS_LOAD_IS_MODULE`), so every failing script took this path. Upstream's `qjs` handles the same situation with `js_std_await()`.

**Fix.**
- `c/embeddedjs.c`: after `JS_EvalFunction()`, settle the promise the way `js_std_await()` does: run pending jobs (and the event loop once if nothing is runnable) until it is fulfilled or rejected, then return the result, or re-throw the rejection so the existing exception path prints it and reports `-1`.
- `c/configmgr.c`: the `-script` branch returns `ZCFG_EVAL_FAILURE` (2) when `ejsEvalFile()` reports a failure, `ZCFG_SUCCESS` otherwise, instead of falling off the end of `main()`.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/585 and https://github.com/zowe/zowe-install-packaging/issues/3639

## Type of change

## PR Checklist

## Testing

New `tests/configmgr/cli/test_script_exit_status.sh` (runs with the existing `run_all.sh` harness on Linux and z/OS USS):

| script | before (staging) | after |
|---|---|---|
| `throw new Error("boom")` | silent, exit 0 | `Error: boom` + stack, exit 2 |
| module with a `ReferenceError` | silent, exit 0 | `ReferenceError: 'noSuchFunction' is not defined`, exit 2 |
| missing script file | message, exit 0 | message, exit 2 |
| healthy script (`configmgr_api.js`, 37 checks) | exit 0 | exit 0 |

The whole `tests/configmgr/cli` suite passes with the change (7 suites, 0 failures) on the Linux build from `build/build_cmgr_clang.sh linux`.

## Further comments

- A script that ends with `std.exit(n)` is unaffected: `std.exit` terminates the process before this code runs.
- A module that is still pending after the job queue and event loop are drained is left alone, as before; only settled promises change behaviour.
- Exit status 2 reuses `ZCFG_EVAL_FAILURE`, which `cfgLoadConfiguration()` already returns for template-evaluation failures.


Martin-Zeithaml:
Tiny quick test works as expected:
```
/git/repos/configmgr/exit.with.error.zcc.680: echo "asdf();" > test680.js
/git/repos/configmgr/exit.with.error.zcc.680: ./configmgr -script ./test680.js
ReferenceError: 'asdf' is not defined
    at <anonymous> (./test680.js:1)

 /git/repos/configmgr/exit.with.error.zcc.680: echo $?
2
```

Martin-Zeithaml (approved):

Issue #571, opened 2026-03-03
configmgr cfgSetConfigPath() does not set, it appends


1000TurquoisePogs:
https://github.com/zowe/zowe-common-c/blob/264e7681def8cfba1f6f9a214b60a2cbbc7511de/c/configmgr.c#L656   calls functions that lead to makePathElement()  which does

```
  if (config->configPath == NULL){
    config->configPath = element;
  } else {
    ConfigPathElement *tail = config->configPath;
    while (tail->next){
      tail = tail->next;
    }
    tail->next = element;
  }
```

So, if you call `cfgSetConfigPath()` twice, you will get bad output because arrays will merge and produce duplicates.

JoeNemo:
@1000TurquoisePogs .  Is the configName the same for both calls, and the pathArg different, or are they both the same for both calls.

PR #682, opened 2026-09-03
Bugfix: cfgSetConfigPath() replaces the configuration path instead of appending to it (#571)

JoeNemo:
## Proposed changes

`cfgSetConfigPath()` appended to the configuration path instead of setting it. `buildConfigPath()` adds each element through `makePathElement()`, which walks to the tail of the existing list and links the new element there, so a second `cfgSetConfigPath()` call on the same config produced a path holding both the old and the new sources. On load they were merged, and arrays in particular ended up with every member duplicated.

The fix drops the existing list head before building the new path. The elements live in the manager's short-lived heap, so nothing needs freeing individually; they go with the manager.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/571

## Type of change

## PR Checklist

## Testing

New `tests/configmgr/cli/test_config_path_reset.sh` drives the JS API (`fixtures/config_path_reset.js`): one ConfigManager, `setConfigPath` to file A (`a: 1, list: [one]`), then `setConfigPath` to file B (`b: 2, list: [two]`), load, and inspect `getConfigData()`.

| check | before | after |
|---|---|---|
| key only in the replaced file (`a`) is absent | FAIL (`a` present) | PASS |
| key in the current file (`b`) is present | PASS | PASS |
| `list` has exactly the current file's member | FAIL (`[one, two]`) | PASS |

The whole `tests/configmgr/cli` suite passes with the change on the Linux build from `build/build_cmgr_clang.sh linux`.

## Further comments

Callers that relied on the append behaviour to accumulate a path in several calls would notice; none exist in zowe-common-c, ZSS or the launcher, which each call it once with the complete path.


c/configmgr.c
```diff
@@ -674,6 +674,12 @@ int cfgSetConfigPath(ConfigManager *mgr, const char *configName, char *configPat
     return ZCFG_UNKNOWN_CONFIG_NAME;
   }
 
+  /* Set, not append: a second call replaces the path rather than merging the
+     two lists and duplicating every array member on load (#571). The
+     elements live in the manager's short-lived heap, so dropping the list
+     head is all that is needed. */
+  config->configPath = NULL;
```
ifakhrutdinov, line 678: Could this be just `// drop old config path`? I don't understand why such a long comment is needed here. I know this is LLM, but maybe we need to trim comments or tell LLMs not to generate any comments and write them ourselves.

JoeNemo: Re-writing comment by hand.

ifakhrutdinov (approved):

Issue #563, opened 2025-11-12
Schema validation error: type of the enum


Martin-Zeithaml:
* unequal constant value at /zowe/setup/certificate/type; expecting value 'PKCS12' of type **'integer'**
* no matching enum value at /zowe/setup/certificate/type; expecting one of values '[JCEKS, JCECCAKS, JCERACFKS, JCECCARACFKS, JCEHYBRIDRACFKno matching enum value at /zowe/setup/certificate/type; expecting one of values '[]' of type **'integer'**

Schema definition seems to be fine:

```json
"type": {
  "type": "string",
  "description": "Keystore type.",
  "enum": ["JCEKS", "JCECCAKS", "JCERACFKS", "JCECCARACFKS", "JCEHYBRIDRACFKS"]
},
```

Note: formatting of the message fixed [here](https://github.com/zowe/zowe-common-c/pull/562).

```
about to validate merged yamls as
{
  "zowe": {
    "setup": {
      "certificate": {
        "type": "jcesomething",
        "keyring": {
          "name": "zowekeyring"
        }
      }
    }
  }
}
validate status = 4
Validity Exceptions:
Validity Exceptions(s) with object at
  Validity Exceptions(s) with object at /zowe
    Validity Exceptions(s) with object at /zowe/setup
      Schema at '/zowe/setup/certificate' invalid
        not oneOf schemas at '/zowe/setup/certificate' are valid, 0 are
          Validity Exceptions(s) with object at /zowe/setup/certificate
            missing required property 'pkcs12' at '/zowe/setup/certificate'
            unequal constant value at /zowe/setup/certificate/type; expecting value 'PKCS12' of type 'integer'
          Validity Exceptions(s) with object at /zowe/setup/certificate
            no matching enum value at /zowe/setup/certificate/type; expecting one of values '[JCEKS, JCECCAKS, JCERACFKS, JCECCARACFKS, JCEHYBRIDRACFKno matching enum value at /zowe/setup/certificate/type; expecting one of values '[]' of type 'integer'
```

ryoung1us:
This is rated "low", but I can not install ZOWE with JCERACFKS with this bug.   Is there a work around that I am missing.  It seems like if this is blocking installs, the severity should be higher.

Martin-Zeithaml:
This bug is only about reporting incorrect data type. That's why it is low priority. We are not aware of any bug blocking configuration of key rings.

Could you provide your configuration, which is failing?

[...]

PR #683, opened 2026-09-03
Bugfix: enum/const validation messages name the schema's type, not 'integer' (#563)

JoeNemo:
## Proposed changes

Schema validation messages for `enum` and `const` mismatches named the wrong type: a string enum was reported as `of type 'integer'`, which is what #563 shows and what the Zowe schema's `zowe.setup.certificate` alternatives produce today.

**Cause.** The messages print `getJSTypeName(valueSpec->type)`. A property spec without a `type` keyword, which is how the `oneOf` alternatives for `certificate` are written, is built with every type name in turn (`typeArray = allJSTypeNames`), so its `typeMask` has every bit set and its `type` is simply the last name assigned, `integer`. The word had nothing to do with the schema.

**Fix.** The message names the declared type only when the schema pins exactly one (one bit in `typeMask`). Otherwise it describes the value the schema itself supplies: the `const` value, or the first `enum` member. `jsonValueTypeName()` maps the JSON value types to the schema type names for that.

Reproduced and fixed against a reduced copy of the certificate `oneOf` (`fixtures/oneof_untyped_schema.json`):

```
before: unequal constant value at zowe.setup.certificate.type; expecting value 'PKCS12' of type 'integer'
        no matching enum value at zowe.setup.certificate.type; expecting one of values '[JCEKS, ...]' of type 'integer'
after:  ... of type 'string'
        ... of type 'string'
```

A schema that does declare `"type": "string"` on those properties was already reported correctly and still is.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/563

## Type of change

## PR Checklist

## Testing

New `tests/configmgr/cli/test_enum_type_message.sh`: validates `oneof_untyped_bad.yaml` against `oneof_untyped_schema.json`, asserts exit 99, that both messages say `of type 'string'`, and that no message says `of type 'integer'`. Fails before, passes after. The whole `tests/configmgr/cli` suite passes on the Linux build from `build/build_cmgr_clang.sh linux`.

## Further comments

Only the message text changes; validation outcomes are untouched. A `const` or `enum` whose members mix types under an untyped spec is described by its first value, which is the same choice the `enum` list in the message already makes.

### DIFF
--- a/c/jsonschema.c
+++ b/c/jsonschema.c
@@ -194,35 +194,22 @@ typedef struct JSValueSpec_tag {
   struct JSValueSpec_tag *parent;
 } JSValueSpec;
 
-/* Type names for validation messages.
-
-   A spec's `type` is meaningful only when the schema names exactly one type.
-   A property without a "type" keyword is built with every type name in
-   turn, so its `type` is merely the last one assigned ("integer") while its
-   typeMask has every bit set; reporting that word made a string enum look
-   like an integer one (#563). When the schema does not pin a single type,
-   describe the value the schema itself supplies instead. */
-static char *jsonValueTypeName(Json *value){
+/* Type word for enum/const messages, taken from the schema's own value */
+static const char *jsonValueTypeName(const Json *value){
   if (value == NULL){
     return "unknown";
   }
   switch (value->type){
   case JSON_TYPE_STRING:  return "string";
   case JSON_TYPE_NUMBER:
-  case JSON_TYPE_DOUBLE:  return "number";
   case JSON_TYPE_INT64:   return "integer";
+  case JSON_TYPE_DOUBLE:  return "number";
   case JSON_TYPE_BOOLEAN: return "boolean";
   case JSON_TYPE_NULL:    return "null";
   case JSON_TYPE_OBJECT:  return "object";
   case JSON_TYPE_ARRAY:   return "array";
   default:                return "unknown";
   }
-}
-
-static char *specTypeNameForMessage(JSValueSpec *spec, Json *sample){
-  int mask = spec->typeMask;
-  bool exactlyOneType = (mask != 0) && ((mask & (mask - 1)) == 0);
-  return exactlyOneType ? getJSTypeName(spec->type) : jsonValueTypeName(sample);
 }
 
 typedef struct AccessPathUnion_tag {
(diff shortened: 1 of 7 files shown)


c/jsonschema.c
```diff
@@ -194,6 +194,37 @@ typedef struct JSValueSpec_tag {
...
 
+/* Type names for validation messages.
+
+   A spec's `type` is meaningful only when the schema names exactly one type.
+   A property without a "type" keyword is built with every type name in
+   turn, so its `type` is merely the last one assigned ("integer") while its
+   typeMask has every bit set; reporting that word made a string enum look
+   like an integer one (#563). When the schema does not pin a single type,
+   describe the value the schema itself supplies instead. */
+static char *jsonValueTypeName(Json *value){
```
ifakhrutdinov: I would make this  `const Json *value` to indicate that the value is not changed in the function; and I'd make the return type `const char *` because we can't physically modify it anyway (or else S0C4).

c/jsonschema.c
```diff
@@ -194,6 +194,37 @@ typedef struct JSValueSpec_tag {
...
+  case JSON_TYPE_INT64:   return "integer";
+  case JSON_TYPE_BOOLEAN: return "boolean";
+  case JSON_TYPE_NULL:    return "null";
+  case JSON_TYPE_OBJECT:  return "object";
+  case JSON_TYPE_ARRAY:   return "array";
+  default:                return "unknown";
+  }
+}
+
+static char *specTypeNameForMessage(JSValueSpec *spec, Json *sample){
```
ifakhrutdinov: Same here: make the arguments you're not mutating `const`.

c/jsonschema.c
```diff
@@ -194,6 +194,37 @@ typedef struct JSValueSpec_tag {
   struct JSValueSpec_tag *parent;
 } JSValueSpec;
 
+/* Type names for validation messages.
+
+   A spec's `type` is meaningful only when the schema names exactly one type.
```
ifakhrutdinov: This is too verbose.


How about this:
```c
/**
 * @brief Converts a JSON type into the corresponding type name.
 *
 * @param[in] value the json to be used.
 * @return the type name or @c NULL if @a value is @c NULL.
 */

```

c/jsonschema.c
```diff
@@ -194,6 +194,37 @@ typedef struct JSValueSpec_tag {
...
+  case JSON_TYPE_OBJECT:  return "object";
+  case JSON_TYPE_ARRAY:   return "array";
+  default:                return "unknown";
+  }
+}
+
+static char *specTypeNameForMessage(JSValueSpec *spec, Json *sample){
+  int mask = spec->typeMask;
+  bool exactlyOneType = (mask != 0) && ((mask & (mask - 1)) == 0);
+  return exactlyOneType ? getJSTypeName(spec->type) : jsonValueTypeName(sample);
```
ifakhrutdinov: Would just calling `jsonValueTypeName(sample)` in all the cases work? That'd simplify the logic.

ifakhrutdinov (approved):
lgtm

PR #684, opened 2026-09-03
Build configmgr-rexx against the same QuickJS as configmgr

JoeNemo:
## Proposed changes

The configmgr workflow moved to `1000turquoisepogs/quickjs-portable` at `feature/update-2024-01-13` in March 2026 (c8bc87b). The configmgr-rexx workflow stayed where Sean pointed it in 2022, `joenemo/quickjs-portable` at `main`, which is the 2021-03-27 engine. Since March the two builds have shipped different QuickJS engines, and any change that touches a newer engine API compiles in one and fails in the other; #680 is the first PR to hit that.

This points the rexx build at the same repository and branch and reports the matching `CONFIG_VERSION`. Four lines.

## Type of change
- [x] Chore, repository cleanup, updates the dependencies.

## Testing

First pass, on z/OS 3.1 with the rexx build's two flag sets: the seven engine sources and `embeddedjs.c qjszos.c qjsnet.c` compiled clean against the 2024 engine. That pass stopped at compile, and the CI link then failed with 31 unresolved `bf_*` symbols: the 2024 engine needs the libbf big-number sources, which `build_cmgr_xlclang.sh` already carries and this script did not. The second commit adds `-DCONFIG_BIGNUM=1`, `libbf.c` and `libbf.o`, matching the configmgr script.

Second pass, the whole script on the Marist system (`STEPLIB=CBC.SCCNCMP sh build_rexxcmgr.sh`, as `.pax/configmgr-rexx/pre-packaging.sh` runs it) against a checkout of the 2024-01-13 branch: compiles and links, `bin/ZWERXCFG` produced, no unresolved symbols, no diagnostics.
## Further comments

Follow-ups this unblocks: #680 can drop a compatibility guard it would otherwise need; the rexx configmgr picks up every engine fix since 2021 (including the ones in the fork's open PRs) at the next build.


ifakhrutdinov (dismissed):
lgtm

ifakhrutdinov (approved):

PR #643, opened 2026-07-18
Some enhancements to the zosfile unit

ChongZhou-Broadcom:
## Purpose of the PR
This is an extension (a super set) of the old unmerged PR #633 . The old one was closed.

The previous PR aligned the `directoryDeleteRecursive()` function with the system command `rm -rf` with regards to the handling of symbolic links.

After that, I noticed that the `directoryCopy()` function has the same issue about symbolic links.

The issue can be summarized as follows: when deleting or copying a directory, symbolic links within the source directory **should be unlinked or recreated in the target directory**, rather than deleting or copying the files or directories to which those links point.

This PR fixes the inconsistant behavior of the previous implementation.

Additionally, it also extracts the directory traversal code into a function shared by all directory operation functions.

## Summary of major changes
- `directoryDeleteRecursive()` and `directoryCopy()` no longer follow links, now they only delete or create links;
- New function `directoryListEntries()` was added and shared by all directory functions.
- New function `fileReadLink2()` was added to implement a convenient and consistent way for checking if a path is a symlink and what the real path is.

## Hints for unit tests

1. The following functions should be tested: `directoryDeleteRecursive`, `directoryCopy`, `directoryChangeModeRecursive`, `directoryChangeOwnerRecursive`, `fileReadLink2`, `directoryListEntries`;
2. Functions listed in point 1 can be tested in a group in a stable order;
3. The test requires two valid system users and groups. (user1 from group1 and user2 from group2)
4. Before all tests, the test framework should create the following directory structure as the test data:
```
test_data/
test_data/dir1/
test_data/dir1/file1
test_data/dir2/
test_data/dir2/file2
test_data/dir2/file1 --(symlink)--> test_data/dir1/file1
test_data/dir2/dir2.1/
test_data/dir2/dir2.1/file3
test_data/dir2/dir2.1/dir1 --(symlink)--> test_data/dir1/
```
All files must be in mode 644, all directories must be in mode 755.
All files and directories must owned by user1 and group1.

5. test 1: directoryListEntries_should_return_expected_entry_list

```c
int rc, rsn;
UnixDirectoryEntries entries;
directoryListEntries("test_data/dir2", &entries, &rc, &rsn);
EXPECT_EQUAL(rc, 0);
EXPECT_EQUAL(rsn, 0);
EXPECT_EQUAL(entries.numEntries, 3);
EXPECT_ARRAY_CONTAINS(entries.entryArray, entries.numEntries, "file1");
[...]

PR #689, opened 2026-09-04
Bugfix: logging component-ID walk is independent of host byte order (#668)

### LINKED ISSUE #668: logging component-ID walk reads the ID's shorts in reverse on a little-endian host
The logging component-ID walk reads the ID's shorts in reverse on a little-endian host

`logConfigureComponent()` and `getComponent()` in `c/logging.c` split a 64-bit
component ID into four shorts by aliasing:

```c
unsigned int vendorID = (compID >> 48) & 0xFFFF;   /* portable */
unsigned short *id = (unsigned short *)&compID;    /* byte-order dependent */
```

The two lines disagree about how to read the same value. The shift is portable;
the cast depends on the host's byte order. Taking
`LOG_COMP_HTTPSERVER = 0x008F000100090000`:

| | `id[0]` | `id[1]` | `id[2]` | `id[3]` |
|---|---|---|---|---|
| big-endian (z/OS) | `008F` | `0001` | `0009` | `0000` |
| little-endian | `0000` | `0009` | `0001` | `008F` |

The loops walk `id[1]`, `id[2]`, `id[3]` to descend the component tree. On z/OS
that is `0001` then `0009` -- the intended path -- terminating on the zero low
short. On a little-endian host it is `0009`, `0001`, then the **vendor ID**, so
the tree is indexed by the wrong values in the wrong order, and the walk never
terminates on a zero because the vendor short is never zero.

Consequence: component registration and log-level lookup do not behave off z/OS
the way they do on z/OS. Nothing on the mainframe is affected -- this only
matters for builds on other platforms, which is where it was noticed.

Suggested fix, matching what the line above already does:

```c
unsigned short idPart = (unsigned short)((compID >> (48 - 16 * i)) & 0xFFFF);
```

or extract the four parts into a local array once, so the loops index something
whose layout does not depend on the host.

Filed out of #667, which fixes a separate bug in the same loops -- an
out-of-bounds read of `id[4]` caused by testing the loop bound after indexing --
and deliberately does not touch this.

JoeNemo:
Fixes #668.

## Proposed changes

The logging component-ID walk read the ID's four shorts in reverse on a little-endian host. A component ID is four big-endian shorts: vendor, then up to three component levels. `logConfigureComponent()`, `getComponent()` (twice) and `logShouldTraceInternal()` aliased the 64-bit ID as `unsigned short *` to get at them, while the vendor check beside them used a shift. On z/OS both agree. On x86 the aliased view is reversed: `id[0]` is the low short, the levels come out backwards, and in `logShouldTraceInternal()` the vendor check itself fails.

The visible effect on Linux: configuring a level-1 component landed on the root component instead, so every zowe component began tracing at that level, siblings included. This is the follow-up #667 left open; that PR fixed the loop bound, not the byte order.

**Fix.** Extract the four shorts with shifts at all four sites. No behaviour change on z/OS, where big-endian aliasing and shifts agree.

This PR addresses Issue: https://github.com/zowe/zowe-common-c/issues/668

## Type of change

## PR Checklist

## Testing

New `tests/logging/component_id_test.c` with `build.sh` (clang with ASan/UBSan on Linux, xlclang on z/OS): configures vendor 8F / level-1 component 1 at DEBUG and asserts

| check | staging logging.c on Linux | this branch |
|---|---|---|
| the configured component traces at DEBUG | PASS | PASS |
| its child (1, 9) inherits the level | PASS | PASS |
| sibling component 2 does NOT trace | **FAIL** | PASS |
| the configured component does not trace above DEBUG | PASS | PASS |

`LOGGING_C=/path/to/other/logging.c sh build.sh` runs the same test against another copy of the file, which is how the "before" column was produced. The Linux configmgr build (`build/build_cmgr_clang.sh linux`) compiles clean and the `tests/configmgr/cli` suite still passes.

## Further comments

The test cannot fail on z/OS, by construction; its value there is as a regression guard for the walk itself. On Linux it is the first test that exercises the logging component tables at all.

## Review changes

Per review, the read is now forked by host byte order behind one named macro, `LOG_COMPONENT_ID_SHORTS(id, compID)` in `logging.h`, used at the four sites in `logging.c` and in the statement-expression form of `logShouldTrace()` that `logging.h` also carries. On a big-endian host (`__ZOWE_BIG_ENDIAN`, which `zowetypes.h` now names next to its existing byte-order test) it expands to the original in-place `(unsigned short *)&compID`, so the z/OS code is what it was; on a little-endian host it extracts the four shorts with shifts. The trailing issue-number comments are gone.

Verified on Linux: `tests/logging` passes and the configmgr CLI suite passes. The Linux build of staging is currently broken by #643's `fileReadLink` signature change; #694 fixes that, and the verification here was done with that one-line fix applied. z/OS compile pending Marist's return; by construction the z/OS preprocessed source is the pre-PR code.


ifakhrutdinov:
Ok, I need to think about this one a bit more...

c/logging.c
```diff
@@ -579,7 +581,8 @@ static LoggingComponent *getComponent(LoggingContext *context, uint64 compID, in
     LoggingComponentTable *componentTable = component->subcomponents;
     maxLevel = component->currentDetailLevel > maxLevel ? component->currentDetailLevel : maxLevel;
 
-    unsigned short *id = (unsigned short *)&compID;
+    unsigned short id[4] = { (unsigned short)(compID >> 48), (unsigned short)(compID >> 32),
+                             (unsigned short)(compID >> 16), (unsigned short)compID }; /* #668 */
```
ifakhrutdinov: Why leave comments like this `/* 665 */`? We have git history which for that.

c/logging.c
ifakhrutdinov, line 762: This function also exists as a macro in `logging.h` so that may also need to be changed.

c/logging.c
JoeNemo: Comment is bogus

JoeNemo:
We should fork the implementation of component id array by endianness of platform and use ifdefs and named macros for clarity.

ifakhrutdinov (approved):
I've built ZIS and compared the listings before and after the changes, they look similar. So, at least big endian is fine.

PR #690, opened 2026-09-04
Bugfix: size the base64 decode buffer with the divide first (#641)

### LINKED ISSUE #641: httpfileservice: base64 result buffer sizing overflows before the divide
writeBinaryDataFromBase64 and writeAsciiDataFromBase64 size the decode buffer as

    resultBufferSize = (conversionLength * 3) / 4;   (also line 915 with dataSize)

The multiply is int and wraps before the divide once the length exceeds INT_MAX/3, so the buffer comes out undersized and decodeBase64 overruns it. Dividing first avoids the overflow and is exact, since decodeBase64 rejects input that is not a multiple of 4:

    resultBufferSize = (conversionLength / 4) * 3;

#617 guards the input buffer only; this is the output side.

JoeNemo:
Fixes #641.

## What

`writeBinaryDataFromBase64()` and `writeAsciiDataFromBase64()` in `httpfileservice.c` size the base64 decode buffer as

```c
resultBufferSize = (conversionLength * 3) / 4;
```

with `int` arithmetic. Above `INT_MAX / 3` bytes the multiply wraps, the quotient comes out small or negative, and `decodeBase64()` then writes the whole decoded payload into that buffer. This PR changes both sites to

```c
resultBufferSize = (conversionLength / 4) * 3;
```

which cannot wrap and is exact here: `decodeBase64()` rejects any length that is not a multiple of 4, so the two expressions agree on every input the decoder accepts.

## Why now

The payload cap that #617 added keeps the length far below the wrap today, so this is latent. #617 guarded the input side; this closes the output side at the source, so the expression stays right whatever the cap becomes. Two lines plus a comment; no behaviour change for any payload the server accepts today.

## Testing

`httpfileservice.c` is compiled by no zowe-common-c build or CI job. It is built only inside ZSS, so the check for this change is a ZSS build:

- Fresh clone of zss `v3.x/staging` (c17bb18) on the Marist system, `deps/zowe-common-c` pointed at this branch, together with the `httpserver.c` change from the sibling PR for zowe/zss#810 (a different translation unit).
- `build/build_zss.sh` (xlc, 31-bit, `-Wc,list`) and `build/build_zss64.sh` (xlclang, 64-bit) both completed with rc 0 and linked `zssServer` and `zssServer64`.
- The 31-bit listing for `httpfileservice.c` carries the new lines and no diagnostics. The only messages in either leg are the pre-existing `IEW2689W` side-file warning and the `CCN3684` note on `main` in `zss.c`.

Not exercised at run time: reaching the wrap needs a payload above `INT_MAX / 3` bytes, which the #617 cap rejects first.


JoeNemo:
Is there any guarding that the length is truly a multiple of 4 because the invariance of the output is dependent on that.

Martin-Zeithaml (approved):
`conversionLength` is expected to be verified [here](https://github.com/zowe/zowe-common-c/blob/6e6fb4b9d5176ce594508d9e1ee6d9000f89aa1b/c/utils.c#L1179).

Then `resultBufferSize` will be always the same (no lost due to floor division).

PR #691, opened 2026-09-04
Bugfix: return the ICSF codes when the session token key cannot be generated (zowe/zss#810)

### LINKED ISSUE #810: `ZWES1036E` prints return and reason code 0
If there is an issue when creating http server, following error message `ZWES1036E` is not helpful:

```
ZWES1035I ZSS Server settings: Address='0.0.0.0', port='7557', protocol='https'
made https server at 0x0
ZWES1036E Server startup problem: Ret='0', res='0x0'
```

https://github.com/zowe/zss/blob/0f9310e748e601725cd4e7b34e704a4b114b87ee/c/zss.c#L1852-L1860

JoeNemo:
Fixes zowe/zss#810.

## What

When ZSS fails to start its HTTP server, it prints

```
ZWES1036E Server startup problem: Ret='0', res='0x0'
```

The codes are zero because of one failure path in `httpserver.c`. `makeHttpServerInner()` generates the session token key with `icsfGenerateRandomNumber()`. When that call fails, `initSessionTokenKey()` logged the ICSF return and reason codes at DEBUG and returned -1, and the constructor returned NULL without touching the caller's `returnCode` and `reasonCode`. The usual cause is that the server user has no access to the `CSFRNG` service, and the operator got nothing that pointed at ICSF.

This PR:

- gives `initSessionTokenKey()` the constructor's `returnCode` and `reasonCode`. Those two slots carry the socket layer's codes, errno and the BPX reason, which `zss.c` compares with `EADDRINUSE`, so the ICSF failure is reported in the same vocabulary: `EACCES` for ICSF return code 8, the usual case of no access to the CSFRNG service, `EIO` otherwise, reason 0. The ICSF return and reason codes themselves are on the SEVERE log line immediately before `ZWES1036E`;
- raises the log line to SEVERE.

The z/OS-only path is the only one touched; the non-z/OS branch of the function is unchanged, and `makeHttpServerInner()` keeps its signature.

## Testing

`httpserver.c` is compiled by no zowe-common-c build or CI job. It is built only inside ZSS, so the check for this change is a ZSS build:

- Fresh clone of zss `v3.x/staging` (c17bb18) on the Marist system, `deps/zowe-common-c` at this change, together with the `httpfileservice.c` change from the sibling PR for #641 (a different translation unit).
- `build/build_zss.sh` (xlc, 31-bit, `-Wc,list`) and `build/build_zss64.sh` (xlclang, 64-bit) both completed with rc 0 and linked `zssServer` and `zssServer64`.
- The 31-bit listing for `httpserver.c` carries the new code and no diagnostics. The only messages in either leg are the pre-existing `IEW2689W` side-file warning and the `CCN3684` note on `main` in `zss.c`.

The failure path itself was not driven: it needs a server user without `CSFRNG` access, and the built server was not started. The success path is the one every ZSS start takes and is unchanged apart from the two extra arguments.


Martin-Zeithaml (dismissed):

c/httpserver.c
```diff
@@ -1526,7 +1534,7 @@ HttpServer *makeHttpServerInner(STCBase *base,
   /* if "noTCP" is true, this server only receives IO thru tunnelling */
   bool noTCP = false;
   SessionTokenKey sessionTokenKey = {0};
-  if (initSessionTokenKey(&sessionTokenKey) != 0) {
+  if (initSessionTokenKey(&sessionTokenKey, returnCode, reasonCode) != 0) {
```
ifakhrutdinov, line 1537: The return and reason code values returned from this function are POSIX ones: you can see that in `zss.c` where rc gets compared with `EADDRINUSE`:
```c
if (returnCode==EADDRINUSE) {
  zowelog(NULL, LOG_COMP_ID_MVD_SERVER, ZOWE_LOG_SEVERE, ZSS_LOG_PORT_OCCUP_MSG, port);
}
```
I don't think it's a good idea to mix those with return and reason codes from ICSF.

JoeNemo: Error codes need to be merged to *ONE* well-known enumeration.  Please express uniformly in POSIX-ese.

h/httpserver.h
```diff
@@ -450,6 +450,8 @@ HttpResponse *makeHttpResponse(HttpRequest *request, ShortLivedHeap *slh, Socket
    */
 #define HTTP_DISABLE_TCP_PORT 0x00DEAD00
 
+/* On failure returnCode is a POSIX errno and reasonCode the z/OS reason code of the
```
ifakhrutdinov, line 453: 🙄

ifakhrutdinov (dismissed):

ifakhrutdinov (approved):

PR #693, opened 2026-09-05
Add writeYAMLWithWidth() so PARMLIB updates fit the record length (#550)

### LINKED ISSUE #550: `updateConfig` does not respect record length
When [updating config](https://github.com/zowe/zowe-install-packaging/blob/27796ca23660798392beae2d71c421a34566f4b1/bin/libs/configmgr.ts#L358) in data set, updated config is written like to unix file. Which is fine for each line with length <= record length.
Otherwise the line is trimmed.

For example `zwe init security --config "PARMLIB(ZOWE.PARMS(YAML))" --jlc --update-config`

The security product is misconfigured and will be updated (should be `RACF`).

```yaml
zowe:
  setup:
    dataset:
      prefix: ZOWE.PR#4459
      jcllib: ZOWE.PR#4459.JCLLIB
    security:
      product: ACF2
    jcl:
      enable: true
      header: '12456'
  runtimeDirectory: "/u/users/robot/pullRequests/pr4459\
/longPathToZoweRuntimeDirectoryToTestSplitFeatureAndPrintWarning"
```

Resulted config is:
```yaml
---
zowe:
  setup:
    dataset:
      prefix: "ZOWE.PR#4459"
      jcllib: "ZOWE.PR#4459.JCLLIB"
    security:
      product: "RACF"
    jcl:
      enable: true
      header: "12456"
  runtimeDirectory: "/u/users/robot/pullRequests/pr4459/longPathToZoweRuntimeDir
...
```

JoeNemo:
Fixes #550.

## What

`zwe ... --update-config` on a `PARMLIB(...)` configuration writes the updated `zowe.yaml` by asking configmgr for YAML text and copying it into the member. libyaml breaks double-quoted strings only at spaces, so a long path such as `runtimeDirectory` comes out on one line, and `cp` then trims that line to the record length without a word. The member ends up with a truncated value.

This PR adds a width-aware writer and exposes it to `zwe`:

- `json2Yaml2BufferWithWidth(json, &buf, &len, maxWidth)`: emits with libyaml's own folding off, then folds every line longer than `maxWidth`. The one construct in the output that can be split without changing its value is a double-quoted scalar, using an escaped line break: `\` at end of line, continuation indented two more spaces, and a continuation that would start with a space escapes it as `\ ` because the reader drops leading white space. Breaks fall after a space when one is available in the second half of the line. A line that cannot be folded within the width, such as a long key, returns `YAML_LINE_TOO_LONG` (20) instead of being written short.
- `ConfigManager.writeYAMLWithWidth(configName, maxLineLength)`: same result shape as `writeYAML`, `[status, text]`. `writeYAML` is unchanged. It is a new method rather than an optional argument because the binding prints a complaint for every call whose argument count differs from the declaration.
- The string case of `writeJsonAsYaml1()` no longer copies the value through a fixed 1024-byte stack buffer; a value longer than that overran it.

The `zwe` side is a small follow-up in zowe-install-packaging: pass the member's LRECL to `writeYAMLWithWidth` in `writeToCfgPath` for `PARMLIB(...)` destinations and stop with a message when the status is non-zero, instead of copying.

## Why the folder is safe without parsing YAML

It never sees arbitrary YAML. It sees the output of this file's own emitter, which is a small dialect: block collections, one node per physical line once libyaml's width is off, keys plain, every string value double-quoted, numbers and booleans plain, no comments, anchors, tags or flow collections, and no raw line breaks inside a scalar because libyaml escapes them as `\n`. Within that dialect:

- An unescaped `"` is always a scalar delimiter. libyaml only forces quoting when a `"` is the first character of a plain scalar, so a key such as `a"b` could have been emitted plain; keys containing a quote are now emitted double-quoted to keep the invariant.
- A break is only ever inserted inside the last double-quoted region on the line, which is the value. Quoted keys are never folded, because an implicit mapping key must stay on one line.
- Escape sequences are kept whole, a continuation that would start with a space escapes it, and trailing spaces before the escaped break are preserved by the YAML rules for double-quoted scalars.
- Anything else fails closed with `YAML_LINE_TOO_LONG`. The folder never guesses.

## What the output looks like

Width 60, values from the test fixture:

```yaml
zowe:
  runtimeDirectory: "/u/users/robot/pullRequests/pr4459/lon\
    gPathToZoweRuntimeDirectoryToTestSplitFeatureAndPrintWa\
    rning"
  description: "a value with spaces that also runs far \
    past the record length of an eighty byte parmlib \
    member line and keeps going"
  spaceAtBreak: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa   bbb   \
    ccc"
  items:
  - "/first/long/list/member/that/does/not/fit/the/record/l\
    ength/at/all/really/not"
  - "short"
  port: 7554
```

## Testing

`tests/configmgr/cli/test_write_yaml_width.sh` runs `fixtures/write_yaml_width.js` through `configmgr -script`: it writes the fixture at width 60, checks that `writeYAML` still produces a wider line (so the fixture exercises folding), that no folded line exceeds 60, that the folded text contains escaped line breaks, that the folded text loads back with data identical to the original (`JSON.stringify` equality, covering escapes, a break inside a run of spaces and a list member), and that a key wider than the width is refused.

`tests/configmgr/cli/test_write_yaml_width_fuzz.sh` runs `fixtures/write_yaml_width_fuzz.js`: 200 random documents whose values are drawn from an alphabet full of quotes, backslashes, spaces, tabs and YAML punctuation, with random key lengths and random widths from 24 to 103. Each document is written as YAML, loaded, folded, stored, loaded again through the real parser and compared with the original data. A refusal is accepted only when a key alone does not fit the width. The seed is fixed, so a failure names a reproducible iteration. On Linux: 156 documents folded and read back identical, 44 legitimately refused, 0 failures. The CLI suite passes on Linux.

On z/OS (Marist) the branch builds with both compilers the repo uses, `build_cmgr_xlclang.sh` (xlclang) and `build_rexxcmgr.sh` (classic xlc, which is what CI's first run of this PR failed on, over `case` labels built from `static const` variables; they are an enum now). Both scripts run against the xlclang-built configmgr pass: all eleven unit checks, and the same 200 random documents with the same 156 folded, 44 refused, 0 failures. The folded file the unit script stores is tagged IBM-1047 with the same shape as above. That run also caught a portability slip in the first version of the folder, which compared the libyaml buffer against native character literals; the folder now uses the ASCII code points that libyaml emits.


JoeNemo:
May not be exercised too much yet.  But as tooling around writing configs increases, will become a problem.

PR #694, opened 2026-09-09
Match the POSIX fileReadLink() to its const declaration

JoeNemo:
#643 changed the declaration of `fileReadLink()` in `unixfile.h` to take `const char *fileName` and updated the z/OS implementation, but not the POSIX one in `platform/posix/psxfile.c`. Since then the Linux build of configmgr (`build/build_cmgr_clang.sh linux`) fails with `conflicting types for 'fileReadLink'`. CI does not build for Linux, so nothing caught it.

One word. With it the Linux build compiles and the configmgr CLI suite passes. No z/OS change.


ifakhrutdinov (approved):
lgtm

PR #598, opened 2026-05-25
Minor XML updates

Martin-Zeithaml:
# baos check

We are correctly checking the buffer position, but line later we will update buffer outside the defined memory.

# `vsnprintf` vs `vsprintf`

To stay "inside" the buffer, use `vsnprintf`.

## Simplified unit test
```c
#include <stdio.h>
#include <stdarg.h>
#include <string.h>

#define BUFFER_SIZE 32

static void syntaxError(char *formatString, ...) {
    char buffer[BUFFER_SIZE];
    va_list argPointer;
    va_start(argPointer, formatString);
    vsprintf(buffer, formatString, argPointer);
    va_end(argPointer);
    printf("output: \"%s\"\n", buffer);
}

static void syntaxError_FIX598(char *formatString, ...) {
    char buffer[BUFFER_SIZE];
    va_list argPointer;
    va_start(argPointer, formatString);
    vsnprintf(buffer, sizeof(buffer), formatString, argPointer);
    va_end(argPointer);
    printf("output: \"%s\"\n", buffer);
}

int main(void) {
    char long_input[200];
    
    memset(long_input, 'A', sizeof(long_input) - 1);
    long_input[sizeof(long_input) - 1] = '\0';

    printf("\ninput:  \"short\"\n");
    syntaxError("short");

    printf("\ninput:  \"%s\" (%d chars)\n", long_input, (int)(sizeof(long_input) - 1));
    syntaxError("%s", long_input);
    
    printf("\ninput:  \"short\"\n");
    syntaxError_FIX598("short");

    printf("\ninput:  \"%s\" (%d chars)\n", long_input, (int)(sizeof(long_input) - 1));
    syntaxError_FIX598("%s", long_input);

    return 0;
}
```

## Result
```
input:  "short"
output: "short"

input:  "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" (199 chars)
[...]

PR #698, opened 2026-09-11
[V2] Minor xml updates

Martin-Zeithaml:
V2 version of #598 

* Check position before update
* `vsnprintf` instead of `vsprintf`
* Return if `EOF` detected in comment


JoeNemo (approved):
Approved.

PR #705, opened 2026-09-11
[V2] Correct string length in calling compareIgnoringCase

ChongZhou-Broadcom:
Porting PR #607 from V3 to V2.


JoeNemo (approved):
Approved.

PR #623, opened 2026-07-02
Open new file with exclusive flag

Martin-Zeithaml:
Open new file with `O_EXCL` flag: `FILE_OPTION_CREATE_IF_NON_EXISTENT = (O_CREAT | O_EXCL)`

PR #707, opened 2026-09-11
Bugfix: allow copy overwrite (open dest with O_CREAT, not O_CREAT|O_EXCL)

struga0258:
## Problem
`fileCopyConverted()` opened the copy destination with `FILE_OPTION_CREATE_IF_NON_EXISTENT`,
which is defined as `O_CREAT | O_EXCL`. When the destination already exists, `open()` fails
with `EEXIST` *before* permissions are checked, so an intended overwrite fails. ZSS surfaces
this as `rc=14` / HTTP 500 ("Error copying file").

## Root cause
`CREATE_IF_NON_EXISTENT` bundles `O_EXCL`, so an earlier attempt to fix this by removing the
redundant standalone `FILE_OPTION_EXCLUSIVE` was a no-op (`0xC0 | 0x40 == 0xC0`).

## Fix
Open the destination with `FILE_OPTION_CREATE` (`O_CREAT`) instead of
`FILE_OPTION_CREATE_IF_NON_EXISTENT`, so `O_TRUNC` overwrites an existing file
(`0221` = `O_CREAT | O_TRUNC | O_WRONLY`).


Martin-Zeithaml:
Blame me for this; this was introduced in #623
* the second updated file `c/httpfileservice.c` will fail too: file exists, `forceWrite` == 1 → falls through to `fileOpen`, which fails `EEXIST` because of `O_EXCL`
* Probably the best solution is to revert #623 and remove both changelog updates (and keep the explanation in commit message here only).

JoeNemo:
Reproduced and verified on z/OS (Marist), with a configmgr script that copies a file onto an existing destination through `xplatform.fileCopy`, which is the same `fileCopy()` the ZSS copy service calls:

```
staging (before this PR)
  PASS: first copy to a new destination
  FAIL: second copy onto the existing destination (status,rc,rsn=-1,117,90308707)   117 = EEXIST
  FAIL: destination holds the second contents ("first contents\n")

with this PR's zosfile.c
  PASS: second copy onto the existing destination
  PASS: destination holds the second contents
```

So the fix does what it says. Two things to weigh before merging:

1. #623 changed two sites the same way. The other is `writeEmptyUnixFile()` in `httpfileservice.c`, the `touch` behind the ZSS create-file service with `forceWrite`: it allows the overwrite, then opens with `FILE_OPTION_CREATE_IF_NON_EXISTENT` and fails the same way. Same one-word change there would close the whole regression.
2. What #623 was after: there is no NOFOLLOW in the BPX open options, so the exclusive flag was the only open-time guard against a destination that is a planted symlink. Overwrite and exclusive are incompatible by definition; the durable answer for an overwrite path is to check the destination is not a symlink first (`fileReadLink2()` from #643 does this) and then open with truncate. Fine as a follow-up; noting it so the reason for #623 is not lost.

The script is small and z/OS-only; happy to hand it over as a `tests/configmgr/cli` fixture if you want it in the PR.

PR #708, opened 2026-09-15
Declare the registers the inline TROO clobbers (#644)

### LINKED ISSUE #644: Clobber list missing registers
**Describe the bug**

The clobber list of the `__asm` code recently added to [c/charsets.c](https://github.com/zowe/zowe-common-c/pull/596/changes#diff-e06baa59a061532cf811790b93396e0f1142d6de0318fd13b28b865c8581c350R260-R272) is missing registers. This will likely lead to crashes and/or unexpected behaviour.

This has been introduced in #596.

**Steps to Reproduce**
n/a

**Expected behavior**
n/a

**Describe your environment**
n/a

**Additional context**
n/a

JoeNemo:
Fixes #644.

The `__troo` replacement for Open XL in `charsets.c` loads R0 (test byte), R1 (table), R2 (input), R8 (output) and R9 (length); `TROO` advances R8, R9 and R2 as it runs, and `IPM 15` writes R15. The `__asm` declared only `"cc"` and `"memory"`, so the compiler was free to keep a live value in any of those six registers across the statement and read garbage afterwards. Whether that happens depends on what the optimizer chose to keep in registers around the call site in `convertCharset()`, which is why nothing has been attributed to it yet.

One line: the six registers are added to the clobber list, in the form `alloc.c` already uses for its own inline assembler.

## Testing

Compiled on z/OS: the configmgr CI build (xlclang, Marist) is green on this branch, and `charsets.c` is part of it, so the register names are accepted. A clobber list cannot change the generated code inside the asm; it only makes the compiler save and restore around it, which is the point. Not exercised at run time: whether a live value was ever kept in one of those registers across the call depends on what the optimizer chose at each call site.


c/charsets.c
```diff
@@ -388,7 +388,7 @@ static int __troo(char *output, char *input, unsigned long inputLength,
         "         ST    15,%0\n"
```
Martin-Zeithaml: The change is fine.

I would only suggest to add comments for the asm code:
```
        "         IPM   15\n"           /* CC -> bits 34-35 of r15 */
        "         SRL   15,28\n"        /* right-justify the 2-bit CC */
        "         ST    15,%0\n"        /* save CC for the caller */
```

Or different wording, just some explanation for non-assembler readers.

Martin-Zeithaml (dismissed):
There is an optional update in this [comment](https://github.com/zowe/zowe-common-c/pull/708/changes#r4082105241).

Otherwise LGTM.

Martin-Zeithaml (approved):

PR #711, opened 2026-09-17
Fix redundant entriesLength*2 in returnEntries

ChongZhou-Broadcom:
https://github.com/zowe/zowe-common-c/blob/d46735f38ef551ff414e4abab2c62ac55893f33a/c/jcsi.c#L372-L375

`entriesLength` has already been doubled 2 lines above, but when updating `entrySet->size`, it is multiplied by 2 again.

Please let me know if unit tests are needed for this PR, thanks.


JoeNemo (approved):
Approved.

